Issue #5, opened 2025-11-14
Not specifying max_segment_size_bytes causes crash when creating the stream queue


SimonUnge:
**Describe the bug**
Not specifying max segment size causes crash:

```
2025-11-13 21:50:28.493204+00:00 [info] <0.11253.0> Stream: __s1_1763070519675129823 will use /.../rabbit@ip-172-31-17-31/mnesia/rabbit@ip-172-31-17-31/stream/__s1_1763070519675129823 for osiris log data directory
2025-11-13 21:50:28.493302+00:00 [info] <0.11252.0> rabbit_stream_coordinator: started writer __s1_1763070519675129823 on 'rabbit@ip-172-31-17-31' in 4437
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>   crasher:
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>     initial call: osiris_writer:init/1
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>     pid: <0.11253.0>
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>     registered_name: []
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>     exception exit: {{badkey,max_segment_size_bytes},
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>                      [{rabbitmq_stream_s3_log_manifest,writer_manifest,1,
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>                           [{file,"rabbitmq_stream_s3_log_manifest.erl"},
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>                            {line,245}]},
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>                       {osiris_log,init,2,
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>                           [{file,"src/osiris_log.erl"},{line,537}]},
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>                       {osiris_writer,handle_continue,2,
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>                           [{file,"src/osiris_writer.erl"},{line,213}]},
2025-11-13 21:50:28.493497+00:00 [error] <0.11253.0>                       {gen_batch_server,handle_continue,2,
[...]

PR #6, opened 2025-11-14
See #5. Fix issue with not configuring max segment size bytes

SimonUnge:
*Issue #5 

Assign, not extract, value.


the-mikedavis (approved):
Good catch! The relevant section in osiris_log is here: https://github.com/rabbitmq/osiris/blob/82ef2fd30f227f5a227d76286d981c668eb9a4cb/src/osiris_log.erl#L501-L503

There's no guarantee that the key will exist so we should be using map put syntax `=>` rather than update syntax `:=`. I didn't run into this crash since I was using `streams-perf-test` which I think sets the relevant x-header by default.

The intent of this update is to prevent clients from messing with the segment size via the x-header which controls it. I'm not 100% sure we really need to control this ourselves but where we're cutting up segments into smaller bits, it could be nice to be in control of the total size.

PR #11, opened 2025-11-25
manifest: Apply uploaded fragments to empty manifest in-order

the-mikedavis:
While running a single-instance broker with this perf-test command for example

    stream-perf-test --producers 1 --consumers 0 --streams sq --size 10000 --time 2

The two seconds of publishing creates nearly a gigabyte of data, so the log manifest attempts to upload many fragments in quick succession. S3 most will likely accept these uploads out of order.

The log manifest was already handling this scenario and queueing up finished uploads to be applied to the manifest in order. However it was always applying the first fragment if the manifest didn't exist. If the first successful upload was not fragment zero then this logic in the log manifest would get stuck, and the manifest would not be correct in the remote tier. To fix this we have to check if the first upload is fragment zero and queue up the fragment as pending if it is not.


SimonUnge (approved):
Cool!

PR #14, opened 2025-11-25
Factor out functional core of log manifest

the-mikedavis:
Previously `rabbitmq_stream_s3_log_manifest` contained all of the logic for uploading fragments and maintaining the manifest in the remote tier. This made it very large and hard to follow and test. This change factors out the logic of maintaining the manifest into a functional core / imperative shell style.

`rabbitmq_stream_s3_log_manifest_machine` is functional. It evolves events to update its state-machine state and produces effects. Events and effects are both records where events describe something that occurred (like a new fragment of data becoming available) and effects describe an intent to perform some impure action (like upload that fragment). By separating out events and effects from the execution of tasks, we make it very easy and fast to test state-machine logic.


the-mikedavis:
I found some easy fixes to make along with these changes while looking at the dialyzer failures. Marking as a draft until I can include those changes in this PR...

PR #35, opened 2026-01-21
Move the S3 integration into this plugin

the-mikedavis:
Previously we were using `rabbitmq_aws`'s functions for executing HTTP calls against AWS services. `rabbitmq_aws` is quite old and has some baggage in its API that makes using `gun` directly more appealing. For example, `rabbitmq_aws` funnels all HTTP requests through a single `gen_server` and it uses the built-in `httpc` HTTP client. We can't use it very effectively without large changes which might accidentally break its other uses within RabbitMQ. It also uses list strings pervasively and, since that code was written, Erlang has included many new tools to comfortably and efficiently work with binaries such as `binary:encode_hex/2`.

In the log reader module we will want to use async requests to start fetching data from further ahead in the stream as we reach the end of the currently cached buffer. And in the log manifest we will want to use chunked transfer encoding to upload fragments without reading entire fragments into memory. Introducing these tricks into `rabbitmq_aws` would've been difficult - using `gun` directly makes this much easier. `rabbitmq_aws` is also a little bit of overkill for this plugin: we only need PUT, GET, POST and DELETE and we will only talk to the S3 endpoints (and also instance metadata, for instance-based auth). By tailoring the client to our use of S3 we end up relying on a smaller footprint of code.

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -1,0 +1,812 @@
+%% Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+%% SPDX-License-Identifier: Apache-2.0
+
+-module(rabbitmq_stream_s3_api_aws).
+-moduledoc """
+A wrapper around the AWS S3 HTTP API.
+""".
+
+-export([
+    init/0,
+    open/0,
+    close/1,
+    get/3,
+    get_range/4,
+    put/4
+]).
+
+-define(ALGORITHM, "AWS4-HMAC-SHA256").
+-define(ISOFORMAT_BASIC, "~4.10.0b~2.10.0b~2.10.0bT~2.10.0b~2.10.0b~2.10.0bZ").
+-define(TABLE, ?MODULE).
+-define(METADATA_TOKEN_TTL_SECONDS, 60).
+-define(REGION_KEY, rabbitmq_stream_s3_api_aws_region).
+
+-behaviour(rabbitmq_stream_s3_api).
+
+%% NOTE: the `rabbitmq_stream_s3_api:connection()` type is a `pid()` here - the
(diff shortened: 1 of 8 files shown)


src/rabbitmq_stream_s3_api_aws.erl
```diff
@@ -0,0 +1,816 @@
...
+            Err
+    end.
+
+-doc "Uploads the given `Data` as an object at key `Key`".
+-spec put(connection(), key(), iodata(), request_opts()) -> ok | {error, any()}.
+put(Conn, Key, Data, Opts) when is_pid(Conn) andalso is_binary(Key) andalso is_map(Opts) ->
+    Headers =
+        case Opts of
+            #{crc32 := Checksum} ->
+                #{<<"x-amz-checksum-crc32">> => base64:encode(<<Checksum:32/unsigned>>)};
```
the-mikedavis: In my manual testing I saw an error from S3 that the CRC32 of a fragment didn't match in a PUT request but I haven't been able to reproduce it :thinking:

the-mikedavis:
One more thing we should test on this branch: we should be able to drop the commits related to `rabbitmq_aws` out of the branch of RabbitMQ we're using (https://github.com/amazon-mq/upstream-to-rabbitmq-server/tree/streams-tiered-storage). That shrinks diff significantly. Before +1102 -1063 ([compare](https://github.com/amazon-mq/upstream-to-rabbitmq-server/compare/7b3faa5cb1bae8f182758f34a82f56aa8b535504...a95427553c4a502d07f31f9c4ac0ed81044ceb07)), after +28 -27 ([compare](https://github.com/rabbitmq/rabbitmq-server/compare/66bd21842f7d929560117b45c0622f86fe5d78a7...8312842b1bb7dc3ff72cdf41c875f9c00af43b7e)).

Issue #17, opened 2025-12-04
Delete stream data from the remote tier when a stream is deleted


the-mikedavis:
**Is your feature request related to a problem? Please describe.**

Currently stream deletion is not handled for the remote tier. There's a TODO in `rabbitmq_stream_s3_log_manifest:delete/1` (callback for `osiris_log_manifest:delete/1`). When a stream queue is deleted, its files in the remote tier should be deleted too.

**Describe the solution you'd like**

When deleting a stream, `rabbit_stream_coordinator` deletes individual members (writers & replicas) without a hint that it's deleting the entire cluster. We might be able to make a change in the server and/or Osiris to expose that information. Or we could work around it by tracking membership ourselves and when we see the number of members drop to zero we know that a stream is deleted. Something periodic similar to CMR could run every now and then and compare the set of stream queues to what we have in the remote tier.

**Describe alternatives you've considered**

The easiest alternative is to do nothing during deletions. S3 storage without access is dirt cheap so storing even TB of data in the long run is not noticeably expensive. Having data that can't actually be deleted which does cost money seems like a gap though, so this suggestion is not really serious.

PR #43, opened 2026-01-28
Protect against conflicting updates to the manifest

the-mikedavis:
*Issue #, if available:* discussion of options in #32. ~This also adds a stub for remote tier deletion (#17).~

*Description of changes:*

When partitioned, a stream writer may still think it is a writer but a new writer can be elected on the majority side of the partition. If that happens, the old writer could update a manifest as it performs retention, for example, while the new writer updates the manifest to reflect newly uploaded fragments. If these writers act independently then they can unwittingly overwrite information from each other.
    
Having two writers live at a time should be rare in practice, so we can use an optimistic lock on a manifest to avoid conflicting changes.
    
The optimistic lock is coordinated by Khepri. Khepri stores a mapping between stream ID and the current incarnation of the manifest (`uid()`). New manifests can be safely uploaded since they have unique IDs, and once uploaded we can attempt to update the entry in Khepri. In the common case that will succeed smoothly. If another writer has made a change in the meantime, our write will be rejected and we can re-resolve the manifest and try again.
    
We leverage Khepri's `#if_payload_version{}` and `#if_data_matches{}` conditions to reject out-of-date writes. `#if_payload_version{}` is robust enough for the optimistic lock on its own in this case, since stream IDs are not duplicated by stream deletion / recreation. But we also check that the writer making the change has an epoch (election number) higher than or equal to the last update. This can help fence off the deposed writer so that it cannot make further changes after the successor writer has made any change.

### DIFF
--- a/include/rabbitmq_stream_s3.hrl
+++ b/include/rabbitmq_stream_s3.hrl
@@ -160,11 +160,16 @@
 %% = 35
 -define(ENTRY_B, 35).
 
-%% A nicer version of the above `?MANIFEST/5' macro. This is the root of the
-%% manifest. A newly created, empty manifest can be identified by checking
-%% `#manifest.next_offset =:= 0'. Checking `#manifest.total_size =:= 0' or
-%% `#manifest.entries =:= <<>>' is not sufficient since the remote tier can
-%% become empty from retention.
+%% A nicer version of the above `?MANIFEST/5' macro.
+%%
+%% This is the root of the manifest. A newly created, empty manifest can be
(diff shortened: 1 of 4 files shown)


the-mikedavis:
I dropped off the stored-procedures and keep-while parts of this in `rabbitmq_stream_s3_db`. There's a bug with stored procedures that causes an error within Khepri at the moment (https://github.com/rabbitmq/khepri/issues/356). That will need to be fixed first if we want to use this kind of mechanism for deletions.

I'll split those changes into a separate PR

PR #77, opened 2026-03-18
Fix `find_fragment` timestamp predicate and add property-based tests

lukebakken:
## Summary

Fixes a bug in `find_fragment` that caused timestamp subscriptions to start from the wrong position, fixes several other bugs in the remote tier read path, and adds comprehensive property-based tests.

## Bug fix: `find_fragment` timestamp predicate

`find_fragment` used `Ts >= FTs` to select the manifest fragment for a timestamp subscription. When `Ts` falls between two fragments — after fragment N's `LTs` but before fragment N+1's `FTs` — the predicate incorrectly returned fragment N, causing the reader to deliver the wrong message.

The correct predicate is `Ts >= LTs`: find the first fragment whose last timestamp is at or after `Ts`. The `read_from_remote_tier_by_timestamp` integration test is also fixed to capture `Timestamp3` before publishing offset 2.

## Other fixes

- Fix `range_spec_to_location_number/2`: suffix range used addition instead of subtraction; open-ended byte range returned `infinity` instead of the remaining file size
- Fix `init_offset_reader/2` calling `osiris_log` directly for the local path; route through `init_local_reader/2` instead
- Preserve `?C_OSIRIS_LOG_FIRST_OFFSET` after local segment deletion; only update the shared atomic used for local/remote tier routing

## Tests

- Add integration tests for reading from the remote tier by offset and timestamp
- Add property-based tests for `range_spec_to_location_number/2`, `find_fragment` timestamp behaviour, `find_index_position` offset and timestamp specs, and `rabbitmq_stream_s3_array` (`partition_point`, `binary_search_by`, `rfind`, `fold`)
- Add diagnostic logging to surface manifest entry timestamps in CI output
- Upload CT logs as an artifact on test failure

### DIFF
--- a/src/rabbitmq_stream_s3_log_manifest.erl
+++ b/src/rabbitmq_stream_s3_log_manifest.erl
@@ -227,6 +227,18 @@ recover_fragments(IdxFile, Acc) ->
     ).
 
 recover_fragments(
+    _Threshold0,
+    _SegmentFile,
+    SegmentOffset,
+    SeqNo0,
+    _NumChunks0,
+    Fragments0,
+    <<>>
+) ->
+    %% The index file exists but contains no records yet (e.g. the stream was
+    %% created but no segment has been flushed). Treat as an empty fragment.
+    {#fragment{segment_offset = SegmentOffset, seq_no = SeqNo0}, Fragments0};
+recover_fragments(
     Threshold0,
     SegmentFile,
     SegmentOffset,


the-mikedavis:
Ah yep I think the :genie: is on the right track here. If a timestamp is between two chunks then we should resolve to the later chunk. (Which is different than offsets, we would resolve to the earlier chunk for a requested offset which is between two chunks.) We should do the same for fragment boundaries :+1:

docs/configuration.md
the-mikedavis: This is great to have. The current docs point to rabbitmq_aws which doesn't make any sense anymore!

PR #91, opened 2026-04-03
Optimize remote-tier reader

### LINKED ISSUE #94: Pre-fetch stream data in the background in the remote reader
(Split from https://github.com/amazon-mq/rabbitmq-stream-s3/issues/42)

The remote reader currently fetches a buffer (5MiB) from the requested offset and serves future reads using that buffer, as long as the buffer contains the requested bytes. When the buffer runs out, the next slice of segment data is requested synchronously in the remote reader's gen_server. This makes for very uneven footprint which swings widely with the latency of reading from S3.

Instead of letting the pre-fetched buffer run out, we should notice when the buffer is running low and request the next slice of data in the background. `gun` supports both synchronous and asynchronous request styles. Since the remote reader is already a gen_server, the asynchronous style is a better fit anyways.

This will increase the complexity of the remote reader significantly because it will need to be refactored to be state-machine-like (i.e. storing whether a background request is ongoing, and what kind of request it is, and the results of failed requests). However this should significantly increase throughput when consuming with `stream-perf-test`, which consumes as fast as possible. And for all readers it should make consumption throughput much more even.

This will open up the door for tuning the read-ahead bytes dynamically based on consumption speed (#95).

### LINKED ISSUE #95: Dynamically adjust remote reader pre-fetches
(Split from https://github.com/amazon-mq/rabbitmq-stream-s3/issues/42). This is blocked on #94.

`stream-perf-test` consumes a stream at max speed because it doesn't do anything interesting with the messages. Real stream consumers will likely be slower. If we hard-code constants in the plugin with optimize for `stream-perf-test`, we will most likely waste memory for actual consumers.

We should tune the number of bytes that we fetch from the remote based on the demand that the reader exerts. [AIMD](https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease) is a simple but effective algorithm for this. Also see conceptually similar work upstream in `gen-batch-server`, [gen-batch-server#29](https://redirect.github.com/rabbitmq/gen-batch-server/pull/29#issuecomment-4163125446), [gen-batch-server#23](https://redirect.github.com/rabbitmq/gen-batch-server/issues/232).

the-mikedavis:
*Issue #, if available:* fixes #94, fixes #95.

*Description of changes:* This is still WIP.

The first commit updates the format of the fragment object to move the fragment's metadata (first offset, first/last timestamp, next offset, etc.) to the beginning of the object from the end. This will make it much cheaper to move from one fragment to the next in the remote-tier reader. By overreading from the beginning of the object we can get the `#fragment_info{}` containing info that makes future pre-fetches cheap, and we get a bunch of bytes of chunk data which we will read. **This is incompatible with the old layout.** Existing deployments of the plugin will need to delete all remote tier data before or after upgrading.

TODO: I believe some of the fragment diagrams are no longer correct. We'll need to update those SVGs.

Previously the remote reader would request a large buffer and use that to serve until a read requested data past the end of the buffer. Then it would synchronously fetch the next buffer. This lead to a really choppy consumption rate which was very sensitive to remote tier latency.
    
With this change, we request a large buffer and then request the next buffer once the current read position nears the end. We use the async APIs introduced in the parent commit to run this request in the background, so the remote reader process remains responsive and serves any available data as the request runs. This roughly doubles the throughput when consuming with stream-perf-test and makes consumption much more even.

### DIFF
--- a/src/rabbitmq_stream_s3_remote_reader.erl
+++ b/src/rabbitmq_stream_s3_remote_reader.erl
@@ -6,7 +6,11 @@
 A gen_server process which reads stream data from the remote tier.
 
 Reads from the remote tier suffer from high latency, so this server pre-fetches
-stream data aggressively.
+stream data aggressively. The pre-fetch amount is adjusted dynamically based on
+the log reader's demand and S3's supply following a basic additive increase /
+multiplicative decrease ([AIMD]) algorithm.
+
+[AIMD]: https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease
 """.
 
 -include_lib("kernel/include/logger.hrl").
@@ -16,9 +20,16 @@ stream data aggressively.
 
 -behaviour(gen_server).
 
-%% TODO: this should be adjusted dynamically based on demand.
-%% <https://github.com/amazon-mq/rabbitmq-stream-s3/issues/95>
--define(READAHEAD, "16MiB").
+%% 1 MiB
+-define(MIN_READ_SIZE, 1048576).
+%% 64 MiB
+-define(MAX_READ_SIZE, 67108864).
+%% 4 MiB
+-define(INITIAL_READ_SIZE, 4194304).
+%% Number of consecutive buffer hits before we increase read_size.
+-define(HITS_TO_GROW, 8).
+%% Additive increase step size (1 MiB in bytes).
+-define(GROW_STEP, 1048576).
 
 -define(C_BUFFER_HIT, 1).
 -define(C_BUFFER_MISS, 2).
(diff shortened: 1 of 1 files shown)


the-mikedavis:
In the long run to maximize throughput we will probably want to have multiple requests in parallel for a single remote reader, and request info about the next fragment more eagerly (say when we get to 50% of the current fragment). To do either we will need more than one connection per remote reader. So I think we should take this opportunity to refactor the API to use connection pooling instead of explicit open/closing of conns.

I will port out some of the more boring changes here and make that refactor before the change to the remote reader.

lukebakken:
I was _just_ reading these docs yesterday -
* https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-design-patterns.html
* https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html

the-mikedavis:
remote-reader-dashboard

I added some instrumentation so we can diagnose issues with the buffer.

Issue #33, opened 2026-01-20
Handle fragment deletion in readers


the-mikedavis:
A reader might perform a range request against a fragment object which is deleted by remote tier retention (#18).

In Osiris, when a reader is switching segments, if the segment doesn't exist, the reader moves up to the first offset at whatever current head of the stream. Reads within an open segment should not fail because open file descriptors can be read even after the file is unlinked. (TODO: is this guaranteed to be true? Even on Windows?)

The problem with reads from the remote tier is that byte range requests will fail once the object is removed. A reader could be within the middle of a chunk and require more data to complete the body of the chunk when the fragment object is deleted. The callback `osiris_log_reader:send_file/3` (used for streams protocol readers) operates on entire chunks but the callback `osiris_log_reader:chunk_iterator/3` can return successfully without fully reading the chunk. If we read ahead by at least the current chunk size then we can handle 404s and jump ahead to the earliest tiered fragment.

PR #106, opened 2026-04-08
Fix consumer crash when reading from S3 after retention

lukebakken:
Fixes #105. Related: #33.

## What this fixes

When consumers fall back to reading from S3 after local segments are deleted by retention, `rabbit_stream_reader` crashes with `{badmatch, {error, invalid_chunk_header}}` and the supervisor hits `reached_max_restart_intensity`. This is reproducible on `main` within ~3 minutes of a high-throughput test with `--max-length-bytes 10gb`.

See #105 for the full root cause analysis.

## Changes

**Fix `advance_fragment` using wrong starting position in new fragment** (8ca3e45)

`advance_fragment` set `position = ?SEGMENT_HEADER_B` (8) when attaching to a new fragment. Fragment objects have the layout `[58-byte fragment header][chunk data][index]` — chunk data starts at byte 58 (`?FRAGMENT_HEADER_B`), not byte 8. Diagnostic logging confirmed the crash was occurring at position 66 (an earlier overcorrection) and then at position 8 on `main`.

**Add diagnostic logging around invalid_chunk_header crash** (629669a)

Logs position, fragment, next_offset, and first bytes of the header binary when `parse_header` fails. Also logs the position and result in `read_header1` and the position returned by `find_position0`. These logs identified the root cause.

**Fix undefined `get_fragment_trailer/2` call in `find_position0`** (388c2dd)

A prior partial revert left a call to the non-existent `rabbitmq_stream_s3_server:get_fragment_trailer/2` in `find_position0`. Renamed back to `get_fragment_info/2`.

**Handle `{error, normal}` from `add_child` in `advance_fragment`** (addcd23)

Broadened the error match in `advance_fragment` from `{error, not_found}` to `{error, _}` to handle `{error, normal}` returned when `init/1` stops normally (e.g. fragment deleted by retention).

**Jump to oldest S3 fragment when `advance_fragment` hits a deleted one** (d93589e)

When `advance_fragment` cannot attach to the next fragment because it has been deleted by retention, jump to `first_offset` from the manifest rather than returning `end_of_stream`.

**Handle S3 errors in `rabbitmq_stream_s3_log_reader` gracefully** (7bbf6c7)

Return `{error, not_found}` from `init/1` when the fragment key does not exist in S3, rather than crashing.

**Log unexpected HTTP status codes from S3 API calls at debug level** (56280e0)

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -156,7 +156,11 @@ get(Conn, Key, Opts) when is_pid(Conn) andalso is_binary(Key) andalso is_map(Opt
             {ok, Data};
         {ok, #{status := 404}} ->
             {error, not_found};
-        {ok, Other} ->
+        {ok, #{status := Status} = Other} ->
+            ?LOG_DEBUG(
+                "rabbitmq_stream_s3_api_aws:get/3 unexpected status ~b for key ~ts",
+                [Status, Key]
+            ),
             {error, Other};
         {error, _} = Err ->
             Err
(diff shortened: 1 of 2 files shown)


lukebakken:
The 🧞 and I are flailing on this.

lukebakken:
Removing this from the 1.0.0-beta.1 milestone: it was closed without merging, so no release contains anything from it. #105, the issue it targeted, was fixed separately and carries the 1.0.0-beta.1 milestone itself.

PR #112, opened 2026-04-10
Fix consumer crash when reading from S3 after retention

### LINKED ISSUE #33: Handle fragment deletion in readers

lukebakken:
Fixes #105.

## What this fixes

Consumers crash with `reached_max_restart_intensity` when reading from the remote (S3) tier after local segments are deleted by retention. The root cause is two unimplemented code paths in `rabbitmq_stream_s3_remote_reader:try_read/3`.

## Changes

**Handle deleted fragment by jumping to oldest available**

When a fragment is deleted by retention while a consumer is actively reading it, S3 returns 404, setting `current_not_found = true`. `try_read/3` previously crashed with `erlang:error(unimplemented)`.

Cancel all in-flight requests for the deleted fragment and call `get_range/1` to find the oldest fragment still in the remote tier. If the remote tier is empty (stream fully deleted), return `end_of_stream`. Otherwise reset the reader state to start from the oldest available fragment, setting `info = undefined` so `maybe_start_current_request` issues a new GET from byte 0 of the new fragment. Return `{next_fragment, FirstOffset}` so the log reader updates its `next_offset` accordingly.

Extract `cancel_requests/1` to replace the inline list comprehension in `terminate/2` and use it in both call sites.

**Handle exhausted buffer with no in-flight requests**

When the buffer has insufficient data and there are no in-flight requests, `try_read/3` previously crashed with `erlang:error(unimplemented)`. This occurs when a chunk is larger than the current read-ahead window (AIMD may reduce `read_size` to 1 MiB minimum).

Bump `read_size` to cover the needed bytes and call `maybe_start_current_request` directly. Calling `maybe_start_request` instead would skip the current fragment fetch: its first clause guards on `EndPos - CurrentPos > ReadSize`, which is always true after the initial fetch due to AIMD halving, causing the remote_reader to sit in `await` indefinitely until the 60-second `gen_server:call` timeout.

## Testing

Verified with a 30-minute high-throughput test (6 streams, 12 producers, 12 consumers, `--max-length-bytes 10gb`). The previous `main` crashed within ~3 minutes when retention first fired. With this fix, the full 30-minute test ran without crashes, consumers read from S3 throughout (including the `current_not_found` path being exercised repeatedly in logs), and full replay from `first` across all 6 streams completed successfully (57,813,558 / 57,838,579 messages).

### DIFF
--- a/src/rabbitmq_stream_s3_remote_reader.erl
+++ b/src/rabbitmq_stream_s3_remote_reader.erl
@@ -305,11 +305,7 @@ handle_info(Msg, State) ->
     {noreply, State}.
 
 terminate(_Reason, #?MODULE{requests = Requests}) ->
-    _ = [
-        rabbitmq_stream_s3_api:cancel_async(RequestId, ReqState)
-     || RequestId := #request{state = ReqState} <- Requests
-    ],
-    counters:put(counter(), ?C_REQUESTS_IN_FLIGHT, 0),
+    cancel_requests(Requests),
     ok.
 
 format_status(#{state := #?MODULE{} = State} = Status0) ->
@@ -638,19 +634,39 @@ try_read(#?MODULE{end_pos = EndPos} = State, Offset, Bytes) when Offset + Bytes 
         #?MODULE{requests = Requests} when map_size(Requests) =/= 0 ->
             %% The current fragment's info is being requested. Wait for its arrival.
             {await, State};
-        #?MODULE{current_not_found = true} ->
-            %% TODO: fragment deletion. Skip ahead to the next fragment if the
-            %% stream still exists, or convert to a local reader if that has
-            %% older data than the remote tier. (Possible if retention
-            %% reclaims fragments but not the current segment.)
-            erlang:error(unimplemented);
+        #?MODULE{current_not_found = true, stream = StreamId, requests = Requests} ->
+            %% The current fragment was deleted by retention while being read.
+            %% Cancel all in-flight requests for the deleted fragment, then jump
+            %% to the oldest fragment still available in the remote tier.
+            cancel_requests(Requests),
+            {FirstOffset, _} = rabbitmq_stream_s3_server:get_range(StreamId),
+            Key = rabbitmq_stream_s3:fragment_key(StreamId, FirstOffset),
+            ?LOG_WARNING(
+                "Fragment ~20..0B was deleted by retention while being read. "
+                "Jumping to oldest available fragment ~20..0B",
+                [State#?MODULE.fragment, FirstOffset]
+            ),
+            NewState = State#?MODULE{
+                fragment = FirstOffset,
+                key = Key,
+                info = undefined,
+                buffer = <<>>,
+                start_pos = ?FRAGMENT_HEADER_B,
+                current_pos = ?FRAGMENT_HEADER_B,
+                end_pos = ?FRAGMENT_HEADER_B,
+                next = undefined,
+                requests = #{},
+                current_not_found = false
+            },
+            {next_fragment, NewState, FirstOffset};
         #?MODULE{read_size = ReadSize0} ->
             %% The chunk is larger than the read-ahead. Bump read_size to
-            %% cover the needed bytes so maybe_start_request issues a
-            %% sufficiently large request.
+            %% cover the needed bytes and call maybe_start_current_request
+            %% directly - bypassing the maybe_start_request guard that skips
+            %% current requests when the buffer already exceeds read_size.
             Needed = Offset + Bytes - EndPos,
             ReadSize = max(ReadSize0, Needed),
-            {await, State#?MODULE{read_size = ReadSize}}
+            {await, maybe_start_current_request(State#?MODULE{read_size = ReadSize})}
     end;
 try_read(
     #?MODULE{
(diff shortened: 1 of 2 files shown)


the-mikedavis:
I might've already covered the exhausted buffer part in ee69db70c2d9c4125f23ea7bd37f83f2f4f8de5b

the-mikedavis (commented):
The changes to the current_not_found branch are looking good to me. We might want to log at debug or info levels though since it's pretty benign.

src/rabbitmq_stream_s3_remote_reader.erl
```diff
@@ -638,19 +615,48 @@ try_read(#?MODULE{end_pos = EndPos} = State, Offset, Bytes) when Offset + Bytes
...
+            ?LOG_DEBUG(
+                "try_read: no requests in flight, buffer exhausted. "
+                "offset ~b bytes ~b end_pos ~b state ~0p",
+                [Offset, Bytes, State#?MODULE.end_pos, format_state(State)]
+            ),
+            Required = Offset + Bytes - State#?MODULE.end_pos,
+            BumpedState = State#?MODULE{
+                read_size = max(State#?MODULE.read_size, Required)
+            },
+            {await, maybe_start_current_request(BumpedState)}
```
the-mikedavis: I'm not sure we need to start a request explicitly here. In `maybe_reply/2` we always start a request if `try_read/3` returns `await`. The comment about the guard is correct though. I'm not sure the layout of `maybe_start_request/1` actually makes sens. We should always prioritize reading up to the `index_start_pos` in the current fragment and independently of that we can request the next fragment.

lukebakken: The 🧞 traced the exact execution path with the values from the debug log to verify this.

When `try_read/3` returns `{await, State1}`, `maybe_reply/2` does:

```erlang
State2 = adjust_read_size(miss, State1),
State = maybe_start_request(State2),
```

`adjust_read_size(miss, ...)` halves `read_size`. If it was bumped to `Needed = 4280752`, after the miss it becomes `max(1048576, 2140376) = 2140376`.

Then in `maybe_start_request/1`, the guard is:

```erlang
when EndPos - CurrentPos > ReadSize ->
    maybe_start_next_request(State);  %% skips maybe_start_current_request
```

With the values from the debug log: `EndPos - CurrentPos = 4194363 - 58 = 4194305`, and `ReadSize = 2140376`. Since `4194305 > 2140376`, the first clause fires and only `maybe_start_next_request` is called - `maybe_start_current_request` is never reached. No current fragment request is started.

This is what caused the 60-second `gen_server:call` timeout: the remote_reader sits in `await` indefinitely with no in-flight request.

Agree with your point that `maybe_start_request/1` layout is the real issue - it shouldn't skip the current fragment fetch just because the buffer exceeds `read_size`. Calling `maybe_start_current_request` directly here is a targeted fix, but restructuring `maybe_start_request` to always prioritize the current fragment independently of the next-fragment pre-fetch would be the cleaner long-term solution.

the-mikedavis: Ah good point. With the AIMD adjustments, read_size will be decreased. We can revisit maybe_start_request/1 later and try to make more sense of it.

the-mikedavis (approved):
This is looking good for solving the larger-than-read-ahead chunk size problem 👍

src/rabbitmq_stream_s3_remote_reader.erl
```diff
@@ -638,19 +634,39 @@ try_read(#?MODULE{end_pos = EndPos} = State, Offset, Bytes) when Offset + Bytes
...
-        #?MODULE{current_not_found = true} ->
-            %% TODO: fragment deletion. Skip ahead to the next fragment if the
-            %% stream still exists, or convert to a local reader if that has
-            %% older data than the remote tier. (Possible if retention
-            %% reclaims fragments but not the current segment.)
-            erlang:error(unimplemented);
+        #?MODULE{current_not_found = true, stream = StreamId, requests = Requests} ->
+            %% The current fragment was deleted by retention while being read.
+            %% Cancel all in-flight requests for the deleted fragment, then jump
+            %% to the oldest fragment still available in the remote tier.
```
the-mikedavis, line 640: We may also want to handle stream deletion while a reader is consuming. I think here if `get_range/1` returns `empty` we could say `end_of_stream`

lukebakken: See the latest code!

the-mikedavis (approved):

PR #114, opened 2026-04-10
reader: Convert a lagging local reader to remote reader

the-mikedavis:
If a reader starts at the tail it will read from the local tier. It might process messages slower than messages are being published, and the segment which is being read might disappear. When that happens we should close the local reader and open a remote one at that offset. The default Osiris behavior would be to skip ahead to the next available segment. This change builds on new work in the Osiris branch which refactors the osiris_log_reader functions to expose the gap where a local reader recognizes that it is missing the data at its offset.

### DIFF
--- a/src/rabbitmq_stream_s3_log_reader.erl
+++ b/src/rabbitmq_stream_s3_log_reader.erl
@@ -349,11 +349,22 @@ send_file(
 send_file(Socket, #?MODULE{mode = Local0} = State0, Callback) ->
     case osiris_log:send_file(Socket, Local0, Callback) of
         {ok, Local} ->
-            State = State0#?MODULE{mode = Local},
-            {ok, State};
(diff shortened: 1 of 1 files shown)


the-mikedavis:
To test this I published to a stream at a high rate

```
stream-perf-test --producers 1 --consumers 0 --streams sq-3 --size 1000 --rate 20000
```

and simultaneously consumed with a low rate

```
perf-test -x 0 -y 1 -sq -u sq-3 --predeclared --consumer-rate 5000
```

`perf-test` consumers start at the tail by default. It eventually fell behind after a minute or two.

become-remote

(I also added a PR to stream-perf-test to set the consumer rate, https://github.com/rabbitmq/rabbitmq-stream-perf-test/pull/396)

lukebakken (commented):
👍

PR #117, opened 2026-04-11
Improve HTTP pool

the-mikedavis:
This is series of small improvements to the HTTP connection pool to S3 based on what some official SDKs do.

* Reap idle connections eagerly instead of waiting for S3 to close them.
* Along those lines, allow the pool to shrink to zero so that an idle broker doesn't waste resources reconnecting every few seconds/minutes. This also means turning off automatic reconnection (default in Gun).
* Switch from FIFO to LIFO checkouts (`queue:queue(conn())` -> `[conn()]` stack) to prioritize reusing hot connections.
* Tune process flags to garbage-collect binaries quicker and reduce memory pressure.


lukebakken:
👍 - from the 🧞 itself:

> Good PR — solid work from Michael.

PR #121, opened 2026-04-14
Fix consumer connection crashes when reading from the remote tier

### LINKED ISSUE #120: Consumer connection crashes when S3 read exceeds 60s timeout under high load
## Summary

Under high consumer load, `rabbit_stream_reader` crashes with a `{timeout, {gen_server, call, ...}}` exception when reading from the remote (S3) tier. The supervisor (`rabbit_stream_connection_sup`) restarts the connection repeatedly until it hits `reached_max_restart_intensity`, at which point the consumer connection is permanently dropped.

## Call stack

```
gen_server:call/3                              (gen_server.erl:1222)  ← exits after 60s
rabbitmq_stream_s3_remote_reader:read/5        (rabbitmq_stream_s3_remote_reader.erl:170)
timer:tc/4                                     (timer.erl:654)
rabbitmq_stream_s3_log_reader:read/4           (rabbitmq_stream_s3_log_reader.erl:785)
rabbitmq_stream_s3_log_reader:send_file/3      (rabbitmq_stream_s3_log_reader.erl:306)
osiris_log_reader:send_file/3                  (osiris_log_reader.erl:86)
rabbit_stream_reader:send_chunks/8             (rabbit_stream_reader.erl:3797)
rabbit_stream_reader:handle_frame_post_auth/4  (rabbit_stream_reader.erl:2100)
```

## What happens

1. `rabbit_stream_reader` calls `rabbitmq_stream_s3_log_reader:send_file/3` synchronously from within its `gen_statem` callback.
2. `send_file` calls `rabbitmq_stream_s3_remote_reader:read/4`, which issues `gen_server:call(RemoteReaderPid, #read{...}, 60000)`.
3. The remote reader stores the read as `pending_read` and returns `{noreply, ...}`, waiting for S3 HTTP response data to arrive as messages to its mailbox.
4. Under high load, the S3 data takes longer than 60 seconds to arrive, so the `gen_server:call` exits with `{timeout, {gen_server, call, [Pid, {read, Offset, Bytes, within_chunk}, 60000]}}`.
5. The exception propagates through `rabbit_stream_reader`, crashing the stream connection process.
6. `rabbit_stream_connection_sup` restarts it, but the same timeout recurs until `reached_max_restart_intensity`.

The 60-second timeout is `?GEN_SERVER_CALL_TIMEOUT` from `rabbit.hrl` - a general-purpose RabbitMQ timeout not tuned for large or slow S3 reads.

## Root cause

`rabbit_stream_reader` runs `send_chunks` synchronously in its `gen_statem` callback. It has no way to yield while waiting for S3 data. The remote reader correctly defers its reply until data arrives, but the caller's 60-second deadline is not coordinated with how long S3 actually takes.

## Reproduction

Run `stream-perf-test` with 6 streams, 12 producers, and 12 consumers against a cluster with the S3 plugin enabled. Consumers reading from the remote tier will begin crashing within a few minutes.

```
stream-perf-test \
  --streams 6 \
  --producers 2 \
  --consumers 2 \
  --rate 5000
```

## Log evidence

From `rabbit@<node>/log/*.log`:

```
** State machine <0.1736.0> terminating
** Reason for termination = exit:{timeout,
                                  {gen_server,call,
                                   [<0.3853.0>,
                                    {read,12663154,265216,within_chunk},
                                    60000]}}
** Callback modules = [rabbit_stream_reader]
```

Followed by:

```
[error] reason: reached_max_restart_intensity
        offender: [{id,rabbit_stream_reader}, ...]
        supervisor: {<pid>,rabbit_stream_connection_sup}
```

Observed on all three nodes of a 3-node cluster simultaneously.

lukebakken:
Fixes #120.

### 1. `retry_requests` message silently dropped (primary bug)

`handle_info(retry_requests, State)` was placed after the generic `handle_info(Msg, #?MODULE{} = State)` clause, making it dead code. Every `retry_requests` message sent via `erlang:send_after` was matched by the generic clause, passed to `match_async`, returned `error`, and logged as an unexpected message - the retry was silently dropped.

This caused the remote reader to stall permanently when it had a pending read and no in-flight S3 requests: the retry message it scheduled to recover was never processed, and the pending read timed out after 60 seconds, crashing the consumer connection.

### 2. `maybe_start_request` blocked the remote reader

`maybe_start_request/1` called `get_range_async`, which called `pool:checkout/2` - a blocking `gen_server:call`. When the HTTP connection pool was exhausted, this blocked the remote reader's process inside `handle_info`, preventing it from processing the HTTP response messages needed to complete in-flight requests and free connections.

### 3. Crash in `find_position0/3` on S3 error

`find_position0` pattern-matched `{ok, Info} = get_fragment_info(...)` unconditionally, crashing with `badmatch` when the S3 request failed (e.g. `{error, {down, {shutdown, closed}}}`).

## Fix

- Move `handle_info(retry_requests, ...)` before the generic `handle_info` clause so it is actually reachable.
- Add `try_checkout/1` to `rabbitmq_stream_s3_api_aws_pool`, which returns `{ok, Conn}` or `busy` immediately without blocking. Use it in `request_async` so `get_range_async` returns `{error, pool_busy}` instead of blocking.
- Introduce a `{retry, State}` return value from `maybe_reply/2` to signal that no requests are in-flight and a retry must be scheduled, preventing the pending read from stalling forever.
- Propagate errors from `get_fragment_info` in `find_position0` rather than crashing on `badmatch`.

## Testing

Reproduced with `stream-perf-test` running 6 streams, 12 producers, and 12 consumers against a 3-node cluster with the S3 plugin enabled. Consumer connections crashed within ~4 minutes before this fix. No crashes observed in a 45-minute run after the fix.

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -611,7 +611,7 @@ cancel_async(StreamRef, #{conn := Conn, stream_ref := StreamRef} = State) ->
 finish_async(#{conn := Conn, pool := Pool} = State) ->
     case State of
         #{timer_ref := TimerRef} ->
-            erlang:cancel_timer(TimerRef),
+            _ = erlang:cancel_timer(TimerRef),
             receive
                 {request_timeout, _} -> ok
             after 0 -> ok
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -56,11 +56,13 @@ A wrapper around the AWS S3 HTTP API.
 -define(C_TOTAL_REQUESTS, 2).
 -define(C_RESPONSE_500, 3).
 -define(C_RESPONSE_503, 4).
+-define(C_REQUEST_TIMEOUTS, 5).
 -define(COUNTERS, [
     {active_requests, ?C_ACTIVE_REQUESTS, gauge, "Current number of requests to S3"},
     {total_requests, ?C_TOTAL_REQUESTS, counter, "Total number of requests to S3"},
     {response_500, ?C_RESPONSE_500, counter, "Number of HTTP 500 responses"},
-    {response_503, ?C_RESPONSE_503, counter, "Number of HTTP 503 responses"}
+    {response_503, ?C_RESPONSE_503, counter, "Number of HTTP 503 responses"},
+    {request_timeouts, ?C_REQUEST_TIMEOUTS, counter, "Number of S3 requests that timed out"}
 ]).
 -define(COUNTER_KEY, {?MODULE, counter}).
 
@@ -599,6 +601,7 @@ handle_async(
     #{conn := Conn, stream_ref := StreamRef} = State
 ) ->
     ?LOG_WARNING("S3 request timed out on ~tw/~tw", [Conn, StreamRef]),
+    counters:add(counter(), ?C_REQUEST_TIMEOUTS, 1),
     gun:cancel(Conn, StreamRef),
     finish_async(State),
     {done, {error, timeout}}.
@@ -704,9 +707,15 @@ await_response(Conn, StreamRef, Timeout) ->
                     },
                     postprocess_response(Response),
                     {ok, Response};
+                {error, timeout} = Err ->
+                    counters:add(counter(), ?C_REQUEST_TIMEOUTS, 1),
+                    Err;
                 {error, _} = Err ->
                     Err
             end;
+        {error, timeout} = Err ->
+            counters:add(counter(), ?C_REQUEST_TIMEOUTS, 1),
+            Err;
         {error, _} = Err ->
             Err
     end.
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -704,15 +704,19 @@ request_async(Method, Path, Headers0, Body, Opts) ->
                 Body,
                 Opts
             ),
-            Cnt = counter(),
-            counters:add(Cnt, ?C_ACTIVE_REQUESTS, 1),
-            counters:add(Cnt, ?C_TOTAL_REQUESTS, 1),
             Pool = ?GENERAL_POOL,
-            Conn = rabbitmq_stream_s3_api_aws_pool:checkout(Pool, 10_000),
-            %% NOTE: no need to wrap this in try/catch and checkin the conn
-            %% since gun:request/5 cannot exit/error/throw.
-            StreamRef = gun:request(Conn, Method, Path, Headers, Body),
-            {ok, StreamRef, #{pool => Pool, conn => Conn, stream_ref => StreamRef}};
+            case rabbitmq_stream_s3_api_aws_pool:try_checkout(Pool) of
+                {ok, Conn} ->
+                    Cnt = counter(),
+                    counters:add(Cnt, ?C_ACTIVE_REQUESTS, 1),
+                    counters:add(Cnt, ?C_TOTAL_REQUESTS, 1),
+                    %% NOTE: no need to wrap this in try/catch and checkin the conn
+                    %% since gun:request/5 cannot exit/error/throw.
+                    StreamRef = gun:request(Conn, Method, Path, Headers, Body),
+                    {ok, StreamRef, #{pool => Pool, conn => Conn, stream_ref => StreamRef}};
+                busy ->
+                    {error, pool_busy}
+            end;
         {error, _} = Err ->
             Err
     end.
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -110,7 +110,10 @@ Called "HTTP Verb" in S3 docs. "GET", "PUT", "HEAD", "POST", "DELETE", etc..
     %% binary when a fairly large amount of data has been collected.
     data => [binary()],
     pending_bytes => non_neg_integer(),
-    timeout => timeout()
+    timeout => timeout(),
+    %% Timer reference for request timeout. Set when a `timeout` is given in
+    %% request opts. Cancelled and flushed in `finish_async/1`.
+    timer_ref => reference()
 }.
 %% Re-use the gun stream ref since it's already a reference.
 -type async_req() :: gun:stream_ref().
@@ -486,6 +489,7 @@ match_async(Msg, Reqs) ->
             {gun_error, _, StreamRef, _} -> StreamRef;
             {gun_response, _, StreamRef, _, _, _} -> StreamRef;
             {gun_data, _, StreamRef, _, _} -> StreamRef;
+            {request_timeout, StreamRef} -> StreamRef;
             _ -> undefined
         end,
     case Reqs of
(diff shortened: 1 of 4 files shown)


src/rabbitmq_stream_s3_api_aws_pool.erl
```diff
@@ -104,6 +105,12 @@ checkout(Pool, Timeout) ->
 checkin(Pool, Conn) ->
     gen_server:cast(Pool, {checkin, Conn}).
 
+%% Non-blocking checkout. Returns `{ok, Conn}` if a connection is immediately
+%% available, or `busy` if the pool is exhausted.
```
the-mikedavis: Let's use `-doc """...""".` strings since we'll require at least OTP27


```suggestion
-doc """
Non-blocking checkout. Returns `{ok, Conn}` if a connection is immediately
available, or `busy` if the pool is exhausted.
""".
```

src/rabbitmq_stream_s3_api_aws_pool.erl
```diff
@@ -104,6 +105,12 @@ checkout(Pool, Timeout) ->
...
+-spec try_checkout(pool()) -> {ok, conn()} | busy.
+try_checkout(Pool) ->
+    gen_server:call(Pool, {try_checkout, erlang:make_ref()}, 5000).
```
the-mikedavis, line 115: I think we might need to repeat the cancel_checkout pattern in the `catch` clause from above here. I'm not totally sure though.

src/rabbitmq_stream_s3_remote_reader.erl
```diff
@@ -351,11 +361,15 @@ maybe_start_current_request(
...
-            State0#?MODULE{requests = Requests0#{RequestId => Request}}
+            case rabbitmq_stream_s3_api:get_range_async(CurrentKey, Range) of
+                {ok, RequestId, AsyncState} ->
+                    Request = #request{fragment = Fragment, range = Range, state = AsyncState},
+                    counters:add(counter(), ?C_REQUESTS_IN_FLIGHT, 1),
+                    counters:add(counter(), ?C_TOTAL_REQUESTS, 1),
+                    State0#?MODULE{requests = Requests0#{RequestId => Request}};
+                {error, pool_busy} ->
+                    State0
+            end
```
the-mikedavis: This code could probably use a helper function at this point, since it's duplicated in the branch below

lukebakken:
The 🧞 and I found something else related to fix here, hang tight.

src/rabbitmq_stream_s3_remote_reader.erl
```diff
@@ -79,7 +83,8 @@ multiplicative decrease ([AIMD]) algorithm.
     %% In the future we could make parallel requests within the same fragment,
     %% so we may need to track the byte range. This is unused currently.
     range :: {byte_offset(), byte_offset()},
-    state :: rabbitmq_stream_s3_api:async_state()
+    state :: rabbitmq_stream_s3_api:async_state(),
+    started_at :: integer()
```
the-mikedavis: Rather than a periodic check based on the started-at time I think we should be able to store the timer ref returned by send_after, and also send the request's ID in the timeout message so we can quickly identify which request timed out

src/rabbitmq_stream_s3_api_aws.erl
```diff
@@ -588,14 +592,37 @@ handle_async(
...
     finish_async(State),
     PendingData = iolist_to_binary(lists:reverse(Data0, [Data])),
-    {data, PendingData, done}.
+    {data, PendingData, done};
+handle_async(
+    {request_timeout, StreamRef},
+    StreamRef,
+    #{conn := Conn, stream_ref := StreamRef} = State
+) ->
+    ?LOG_WARNING("S3 request timed out on ~tw/~tw", [Conn, StreamRef]),
```
the-mikedavis, line 603: Another opportunity for a metric 😉

src/rabbitmq_stream_s3_api_aws.erl
```diff
@@ -588,14 +592,37 @@ handle_async(
...
 -spec cancel_async(async_req(), async_state()) -> ok.
 cancel_async(StreamRef, #{conn := Conn, stream_ref := StreamRef} = State) ->
     gun:cancel(Conn, StreamRef),
     ok = finish_async(State).
 
-finish_async(#{conn := Conn, pool := Pool}) ->
+finish_async(#{conn := Conn, pool := Pool} = State) ->
+    case State of
+        #{timer_ref := TimerRef, stream_ref := StreamRef} ->
+            case erlang:cancel_timer(TimerRef) of
```
the-mikedavis, line 617: I believe this function has options it can accept to flush? I might be thinking of another similar function though.

the-mikedavis: I think it was the connection reaping code in the pool that I thought of. That should have a similar cancellation mechanism.

src/rabbitmq_stream_s3_api_aws_pool.erl
the-mikedavis: Just food for thought here: it would definitely be less likely but maybe still possible that we end up checking out a connection to a `try_checkout/1`? Just since the request might sit in the mailbox behind other messages. I would need to really think this one through though. Not a big risk. The cancellation mechanism is mostly meant to correctly handle callers that also wrap in `try`/`catch` which I don't think we do in the calling code.

src/rabbitmq_stream_s3_api_aws_pool.erl
```diff
@@ -152,6 +162,28 @@ handle_call(
...
+handle_call(
+    {try_checkout, _Checkout},
+    {Pid, _} = _From,
+    #?MODULE{
+        available = Available0,
+        checkouts = Checkouts0,
+        checkouts_rev = CheckoutsRev0
+    } = State0
+) ->
+    MRef = erlang:monitor(process, Pid),
```
the-mikedavis: Instead of monitor+demonitor we can just move this into the branch for when a connection is available

the-mikedavis: This also looks like a good opportunity to introduce a helper function (`try_checkout` or something similar) which returns a Conn and the updated state or undefined if one is not available

src/rabbitmq_stream_s3_log_reader.erl
```diff
@@ -350,19 +350,6 @@ send_file(Socket, #?MODULE{mode = Local0} = State0, Callback) ->
...
-                {ok, State} ->
-                    send_file(Socket, State, Callback);
-                false ->
-                    case osiris_log:open_next_segment(Local1) of
-                        {ok, Local} ->
-                            send_file(Socket, State0#?MODULE{mode = Local}, Callback);
-                        not_found ->
-                            {end_of_stream, State0#?MODULE{mode = Local1}}
-                    end
-            end;
```
the-mikedavis, line 365: I think your local checkout of osiris might be behind, this isn't dead code. It's necessary for turning a lagging local reader into a remote reader

the-mikedavis (commented):
Overall looks great! I have a bunch of minor comments

src/rabbitmq_stream_s3_log_reader.erl
lukebakken: Hm that's odd, my `osiris` is up-to-date. Thanks for catching this.

lukebakken: Ah, my local `.plt` was stale.

src/rabbitmq_stream_s3_api_aws_pool.erl
lukebakken: Just so you know, here's what the 🧞 had to say:

```
11% > So, in his last review, what did the-mikedavis mean by this comment? \
\

Just food for thought here: it would definitely be less likely but maybe still possible that we end up checking out a connection to a try_checkout/1? Just since the request might sit in the mailbox behind other messages. I would need to really think this one through though. Not a big risk. The cancellation mechanism is mostly meant to correctly handle callers that also wrap in try/catch which I don't think we do in the calling code. \
\
https://github.com/amazon-mq/rabbitmq-stream-s3/pull/121/changes#r3087979703

> He's referring to the cancel_checkout mechanism in the pool's checkout handler. Let me read it:
Reading file: /home/lrbakken/development/amazon-mq/upstream-to-rabbitmq-server/deps/rabbitmq_stream_s3/src/rabbitmq_stream_s3_api_aws_pool.erl, from line 100 to 145 (using tool: read)
 ✓ Successfully read 1415 bytes from /home/lrbakken/development/amazon-mq/upstream-to-rabbitmq-server/deps/rabbitmq_stream_s3/src/rabbitmq_stream_s3_api_aws_pool.erl
 - Completed in 0.0s

> His point is this: try_checkout/1 is a gen_server:call. The pool processes messages in order. If the pool's mailbox has other messages queued ahead of the {try_checkout, ...} call - for example, a {checkin, Conn} cast that would make a connection available - those messages are processed
first. By the time the pool processes the try_checkout, a connection may now be available that wasn't when the caller decided to send the request.

But more specifically, he's thinking about the reverse: could a try_checkout call get a connection checked out to it even though the caller has already given up (e.g. timed out on the gen_server:call)? With checkout/2, the caller wraps in try/catch and sends cancel_checkout if it times out
or crashes, so the pool can release the pending slot. With try_checkout, there's no pending slot - it either gets a connection immediately or returns busy.

The scenario he's worried about: the gen_server:call in try_checkout has a 5-second timeout. If the pool is slow to respond (e.g. blocked in open/0 which calls timer:sleep while fetching instance metadata), the caller's gen_server:call could time out and raise an exception, but the pool
still processes the {try_checkout, ...} message later and checks out a connection to the now-dead caller. That connection is then leaked - never checked back in.

His conclusion: not a big risk in practice because request_async doesn't wrap try_checkout in try/catch, so if the call times out it propagates as an exception up the call stack and the connection would be leaked. But since try_checkout only succeeds when a connection is immediately
available (no queuing), the window is very small. The cancel_checkout mechanism in checkout/2 exists precisely to handle this for the blocking/queuing case.
```

the-mikedavis: Seems reasonable 👍 

(This was in reply to the thread)

the-mikedavis (approved):
Lots of great fixes in here. Thanks!

PR #126, opened 2026-04-17
Add dependabot config to update GHA workflows

### LINKED ISSUE #116: Use dependabot to upgrade actions
I believe some actions are already behind, like the one to save RabbitMQ server in the cache. We should have dependabot send us a PR every week (maybe two weeks?) grouping all CI dependency upgrades together.

We could also switch from tags to hashes, but we don't use any sketchy actions and the workflows don't have permissions to do anything interesting, so I'm not sure it's necessary.

lukebakken:
Fixes #116

Copied from [here](https://github.com/amazon-mq/rabbitmq-queue-migration/blob/main/.github/dependabot.yml) and then modified.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
the-mikedavis, line 6:
```suggestion
      interval: "weekly"
    groups:
      github-actions:
        patterns:
          - "*"
```

We can group together all actions so we get 1 PR containing all updates

lukebakken: Oh man I need to do that everywhere, thanks.

PR #127, opened 2026-04-17
Start HTTP pools before server process

lukebakken:
Closes #119.

## Problem

The `rabbitmq_stream_s3_server` boot step starts the server process before the HTTP connection pools exist. Tasks spawned immediately by the server (e.g. `resolve_manifest` triggered by stream coordinator recovery) call into `rabbitmq_stream_s3_general_pool`, which is only started later by the supervision tree via `application:ensure_all_started`. This causes `{noproc, ...}` errors and task retries on every node restart.

The root cause is the ordering in `rabbit:start_apps/2`: boot steps run first, then `application:ensure_all_started` starts the supervision tree.

## Fix

Add a `rabbitmq_stream_s3_http_pools` boot step in `rabbitmq_stream_s3_sup` that starts both pools under `rabbit_sup` before the server process starts. The server boot step now `requires` this new step instead of `rabbitmq_stream_s3` directly:

```
rabbitmq_stream_s3 -> rabbitmq_stream_s3_http_pools -> rabbitmq_stream_s3_server -> core_initialized
```

The pools are removed from the supervision tree since they are started by the boot step.

## Additional changes

- Add `rabbitmq_stream_s3_config` module that centralises all `application:get_env/2,3` calls for the application. All call sites updated. Includes EUnit tests for defaults and configured values.
- Fix the `-type config()` spec in `rabbitmq_stream_s3_api_aws_pool` to include the required `name` field, and fix the corresponding test.
- Fix a pre-existing key name discrepancy: the cuttlefish schema writes region endpoint overrides to `aws_region_endpoints` but the code was reading `region_endpoints`. The code now reads `aws_region_endpoints` to match the schema.

PR #128, opened 2026-04-17
Add `rabbitmq_stream_s3_config` module

lukebakken:
Requires #127.

## Summary

Add a `rabbitmq_stream_s3_config` module that centralises all `application:get_env/2,3` calls for the `rabbitmq_stream_s3` application. All call sites updated to use it. The module includes EUnit tests for all default values and a configured value.

## Additional fixes

- Fix the `-type config()` spec in `rabbitmq_stream_s3_api_aws_pool` to include the required `name` field, and fix the corresponding test.
- Fix a pre-existing key name discrepancy: the cuttlefish schema writes region endpoint overrides to `aws_region_endpoints` but the code was reading `region_endpoints`. The code now reads `aws_region_endpoints` to match the schema.
- Update `start_pools/0` in `rabbitmq_stream_s3_sup` to use `rabbitmq_stream_s3_config` for pool size configuration.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -48,7 +48,7 @@ jobs:
       - name: make
         run: make -C ${{ github.workspace }}/rabbitmq-server/deps/rabbitmq_stream_s3
       - name: tests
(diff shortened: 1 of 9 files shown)


the-mikedavis (approved):
:shipit:

PR #130, opened 2026-04-21
Fix remote reader mailbox flood from cancelled S3 requests

### LINKED ISSUE #122: Remote reader mailbox flood after retention-triggered request cancellation
## Summary

When a fragment is deleted by retention while being read, `try_read/3` sets `current_not_found = true` and `maybe_start_request` calls `cancel_requests/1` to cancel all in-flight S3 requests. `gun:cancel` is called for each request, but gun still delivers any data that was already buffered or in-flight on the TCP connection. These arrive as `gun_data` messages whose `StreamRef` is no longer in the `requests` map, so `match_async` returns `error` and each message is logged as an unexpected message and discarded.

Under high load, this produces a flood of hundreds of stale `gun_data` messages in the remote reader's mailbox. The remote reader processes them one at a time in `handle_info`, and while doing so it cannot process the gun responses for the new request it issued after the jump. If the backlog is large enough, the new request's response is delayed past the 30-second request timeout, which cancels and re-issues the request - potentially triggering another flood. Eventually `rabbit_stream_reader` times out on its 60-second `gen_server:call` to the remote reader and crashes the consumer connection.

## Evidence

From a high-throughput test run (6 streams, 12 producers, 12 consumers):

```
15:51:43 [warning] Fragment 4810243 was deleted by retention while being read.
         Jumping to oldest available fragment 5533579
15:51:43 - 15:52:17  235 "received unexpected message" log entries
15:52:17 [warning]   S3 request timed out on <conn>/<stream_ref>  ← 30s timeout fires
15:52:43 [debug]     rabbit_stream_reader terminating ... {timeout, {gen_server, call, [..., 60000]}}
```

The 30-second request timeout (added in PR #121) fires correctly, but the mailbox backlog prevents the re-issued request from being processed in time.

## Instance type

`m7g.large`

## Fix direction

The `error` branch in `handle_info` processes unexpected messages one at a time. Options:

- Drain all pending unexpected messages in a single `handle_info` pass before returning, rather than processing them one at a time.
- Suppress the debug log for stale `gun_data` messages to reduce overhead (the log itself adds latency under flood conditions).
- Track cancelled `StreamRef`s and explicitly ignore their responses without going through `match_async`.

lukebakken:
Fixes #122.

## Problem

When retention deletes the fragment being read, or an S3 request exceeds the 30-second timeout, the remote reader receives `gun_data` frames for streams that gun has already cancelled, or whose error response body is still in flight. The reader has no matching request in its `requests` map, so `match_async/2` returns `error` and each stale frame is logged one at a time at debug level.

Under high load this produces thousands of stale messages. Processing them serially delays the re-issued request past the 60-second `gen_server:call` timeout that `rabbit_stream_reader` uses, which crashes the consumer connection.

## Root causes

Two `handle_async` paths return `{done, ...}` while gun still has frames to deliver for the same stream:

1. A `nofin` non-success response (404, 500, 503, other). The response headers arrive with `nofin`, the handler returns `{done, {error, ...}}`, and gun delivers the XML error body as `gun_data` frames against the now-removed ref.
2. `request_timeout`, which calls `gun:cancel` and returns `{done, {error, timeout}}`. Frames already buffered on the TCP connection arrive after the ref is removed from `requests`.

## Solution

**Drain error response bodies.** On a `nofin` non-success response, `handle_async` cancels the request timer, marks the state with `draining => Reason`, and returns `{continue, State}`. New clauses discard `gun_data nofin` frames during the drain and, on the terminal `gun_data fin`, call `finish_async` and return `{done, {error, Reason}}`. The request stays in `requests` throughout the drain, so no frame is orphaned.

**Track refs cancelled by `gun:cancel`.** A new `cancelled_requests` field is added to the reader state and the callback changes from `match_async/2` to `match_async/3` so callers pass the cancelled set alongside the active requests. `request_timeout` now returns `{done_cancel, {error, timeout}}`; the reader adds the ref to `cancelled_requests` on this return. Stale frames are matched as `{cancelled, Ref, final | more}` and dropped silently. The ref is removed from `cancelled_requests` on its terminal frame.

Extract `cancel_request_timer/1` from `finish_async/1` for use by the drain path.

## Testing

EUnit coverage:

- `rabbitmq_stream_s3_api_aws`: `match_async_active_request_test`, `match_async_cancelled_request_test`, `match_async_unknown_request_test`, `match_async_connection_error_ignores_cancelled_test`
- `rabbitmq_stream_s3_api_fs`: `match_async_test`

`gmake` and `gmake dialyze` pass cleanly.

`high-throughput-test.sh` on a test deployment, 45-minute run with 6 streams, 12 producers, and 12 consumers: 895 retention jumps across three nodes produced zero "received unexpected message" log entries. Prior runs on unfixed code produced over 16,000.

### DIFF
--- a/src/rabbitmq_stream_s3_api.erl
+++ b/src/rabbitmq_stream_s3_api.erl
@@ -30,7 +30,7 @@ file-system operations. Use that in non-unit tests.
     delete/2,
     delete_prefix/1,
     delete_prefix/2,
-    match_async/2,
+    match_async/3,
     handle_async/3,
     cancel_async/2,
     request_duration_prometheus_format/0
@@ -79,13 +79,20 @@ examples.
 """.
 -callback delete(key() | [key()], request_opts()) -> ok | {error, any()}.
 -callback delete_prefix(key(), request_opts()) -> {ok, map()} | {error, any()}.
--callback match_async(Msg :: term(), #{async_req() := async_state()}) ->
+-callback match_async(
+    Msg :: term(),
+    Reqs :: #{async_req() := async_state()},
+    CancelledReqs :: #{async_req() => _}
+) ->
     {ok, async_req()}
+    | {cancelled, async_req(), final | more}
     | error.
 -callback handle_async(Msg :: term(), async_req(), async_state()) ->
     {continue, async_state()}
     | {data, binary(), async_state() | done}
-    | {done, ok | {error, any()}}.
+    | {done, ok | {error, any()}}
+    | {done_cancel, {error, any()}}
+    | ignore.
 -callback cancel_async(async_req(), async_state()) -> ok.
 
 -define(C_GET, 1).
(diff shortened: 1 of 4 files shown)


the-mikedavis (approved):
Looks good 👍

I was curious if we could do something about discarding the message body for 404 requests, but this covers that now.

PR #132, opened 2026-04-23
Fix `unimplemented` crash on retention race in `try_read/3`

### LINKED ISSUE #131: Remote reader crashes with `unimplemented` on retention race in `try_read/3`
## Summary

`rabbitmq_stream_s3_remote_reader:try_read/3` crashes with `erlang:error(unimplemented)` at `rabbitmq_stream_s3_remote_reader.erl:686` when a retention race condition occurs: the request for the next fragment returned 404, and by the time `try_read/3` re-queries the remote tier range, the oldest available offset has already advanced past the fragment the reader was about to request.

The crash takes down the remote reader process. Every `rabbit_stream_reader` consumer connection that has an in-flight `gen_server:call` to the crashed reader then terminates with a 60-second `timeout` exit, disconnecting the clients.

## Affected code

The branch containing the unimplemented path:

https://github.com/amazon-mq/rabbitmq-stream-s3/blob/main/src/rabbitmq_stream_s3_remote_reader.erl#L660-L698

```erlang
try_read(
    #?MODULE{
        stream = StreamId,
        info = #fragment_info{next_offset = NextFragment, index_start_pos = IdxStartPos},
        next = not_found
    } = State0,
    Offset,
    Bytes
) when Offset >= IdxStartPos ->
    %% The request for the next fragment gave a 404. The fragment doesn't exist
    %% or hasn't been uploaded yet.
    RemoteRange = rabbitmq_stream_s3_server:get_range(StreamId),
    ...
    case RemoteRange of
        {RemoteStartOffset, _EndOffset} when RemoteStartOffset >= NextFragment ->
            %% re-set the state at this offset. Also take the min between this
            %% and the first local offset.
            erlang:error(unimplemented);
        ...
    end;
```

This branch has been `unimplemented` since commit 4a10750 ("Fix two `try_read/3` crashes when reading from the remote tier").

## Reproduction

Observed reliably during `high-throughput-test.sh` runs against the `alloy-2026-04-16` test deployment on 2026-04-21 and 2026-04-22. The test publishes a sustained high-throughput load with aggressive retention, producing thousands of 404 responses for just-evicted fragments and thousands of retention-induced fragment jumps. The race window is small but hit multiple times per hour.

Example counts from a 45-minute run:
- 3042 retention jumps cluster-wide
- 2 `unimplemented` crashes (one on each of nodes 0 and 1)
- 6 `rabbit_stream_reader` state machine terminations caused by the resulting `gen_server:call` timeouts

The race was previously masked by the mailbox flood bug fixed in #130. With that fix merged, the `unimplemented` path is now the dominant crash source in sustained retention workloads.

## Stack trace

Captured from node0 of the `alloy-2026-04-16` deployment:

```
[error] <0.4374.0> ** Reason for termination ==
[error] <0.4374.0> ** {unimplemented,
[error] <0.4374.0>        [{rabbitmq_stream_s3_remote_reader,try_read,3,
[error] <0.4374.0>             [{file,"rabbitmq_stream_s3_remote_reader.erl"},{line,686}]},
[error] <0.4374.0>         {rabbitmq_stream_s3_remote_reader,maybe_reply,2,
[error] <0.4374.0>             [{file,"rabbitmq_stream_s3_remote_reader.erl"},{line,567}]},
[error] <0.4374.0>         {rabbitmq_stream_s3_remote_reader,handle_call,3,
[error] <0.4374.0>             [{file,"rabbitmq_stream_s3_remote_reader.erl"},{line,228}]},
[error] <0.4374.0>         {gen_server,try_handle_call,4,[{file,"gen_server.erl"},{line,2381}]},
[error] <0.4374.0>         {gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,2410}]},
[error] <0.4374.0>         {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,329}]}]}
[error] <0.4374.0> ** Client <0.1073.0> stacktrace
[error] <0.4374.0> ** [{gen,do_call,4,[{file,"gen.erl"},{line,260}]},
[error] <0.4374.0>     {gen_server,call,3,[{file,"gen_server.erl"},{line,1218}]},
[error] <0.4374.0>     {rabbitmq_stream_s3_remote_reader,read,5,
[error] <0.4374.0>         [{file,"rabbitmq_stream_s3_remote_reader.erl"},{line,186}]},
[error] <0.4374.0>     {timer,tc,4,[{file,"timer.erl"},{line,654}]},
[error] <0.4374.0>     {rabbitmq_stream_s3_log_reader,read,4,
[error] <0.4374.0>         [{file,"rabbitmq_stream_s3_log_reader.erl"},{line,793}]},
[error] <0.4374.0>     {rabbitmq_stream_s3_log_reader,read_header1,1,
[error] <0.4374.0>         [{file,"rabbitmq_stream_s3_log_reader.erl"},{line,519}]},
[error] <0.4374.0>     {rabbitmq_stream_s3_log_reader,send_file,3,
[error] <0.4374.0>         [{file,"rabbitmq_stream_s3_log_reader.erl"},{line,297}]},
[error] <0.4374.0>     {osiris_log_reader,send_file,3,
[error] <0.4374.0>         [{file,"src/osiris_log_reader.erl"},{line,86}]}]
```

## Expected behaviour

When the next fragment's requested offset is below the remote tier's current start offset (the fragment has been evicted by retention since the 404), the reader should reset its state to the new `first_offset`, similar to the existing retention-jump branch in the non-`next = not_found` path. The TODO comment in the source indicates the intended logic: "re-set the state at this offset. Also take the min between this and the first local offset."

## Suggested approach

Mirror the existing `current_not_found = true` retention-jump logic: call `rabbitmq_stream_s3_server:get_range/1`, compute the new fragment key from the first available offset, and rebuild the reader state (buffer, positions, fragment info) before returning `{next_fragment, NewState, FirstOffset}` to the caller. The "min between this and the first local offset" aspect of the TODO deserves clarification: it likely means the reader should become local if the offset is also available locally, to avoid a needless S3 round trip.

lukebakken:
Fixes #131.

## Problem

`rabbitmq_stream_s3_remote_reader:try_read/3` crashes with `erlang:error(unimplemented)` when a 404 on the next-fragment request is followed by the remote tier's start offset advancing past `NextFragment`. This happens when retention evicts the next fragment and keeps going, so by the time the reader re-queries the remote range, the oldest available offset is at or beyond the fragment the reader was about to request.

The crash takes down the remote reader. Every `rabbit_stream_reader` consumer connection with an in-flight `gen_server:call` to that reader then terminates with a `timeout` exit, disconnecting the client.

## Solution

Mirror the existing retention-jump logic used in the `current_not_found = true` branch of the `EndPos` clause of `try_read/3`:

1. Read the remote range
2. Jump to the oldest available fragment offset
3. Reset the reader state (fragment, key, buffer, positions, info, next, current_not_found)
4. Return `{next_fragment, State, RemoteStartOffset}`

Unlike the `current_not_found = true` branch, `requests` can be non-empty here. A trailing current-fragment request may still be in flight when the next-fragment probe receives its 404. Cancel those requests and merge their refs into `cancelled_requests` so that late frames are dropped silently by the `match_async/3` mechanism added in #130.

## Testing

The 45-minute `high-throughput-test.sh` on the test deployment against this branch produced:

- 0 `unimplemented` crashes across all three nodes (baseline on `main`: 2 crashes in the same workload)
- 2 occurrences of the new `"Fragment ... finished but next fragment ... was evicted"` warning, confirming the new code path was exercised
- 2,485 total retention-jump events handled cleanly
- 0 `received unexpected message` log entries, confirming PR #130's mailbox-flood fix still holds under the new code path

`gmake`, `gmake dialyze`, and `gmake eunit` (54 tests) all pass on this branch.

### DIFF
--- a/src/rabbitmq_stream_s3_remote_reader.erl
+++ b/src/rabbitmq_stream_s3_remote_reader.erl
@@ -681,9 +681,40 @@ try_read(
     ),
     case RemoteRange of
         {RemoteStartOffset, _EndOffset} when RemoteStartOffset >= NextFragment ->
-            %% re-set the state at this offset. Also take the min between this
-            %% and the first local offset.
-            erlang:error(unimplemented);
+            %% Retention has evicted the next fragment and advanced past it.
+            %% Jump to the oldest fragment still available. Cancel any
+            %% in-flight requests and track their refs as cancelled so that
+            %% late frames are dropped silently.
+            #?MODULE{
+                fragment = CurrentFragment,
+                requests = Requests,
+                cancelled_requests = Cancelled0
+            } = State0,
+            cancel_requests(Requests),
+            Cancelled = maps:merge(
+                Cancelled0,
+                #{Req => ok || Req := _ <- Requests}
+            ),
+            Key = rabbitmq_stream_s3:fragment_key(StreamId, RemoteStartOffset),
+            ?LOG_WARNING(
+                "Fragment ~20..0B finished but next fragment ~20..0B was evicted. "
+                "Jumping to oldest available fragment ~20..0B",
+                [CurrentFragment, NextFragment, RemoteStartOffset]
+            ),
+            State = State0#?MODULE{
+                fragment = RemoteStartOffset,
+                key = Key,
+                info = undefined,
+                buffer = <<>>,
+                start_pos = ?FRAGMENT_HEADER_B,
+                current_pos = ?FRAGMENT_HEADER_B,
+                end_pos = ?FRAGMENT_HEADER_B,
+                next = undefined,
+                current_not_found = false,
+                requests = #{},
+                cancelled_requests = Cancelled
+            },
+            {next_fragment, State, RemoteStartOffset};
         {_StartOffset, EndOffset} when EndOffset < NextFragment ->
             State = goto_next_fragment(State0),
             {become_local, State};
--- a/.github/workflows/format-check.yaml
+++ b/.github/workflows/format-check.yaml
@@ -17,8 +17,8 @@ jobs:
       - uses: actions/checkout@v6
       - uses: erlef/setup-beam@v1
         with:
-          otp-version: 27
-          rebar3-version: 3.22
+          otp-version: '27'
+          rebar3-version: '3.27'
       - run: |
           git clone --branch v1.7.0 https://github.com/WhatsApp/erlfmt.git ${{ github.workspace }}/erlfmt
           cd ${{ github.workspace }}/erlfmt && rebar3 as release escriptize


lukebakken:
@the-mikedavis FYI I just converted this to draft to do one more long-running test (2 hours).

lukebakken:
@the-mikedavis ok, this is OK for review.

src/rabbitmq_stream_s3_remote_reader.erl
```diff
@@ -681,9 +681,40 @@ try_read(
...
+            %% late frames are dropped silently.
+            #?MODULE{
+                fragment = CurrentFragment,
+                requests = Requests,
+                cancelled_requests = Cancelled0
+            } = State0,
+            cancel_requests(Requests),
+            Cancelled = maps:merge(
+                Cancelled0,
+                maps:from_keys(maps:keys(Requests), ok)
```
the-mikedavis:
```suggestion
                #{Req => ok || Req := _ <- Requests}
```

I think this should work the same way and avoid the intermediary list allocation. This probably doesn't have much performance impact of course, I just like avoiding the unnecessary allocs

the-mikedavis (approved):
👍

Issue #36, opened 2026-01-21
Expand the test suite


the-mikedavis:
Currently the test suite is limited to a few EUnit unit tests and a suite which tests the logic of `rabbitmq_stream_s3_log_manifest_machine`. That leaves a lot of uncovered code which needs to be tested manually to make sure the plugin works after any change.

#35 introduces a high-level behaviour `rabbitmq_streams_s3_api` which allows plugging in different backends for the remote tier and splits out the AWS S3 integration as one backend. We could add another backend (say, `rabbitmq_streams_s3_api_fs`) which could store objects in a directory, and CommonTest suites could use their priv dir to simulate a remote tier. This would be a nice alternative to mocking (via `meck`) all interactions with S3 - that would probably become messy over time. We could use this for higher-level integration tests covering entire happy+sad-path workflows.

It would also be nice to have an integration test which would run in CI automatically but not locally (unless you opt in) which tests against AWS S3. TODO for me: figure out how we can provide an AWS account / credentials for CI to make that possible.

the-mikedavis:
@lukebakken I was thinking about this w.r.t. our recent work on the remote reader. I wonder if we could set up suites that work at the Osiris level, like [osiris_SUITE](https://github.com/rabbitmq/osiris/blob/main/test/osiris_SUITE.erl) or [osiris_log_SUITE](https://github.com/rabbitmq/osiris/blob/main/test/osiris_log_SUITE.erl). The end-to-end tests we have in `stream_s3_SUITE` are good for common happy/sad paths. But there are cases like retention taking away fragments that become harder to test at that high level. With lower-level APIs we could seed local and remote logs to suit the test and avoid any timing or concurrency issues.

I haven't looked into how feasible this is, just food for thought.

Edit: added an initial implementation of this idea in https://github.com/amazon-mq/rabbitmq-stream-s3/pull/134

the-mikedavis:
I think the current suite has gotten quite robust lately, so I think we can close this out. I think we're better served by more focused issues now, e.g. #202.

PR #134, opened 2026-04-29
Add a test suite for the log reader

the-mikedavis:
This is a series of commits to add a new test suite for the log reader which we can test very quickly (i.e. without spawning RabbitMQ) and have fine control the local and remote tier logs. This makes it possible to test tricky setups and edge-cases that would otherwise be very complicated to test. This adds two tests to start with. We can expand from there.

* Basic setup where the remote tier has some fragments and then the local tier has data.
* Edge-case where chunks are so small that the log-reader's over-reading behavior would've caused the remote reader to hang.

Connects #36. This is what I was thinking of in https://github.com/amazon-mq/rabbitmq-stream-s3/issues/36#issuecomment-4225993977.

### DIFF
--- a/src/rabbitmq_stream_s3_api_fs.erl
+++ b/src/rabbitmq_stream_s3_api_fs.erl
@@ -31,6 +31,7 @@ associated file in that folder.
 % Auxiliary function for testing
 -export([
     get_stream_data/1,
+    list_fragments/1,
     clear/0,
     set_data_dir/1
 ]).
@@ -256,9 +257,8 @@ stream_finish({Fd, Crc}, Crc32) ->
     {ok, Manifest, [FragmentFile]} | {error, not_found | path_not_set}
 when
     StreamName :: binary(),
-    Manifest :: Path | undefined,
-    FragmentFile :: Path,
-    Path :: binary().
+    Manifest :: file:filename() | undefined,
+    FragmentFile :: file:filename().
 get_stream_data(StreamName0) ->
     case data_dir() of
         undefined ->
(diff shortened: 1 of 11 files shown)


lukebakken (commented):
*(Superseded - see the corrected review below.)*

lukebakken (commented):
First-pass AI review. Build, dialyze, eunit, xref, and the new `log_reader_SUITE` CT suite all pass locally.

## Summary

Four clean commits with a logical progression: behaviour abstraction, manifest module extraction, test suite, bug fix with test. No bugs or correctness issues found.

## Per-commit notes

### b2ac795 - Setup a behaviour so the manifest server can be changed in tests

The `(rabbitmq_stream_s3_server:backend()):get_manifest(StreamId)` dynamic dispatch pattern is clean. The 5 new xref unresolved calls are expected since xref cannot resolve dynamic module calls. The `application:get_env` default ensures production behavior is unchanged.

### cc1e0c2 - Move manifest manipulation functions to a new module

Pure refactor. `rebalance_edit/2` is a nice cleanup - it encapsulates the `#group_uploaded{}` destructuring that was previously inline in the machine's `apply/3`. All call sites updated, including the eunit tests in the machine module.

Note: `rebalance_edit` carries a TODO about potentially needing to clear `next_offset`. This is pre-existing logic (just moved), not introduced by this PR.

### 77aad6c - Add a test suite for the log reader

The test infrastructure is well-designed. `seed_fragment` + `publish_manifest` + `reclaim_local_segments` provide a nice DSL for setting up precise local/remote tier layouts. The ETS-backed behaviour implementation in `rabbitmq_stream_s3_server_ets` is minimal and correct - return types match the callbacks and the `range()` type.

### fca89c8 - Handle over-reading a chunk header into a fragment's index

The new `try_read` clause correctly detects the over-read condition and caps the returned binary at the index boundary. The guard `Offset + ?CHUNK_HEADER_B =< EndPos` ensures the 48-byte chunk header always fits. The bloom filter, if present, is part of the chunk data section and therefore also within the capped range. The log reader's `read_header2` handles the shorter binary correctly since it only pattern-matches `<<HeaderBin1:?CHUNK_HEADER_B/binary, _/binary>>` and then uses `binary:part(HeaderBin, ?CHUNK_HEADER_B, FilterSize)` where `FilterSize` is within the chunk data.

The test `read_last_chunk_at_index_boundary` uses 1-byte messages to reliably trigger this edge case.

lukebakken:
🥳

PR #136, opened 2026-04-29
Fix remote reader stall after S3 request timeout (HTTP/1.1 HOL blocking)

### LINKED ISSUE #133: Remote reader stalls for 30s after S3 request timeout, crashing consumer connection
## Summary

After an S3 request times out in `rabbitmq_stream_s3_remote_reader`, the reader remains idle for 30 seconds until `rabbit_stream_reader`'s outer 60-second `gen_server:call` gives up, crashing the consumer connection. The reader's pending read is never answered and no retry is visibly scheduled, even though the timeout-handling code is expected to make progress.

Observed 14 times during a 2-hour `high-throughput-test.sh` run against the test deployment. Not blocking for every workload, but a recurring cause of consumer disconnects under sustained high-throughput load with aggressive retention.

## Affected code

https://github.com/amazon-mq/rabbitmq-stream-s3/blob/main/src/rabbitmq_stream_s3_remote_reader.erl#L348-L363

```erlang
{Transient, _} when
    Transient =:= stream_error;
    Transient =:= connection_error;
    Transient =:= timeout
->
    maybe_reply_pending(State);
```

The `slow_down` / `internal_error` branch above it uses `erlang:send_after(RetryDelay0, self(), retry_requests)` to schedule a retry. The branch covering `timeout` (and `stream_error`, `connection_error`) relies on `maybe_reply_pending` to either answer the pending read or return `{retry, State}` from `maybe_reply`, which in turn schedules `retry_requests` after 1 second.

In the observed traces, neither happens.

## Observed timeline

From node2 of `env-2026-04-16`, reader pid `<0.79169.0>`, during a 2-hour high-throughput-test on 2026-04-23:

```
19:51:36  rabbitmq_stream_s3_remote_reader transitioning to next fragment (...11384594 -> ...11442534)
19:51:38  Fragment ...11442534 was deleted by retention while being read. Jumping to oldest available fragment ...11508132
19:52:08  S3 request timed out on <0.80531.0>/#Ref<0.689357524.4206100482.14274>
          (30-second silence, no log output from <0.79169.0>)
19:52:38  rabbit_stream_reader terminating in state 'open' with reason '{timeout,{gen_server,call,[<0.79169.0>,{read,58,...},60000]}}'
```

Stack trace from the terminating `rabbit_stream_reader`:

```
exception exit: {timeout,
                    {gen_server,call,
                        [<0.79169.0>,{read,58,303,chunk_boundary},60000]}}
  in function  gen_server:call/3 (gen_server.erl, line 1222)
  in call from rabbitmq_stream_s3_remote_reader:read/5 (rabbitmq_stream_s3_remote_reader.erl, line 186)
  in call from rabbitmq_stream_s3_log_reader:read/4 (rabbitmq_stream_s3_log_reader.erl, line 793)
  in call from rabbitmq_stream_s3_log_reader:read_header1/1 (rabbitmq_stream_s3_log_reader.erl, line 519)
  in call from rabbitmq_stream_s3_log_reader:send_file/3 (rabbitmq_stream_s3_log_reader.erl, line 297)
  in call from osiris_log_reader:send_file/3 (src/osiris_log_reader.erl, line 86)
  in call from rabbit_stream_reader:send_chunks/8 (rabbit_stream_reader.erl, line 3797)
```

## Reproduction

- 2-hour `high-throughput-test.sh` against `env-2026-04-16` on 2026-04-23: 14 occurrences across three nodes (6 on node0, 2 on node1, 6 on node2)
- 45-minute `high-throughput-test.sh` on 2026-04-23 earlier in the day: 2 occurrences
- 45-minute `high-throughput-test.sh` against the same deployment on 2026-04-22: multiple occurrences prior to #131 being fixed, though those were masked by the `unimplemented` crash

Workload: 6 streams, 12 producers, 12 consumers, `max-length-bytes` retention active. The timeouts tend to cluster near retention jumps but also occur independently of them.

## Expected behaviour

After a request times out, the reader should either:

- Schedule a retry of the pending read within a bounded delay (e.g. 1 second, consistent with `?MIN_RETRY_DELAY_MS`), or
- Reply to the pending read with an error so the caller can decide how to proceed

It should never remain idle long enough for the outer 60-second `gen_server:call` to give up.

## Analysis

The load-bearing assumption in the current code is that `maybe_reply_pending` will drive progress after a timeout. Tracing through the code suggests this should work: `maybe_reply_pending` calls `maybe_reply`, which calls `try_read/3`, which returns `{await, State}` because `info = undefined` after the retention-induced reset. `maybe_start_request` then starts a new request (or, if the pool is busy, `maybe_reply` returns `{retry, State}` which schedules `retry_requests`).

The observed silence for 30 seconds contradicts this. Something in the post-timeout state prevents either branch from running the expected code. Candidates to investigate:

- Whether `pending_read` is actually set at the moment `handle_request_error` is called for the timeout
- Whether `maybe_start_request` is silently failing and `maybe_reply` is taking an unexpected path
- Whether the gun connection or pool is in a state that makes `get_range_async` hang rather than return `{ok, ...}` or `{error, pool_busy}`
- Whether a subsequent message (for a different request) drives the reader into a state where `pending_read` is cleared without being answered

Targeted logging at the entry to `handle_request_error` and the exit of `maybe_reply_pending` would disambiguate.

## Related

- #122, #130: mailbox flood fix, independent of this issue
- #131, #132: `unimplemented` crash fix, independent of this issue (this bug is pre-existing on `main` and remains after #132 is merged)

### LINKED ISSUE #135: Reader killed by 60s caller timeout when retention jumps combine with S3 request timeouts
## Summary

Under sustained high-throughput load with aggressive retention, a remote reader can be killed by its caller's 60-second `gen_server:call` timeout when retention deletes fragments faster than the reader can chase them, and one or more S3 request timeouts consume the remaining budget.

This is a different failure mode from #133. The fix for #133 (close the connection on S3 request timeout to avoid HTTP/1.1 head-of-line blocking) is confirmed working in this run. What remains is a longer-running race where a reader falls far behind, retention continuously invalidates its target fragment, and the reader exhausts the caller's budget before it can return any data.

## Symptoms

- Caller (`rabbit_stream_reader`) terminates with `{timeout,{gen_server,call,[<Reader>,{read,58,...},60000]}}`
- Offset `58` in the call indicates the caller is requesting the start of a new fragment (just past the 56-byte fragment header)
- In the minute before the crash, the reader log shows a long cascade of `Fragment NNN was deleted by retention while being read. Jumping to oldest available fragment MMM` warnings
- One or more `S3 request timed out on <ConnPid>/<Ref>` warnings interleave with the retention jumps

## Environment

- Branch: `fix/gh-133-reader-stall` at `5add5c0` (PR for #133)
- Test: high-throughput workload, 6 streams x 2 producers x 2 consumers, 2 hours
- Observed 5 occurrences on one node; 0 on the other two nodes

## Confirmation that #133 is fixed

In the same run, readers that hit S3 request timeouts recover in 10-20 seconds now (close connection, pool opens fresh one, new request succeeds). Pre-fix, these would have stalled for 30 seconds. Example: reader `<0.30833.0>` hit S3 timeouts at 14:08:11, 14:08:56, 14:09:45, 14:11:11, and 14:11:41, and successfully transitioned to the next fragment after each of the first four - the fifth happened to coincide with the caller's 60-second deadline.

## Readers that survived heavy retention alone

Two readers on the same node hit 675 and 462 retention jumps respectively over the test and never crashed. Retention jumps alone are survivable. The killer is retention + S3 timeout within the caller's 60-second window.

## Deep dive

<details>
<summary>Detailed timeline analysis and remediation directions</summary>

### Test configuration

- Deployment: 3 nodes, remote tier backed by S3
- Workload: `stream-perf-test` with 6 streams, 2 producers and 2 consumers per stream, 2-hour duration
- Retention: aggressive enough to delete fragments in the middle of reads. Peak of 950 S3 objects, final of 950 (retention did not shrink the count, but it did rotate fragments aggressively)

### Affected readers on the failing node

Five readers crashed with `gen_server:call` timeout, all with call signature `{read, 58, 303, chunk_boundary}`:

| Reader | Timestamp | Retention jumps before crash | S3 timeouts before crash |
|--------|-----------|------------------------------|---------------------------|
| `<0.16562.0>` | 13:04:40 | 4 | 2 |
| `<0.16952.0>` | 13:05:02 | 13 | 1 |
| `<0.19551.0>` | 13:17:51 | 13 | 1 |
| `<0.19580.0>` | 13:19:53 | 13 | 1 |
| `<0.30833.0>` | 14:12:11 | 5 | 5 |

### Representative timeline for `<0.16562.0>`

```
13:02:38.563  transitioning to next fragment (41676793 -> 41741628)
13:03:23.048  S3 request timed out on <0.20431.0>     (fix for #133 activates)
13:03:23.051  Fragment 41741628 was deleted by retention. Jumping to 43048401
13:03:33.489  Fragment 43048401 was deleted by retention. Jumping to 43305944
13:03:37.216  Fragment 43305944 was deleted by retention. Jumping to 43446546
13:03:40.150  Fragment 43446546 was deleted by retention. Jumping to 43512211
13:04:10.151  S3 request timed out on <0.20933.0>
13:04:40.153  caller <0.882.0> terminates: gen_server:call timeout to <0.16562.0>
```

The fix for #133 worked at 13:03:23 - the stalled S3 request is abandoned, connection closed, retention jump proceeds. The reader then chases through four deleted fragments in 17 seconds. The S3 request on the fifth jumped-to fragment times out 30 seconds later. The caller's 60-second budget (which started when it issued `{read,58,...}`) expires exactly 30 seconds after that.

Fragment number delta across the jumps for this reader: 41741628 to 43512211, about 1.77 million entries in 17 seconds.

### Representative timeline for `<0.19580.0>`

```
13:18:11 to 13:18:53  13 retention jumps in 42 seconds
  (65233779 -> 65366160 -> 65438950 -> 65506892 -> 65554787 ->
   65620207 -> 65752392 -> 65890783 -> 65956643 -> 66021457 ->
   66076319 -> 66207929 -> 66274441 -> 66405865)
13:19:23.434  S3 request timed out on <0.30121.0>
13:19:53.446  caller <0.19576.0> terminates: gen_server:call timeout to <0.19580.0>
```

Fragment number delta: 65233779 to 66405865, about 1.17 million entries in 42 seconds. The caller budget is consumed entirely by retention-jump chasing plus one 30-second S3 timeout.

### Readers that survived

On the same node, same test run:

- `<0.1329.0>`: 675 retention jumps, no crash, still healthy at end of test (last activity 14:41:41)
- `<0.17098.0>`: 462 retention jumps, no crash, still healthy at end of test (last activity 14:41:48)

Retention jumps by themselves are not fatal. The combination that kills is: retention jumps consuming the first half of the caller's budget, then an S3 timeout (30 seconds) consuming most of the second half, leaving no time to reply.

### Load-bearing observations

- Caller's `gen_server:call` timeout is 60 seconds (in `rabbit_stream_reader`)
- `REQUEST_TIMEOUT_MS` for a single S3 request is 30 seconds
- A retention jump typically costs between 200 ms and 10 seconds of real time (the reader must detect the 404, then list or look up the oldest available fragment, then re-issue the request)
- Two worst-case serial events (one S3 timeout plus one slow retention jump) easily exceed 30 seconds
- In the traces above, the caller's 60-second budget is consumed by one S3 timeout (30 s) plus retention-jump chasing (approaching 30 s of cumulative wall-clock delay)

### Directions to investigate

These are candidate approaches, not proposed fixes. Each needs analysis.

1. Reduce retention-jump cost. Currently each jump issues a fresh S3 GET on the jumped-to fragment. If that fragment is also deleted, this GET returns 404 and the reader jumps again. Caching or batch-listing oldest-available fragment info per stream could short-circuit several jumps.
2. Reset the caller budget when the reader makes measurable forward progress. If a retention jump establishes a new oldest fragment, arguably the caller's 60-second clock should reset, because the previous target is no longer relevant.
3. Cap retention-jump retries per caller request. After N jumps the reader replies with `{error, retention_overtake}` or similar and lets the caller decide.
4. Reduce `REQUEST_TIMEOUT_MS` when retention pressure is high. A 30-second S3 timeout is expensive compared to the caller's 60-second budget. A shorter timeout (10 s or 15 s) with retries inside the reader would let more work fit inside the caller's window, at the cost of spurious timeouts on a genuinely-healthy but slow S3.
5. Have the reader tell the caller when it is making progress. A heartbeat or intermediate reply would let the caller extend its deadline. Larger change.

### Not proposed

Do not simply increase the caller's `gen_server:call` timeout above 60 seconds. That masks the underlying throughput-vs-retention imbalance and makes the system less responsive to real stalls.

</details>

## Related

- Fixed in the same code path: #133 (HTTP/1.1 HOL blocking after S3 request timeout)

lukebakken:
## Summary

On S3 request timeout, `rabbitmq_stream_s3_remote_reader` called `gun:cancel(Conn, StreamRef)` and checked the connection back into the pool. In an HTTP/1.1 client, `gun:cancel` does not cancel the stream on the wire - it only stops forwarding events to the owner. The cancelled stream continues to occupy the connection until its full response is received.

When the reader then started a new request on the same connection (often taken back from the pool), the new request pipelined after the cancelled one and could not complete until the cancelled response fully arrived. If the original request was stalled on the S3 side, the new request stalled too, causing the reader's pending read to miss its 60-second `gen_server:call` deadline and the consumer connection to terminate.

The fix closes the whole connection on timeout. The pool observes the `'DOWN'` message, removes the dead connection, and opens a fresh replacement. The next `try_checkout` returns a healthy connection.

Fixes #133.

Does not fix #135. A separate failure mode exists where retention deletes fragments faster than the reader can chase them; that is tracked in #135.

## Testing

Two-hour high-throughput load runs (6 streams, 2 producers and 2 consumers per stream) before and after the fix.

Before the fix (branch at `d9370c8`): 11 `gen_server:call` timeout crashes across 3 nodes in 2 hours.

After the fix (this PR): 0 crashes on 2 of 3 nodes. 5 crashes on 1 node all attributable to the separate retention-chase issue tracked in #135, not to HTTP/1.1 HOL blocking. In the same run, readers recovered from S3 request timeouts in 10-20 seconds (close connection, pool grows a new one, new request succeeds). Pre-fix, these stalled for 30 seconds and usually killed the consumer.

Evidence that the fix works under the observed failure path: reader `<0.30833.0>` on one node hit five S3 request timeouts in four minutes and recovered successfully after the first four (visible via `transitioning to next fragment` log lines immediately following each timeout). Only the fifth timeout coincided with the caller's 60-second deadline, and that interaction is the `#135` case.

## What this does not change

- `cancel_async/2` still uses `gun:cancel` rather than `gun:close`. It is called during reader termination and retention jumps, and the minimal fix is scoped to the `request_timeout` path. If the same HOL-blocking argument applies there and causes measurable issues, a follow-up can extend the same pattern.

## Deep dive

<details>
<summary>Gun HTTP/1.1 cancel semantics, pool interaction, and fix walkthrough</summary>

### Why `gun:cancel` does not cancel on HTTP/1.1

From `deps/gun/src/gun_event.erl` (verbatim):

> In the case of HTTP/1.1 we cannot actually cancel the stream, we only silence the stream to the user. Further response events may therefore be received and they provide a useful metric as these canceled requests monopolize the connection.

From `deps/gun/src/gun_http.erl`:

```erlang
%% We can't cancel anything, we can just stop forwarding messages to the owner.
cancel(State0, StreamRef, ReplyTo, EvHandler, EvHandlerState0) ->
    case is_stream(State0, StreamRef) of
        true ->
            State = cancel_stream(State0, StreamRef),
            ...
```

`cancel_stream` flips `is_alive` to `false` on the stream but leaves it in the streams list. Gun still parses the server's response from the wire and discards it, but the stream continues to occupy its position in the pipeline.

### Pool does not share connections across concurrent callers

`rabbitmq_stream_s3_api_aws_pool`'s `take_available/2` removes a `Conn` from the `available` list and records it in `checkouts`. One caller at a time. The pool is not serving multiple concurrent requests on one connection.

### Why the fix works despite one-at-a-time-per-connection semantics

Even though only one caller holds a connection at a time, the HTTP/1.1 stream from the cancelled request is still on the wire. When the caller checks the connection back into the pool and either itself or a different caller checks it out again, the new request gets pipelined on the same TCP connection, behind the cancelled-but-still-open stream. Gun's HTTP/1.1 handler cannot begin the new stream's response until it consumes the previous stream's response in full.

If the previous stream was slow for a genuine reason (S3 returning slowly), the new stream waits for that full response too. The reader's pending-read caller sees no data arrive, and its 60-second `gen_server:call` deadline fires.

Closing the connection instead forces the pool to open a fresh TCP connection with no stale stream state.

### Interaction with the pool

`gun:close(Conn)` causes the connection process to exit. The pool monitors each connection via `erlang:monitor(process, ConnPid)`. When the `'DOWN'` message arrives:

```erlang
handle_info({'DOWN', _MRef, process, Pid, _Reason}, ...) ->
    case Monitors0 of
        #{Pid := MRef} ->
            %% A connection is down. Forget it and maybe grow a new one.
            Conn = Pid,
            State1 = ..., State2 = cancel(Pid, State1),
            {noreply, grow(State2)};
```

The pool drops the dead connection from its `monitors` map, cancels any pending checkouts waiting on it, and calls `grow/1` to open a replacement. Subsequent `try_checkout` calls return a healthy connection.

### Why the reader does not receive the dead connection on the next `try_checkout`

The dead connection is in `checkouts` (the caller had it checked out), not `available`. `try_checkout` only draws from `available`. Even if the pool has not yet processed the `'DOWN'` by the time the reader calls `try_checkout` on the next request, the reader gets a different healthy connection.

### Code change

In `rabbitmq_stream_s3_api_aws.erl`:

```erlang
handle_async(
    {request_timeout, StreamRef},
    StreamRef,
    #{conn := Conn, stream_ref := StreamRef} = State
) ->
    ?LOG_WARNING("S3 request timed out on ~tw/~tw", [Conn, StreamRef]),
    counters:add(counter(), ?C_REQUEST_TIMEOUTS, 1),
    gun:close(Conn),
    finish_async_close(State),
    {done_cancel, {error, timeout}}.
```

`finish_async_close/1` is a new helper that does the same bookkeeping as `finish_async/1` (cancels the request timer, drains any late `request_timeout` message, decrements `?C_ACTIVE_REQUESTS`) but omits the pool checkin. The pool's `'DOWN'` handler will clean up the now-dead connection.

### Test evidence that #135 is a different issue

Before this PR, every `gen_server:call` timeout was preceded by 30 seconds of complete reader silence after an S3 request timeout. After this PR, readers continue to log `transitioning to next fragment` or `Fragment NNN was deleted by retention` messages right up until the caller's deadline fires. The failure mode has changed. The remaining 5 crashes (on one node, none on the other two) all involve aggressive retention-jump chasing consuming most of the caller's budget before an S3 request timeout consumes the remainder. That behavior is tracked in #135.

</details>

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -632,8 +632,13 @@ handle_async(
 ) ->
     ?LOG_WARNING("S3 request timed out on ~tw/~tw", [Conn, StreamRef]),
     counters:add(counter(), ?C_REQUEST_TIMEOUTS, 1),
-    gun:cancel(Conn, StreamRef),
-    finish_async(State),
+    %% gun cannot cancel an HTTP/1.1 stream on the wire - it only stops
+    %% forwarding events to the owner. The cancelled stream continues to
+    %% occupy the connection, blocking subsequent requests until its full
+    %% response is received. Close the whole connection instead; the pool
+    %% observes the 'DOWN' and opens a fresh replacement.
+    gun:close(Conn),
+    finish_async_close(State),
     {done_cancel, {error, timeout}}.
 
 -spec cancel_async(async_req(), async_state()) -> ok.
@@ -678,6 +683,27 @@ finish_async(#{conn := Conn, pool := Pool} = State) ->
     end,
     counters:sub(counter(), ?C_ACTIVE_REQUESTS, 1),
     ok = rabbitmq_stream_s3_api_aws_pool:checkin(Pool, Conn).
+
+%% Finish an async request when the connection is being closed (e.g. on
+%% timeout). Same bookkeeping as `finish_async` but omits the pool checkin -
+%% the pool's 'DOWN' handler removes the conn and grows a replacement.
+finish_async_close(State) ->
+    case State of
+        #{timer_ref := TimerRef, stream_ref := StreamRef} ->
+            case erlang:cancel_timer(TimerRef) of
+                false ->
+                    receive
+                        {request_timeout, StreamRef} -> ok
+                    after 0 -> ok
+                    end;
+                _ ->
+                    ok
+            end;
+        _ ->
+            ok
+    end,
+    counters:sub(counter(), ?C_ACTIVE_REQUESTS, 1),
+    ok.
 
 -spec request(http_method(), key(), req_headers(), iodata(), request_opts()) ->
     {ok, http_response()}


the-mikedavis (approved):
Makes sense 👍

(I wonder why S3 doesn't support HTTP/2 where these semantics are a bit more defined? I guess we could go around asking internally 😅)

PR #148, opened 2026-05-19
Redesign upload path to behave like replication

### LINKED ISSUE #34: Backfill uploads of older stream data to the remote tier
Currently when a writer starts, it reads through the active (latest) segment and recovers information about fragments which may be uploaded to the remote tier. If the remote tier has a long outage, the older segments might also have fragments which should be uploaded as well. The manifest server should recognize when there is a 'hole' between the last fragments recorded in the manifest and the current segment and backfill the upload of older fragments from older segments. To do this it doesn't need to block the writer - it can kick off a background task which scans through the index files and sends `#fragment_available{}` notifications back to the manifest server.

Implementing this would also make it possible to use this plugin on a broker which has existing stream data: the backfill process would work through all existing data, 'converting' the stream to a tiered-storage stream.

### LINKED ISSUE #46: Handle stream data loss
Streams don't provide the same strong data durability guarantees as QQs/Raft because they don't ever [`fsync(2)`](https://www.man7.org/linux/man-pages/man2/fsync.2.html) segment or index files. Raft guarantees durability of confirmed writes as long as a quorum of replicas who flushed the write survive (and only if the flush actually makes the data durable, i.e. it's hardware dependent). Without flushing, streams replicas can lose data in hard crashes like uncontrolled power-offs.

When replicas are used, the chances of data loss becomes slimmer but not impossible. Consider three nodes: A at offset 100, B at 100 and C at 75 (behind on replication for whatever reason). If B has a hard crash (e.g. power-off) and comes back and, at the same time, A becomes partitioned, B and C form a majority that can elect a new writer who's commit offset is 75 instead of 100. So the originally confirmed writes of offsets 76..100 are lost. Then B and C can write new records in that range of offsets.

This is handled gracefully by Osiris. When A starts again as a replica and rejoins the cluster, it will truncate its log down to 75 where the higher-epoch new writer started writing new entries. We need to consider this kind of data loss in this plugin too so that a data loss scenario is correctly handled (i.e. we overwrite data) rather than the plugin crashing from failed assumptions.

### LINKED ISSUE #76: Server misses streams when rebooting after segfault
In a long-running test I have three streams publishing in a three node cluster. One node has segfaulted twice and when it does so and is automatically restarted, `rabbitmq_stream_s3_server` does not track all of the stream replicas. The boot step which starts the server might not have the correct dependencies to always start before the stream coordinator.

### LINKED ISSUE #93: Identify new metrics to capture
Following from what we capture and expose now (#88), what new metrics would be interesting to capture? We should think of metrics we'd like to see that would help detect and explain operational issues.

### LINKED ISSUE #140: Refactor upload path to behave like replication
Currently the `osiris_log_manifest` callback implementation gathers information about written chunks and decides when to cut a fragment. Then it passes the fragment info to `rabbitmq_stream_s3_server`. The decision to 'cut' a fragment (decide which chunks become a fragment) is rigid and brittle: either after enough bytes of chunk data have accumulated (hard-coded, can't change later) or when a segment rolls. The `osiris_log_manifest:handle_event/2` callback also seems kind of hacky to me.

I think the better abstraction is for the upload path to mirror `osiris_replica_reader`: we have a `gen_server` getting notifications that more data is available, and then it can upload to the remote tier whenever it sees fit. It would be a sort of non-voting replica. The main difference is that it would be operating on chunks - it wouldn't have a concept of a segment. It would subscribe to commit offset updates from the writer, read committed chunk headers and once enough data has accrued or when a timer fires it would then make an upload.

This introduces a big complication: writers of different epochs could cut a fragment at the same offset with different sizes. To guard against that causing issues, we need to add UIDs to fragments like we use for the manifest objects. That breaks the nice "use the next-offset pointer from the fragment header" utility we have now and makes the remote reader more complicated. It needs to do a range-read on the manifest tree to know the next set of fragments to read. But I think this is much more resilient to failure modes like data loss. It also opens the door for 'compacting' smaller fragments into one larger fragment, if we want to do that eventually. And if the remote reader knows the array of fragments it can download the next-next fragment and next-next-next fragment ahead of time to increase throughput.

This change will be a very large refactor / rewrite but it's a much better abstraction and it should let us really cut down on (or maybe eliminate) `osiris_log_manifest`.

### LINKED ISSUE #141: Mix user-configured retention with retention from offloading data
This depends on #140 because #140 enables gaps in the remote tier and this new strategy can create gaps.

Currently we overwrite local tier retention by setting the `retention` key in `osiris:config()` to just the `{'fun', fun()}` retention spec that deletes segments which have been fully uploaded. The user-configured retention settings become the retention settings for the remote tier. Consider a prolonged object store outage. With this strategy, the local tier will certainly fill up under load even if a single stream has a relatively short retention set like 5 GiB.

The current behavior of the retention spec list (`[osiris:retention_spec()]`) is that the most aggressive retention will take precedence. For example if max-bytes has not been hit but max-age has, segment(s) will be deleted to satisfy max-age. So instead of overwriting the user-configured retention specs with the `{'fun', fun()}` spec which deletes data which has been offloaded to the remote tier, we can add that spec to the list.

The main consideration is that this will create a gap in the remote tier log during an object storage outage. That's ok though, as the local tier log can effectively have a gap in vanilla streaming when retention reclaims a segment which a read is currently reading. That behavior is defined and tested in Osiris. The reader skips ahead to the next earliest available data. Gaps are skips.

### LINKED ISSUE #144: Batch object deletions
We can use a 'sink' gen_batch_server to accept casts to delete objects. S3 has a multi-object deleting operation which we already use for stream deletion. We can have the gen_batch_server opportunistically delete many objects at once on a busy node. This would cover old manifests deleted after a new manifest is updated as well as fragments and groups deleted by retention. With some refactoring it could cover all objects deleted when a stream is deleted as well.

the-mikedavis:
This is the massive redesign from #140. The remote tier becomes a kind of non-voting/observer replica and the plugin's role becomes entirely a background process. Sorry for the massive PR and messy commit history :sweat_smile:

Fixes #140
Fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/144
Fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/141
Fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/46
Fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/76
Fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/34
Fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/93

## Why

The previous upload path had design issues which put us into a corner for future extensibility.

1. No bounded time-to-S3. Fragments uploaded only when a size threshold was crossed or a segment rolled. A low-throughput stream could take minutes or hours before committed data reached S3, making disaster recovery and compliance guarantees impossible. For example if you wanted to build a disaster recovery mechanism using tiered streams, you had no way to ensure that the last N minutes of data would be persisted.
2. Writer hot path carried upload work. The `osiris_log_manifest` callback ran on every chunk write, passing iodata through for CRC and size tracking. This added latency to the write path for work that belongs elsewhere.
3. Upload state split across two processes. Fragment assembly lived in the writer (via `osiris_log_manifest`), while scheduling and manifest application lived in `rabbitmq_stream_s3_server`. The split forced complex out-of-order upload handling that a single owner would not need.

## What changed

A per-stream remote replica reader process now owns the entire upload lifecycle: draining committed chunks, assembling fragments, uploading to S3, persisting the manifest, broadcasting to replicas, and evaluating retention. The writer's hot path has zero plugin overhead.

Key architectural changes:

- `rabbitmq_stream_s3_server` and `rabbitmq_stream_s3_machine` eliminated.
- `osiris_log_manifest` behaviour dropped. Replaced by `osiris_log_hooks` (lifecycle-only callbacks, no per-chunk work).
- Per-node governor paces S3 transfers via a token bucket. No fixed concurrency limit.
- Functional core / imperative shell split in the replica reader enables deterministic testing of all state machine interleavings without processes or mocks. (Same approach as the old server / machine split.)
- Fragment objects reduced from 58-byte metadata header to 8-byte magic+version. All metadata lives in the manifest.
- Fragment keys include a 32-bit UID, eliminating the deposed-writer overwrite race.
- Manifest entries grow from 30 to 34 bytes (UID field added).
- Fragment iterator hides manifest tree structure from the read path.
- The plugin is now a completely background process. The plugin can be enabled/disabled at runtime.

## Testing

- `replica_reader_core_SUITE`: pure functional tests for all state machine interleavings (out-of-order completions, persist triggers, failures, rebalancing).
- `replica_reader_SUITE`: integration tests with real osiris writer and FS backend.
- `fragment_iterator_SUITE`: manifest tree traversal without S3.
- `broker_SUITE`: end-to-end with a running broker (upload, deletion, cross-tier reads, writer restart).
- `log_reader_SUITE`, `prop_SUITE`: read path and property-based tests.

Total coverage is now at 57%, with much of the core logic covered at 100%.


the-mikedavis:
This builds on some newer changes in Osiris to replace `osiris_log_manifest` with a thin set of lifecycle hooks (`osiris_log_hooks`). https://github.com/rabbitmq/osiris/pull/196#issuecomment-4488799121

Another notable part of this change: `streams_s3.log_reader` and `streams_s3.log_manifest` are no longer set in configuration. The necessary variables are set and unset during plugin enable/disable. Configs setting these values will need to be updated to remove them.

the-mikedavis:
I'm going to merge this even though I'm not super confident that it isn't hiding bugs. I think it's better to handle findings/review as follow-ups. And I want to start referencing the documentation and making unrelated follow-up changes now.

PR #152, opened 2026-05-20
replica reader: Replace spawn_link with spawn_monitor for group uploads

### LINKED ISSUE #151: Group-upload spawn_link can permanently deadlock the replica reader
## Summary

The group-upload task in `rabbitmq_stream_s3_replica_reader` uses `spawn_link` (line 659), which can permanently deadlock the replica reader if the spawned process is killed externally. Once deadlocked, no further manifests are persisted to Khepri - the stream appears to stop growing for remote-tier consumers, and fragments uploaded after the last successful persist become orphaned in S3.

## Mechanism

Three facts combine:

1. The replica reader traps exits (`process_flag(trap_exit, true)` at line 251), so linked process deaths arrive as `{'EXIT', Pid, Reason}` messages rather than killing the reader.
2. The catch-all `handle_info(_Info, State)` at line 416 silently discards those messages.
3. The core state's `rebalance_in_flight` flag is only cleared by receiving `{group_upload_result, ...}` - via `group_upload_complete` (core line 306) or `group_upload_failed` (core line 349). No other code path resets it.

The `try/catch` inside the spawned fun covers Erlang-level exceptions, but not external kills (OOM killer, `exit(Pid, kill)` from a supervisor or memory alarm handler). In those cases no `{group_upload_result, ...}` message is sent, the exit signal hits the catch-all and vanishes, and `rebalance_in_flight` stays `true` forever.

## Consequence

Once `rebalance_in_flight` is stuck `true`:

```erlang
maybe_start_persist(#state{rebalance_in_flight = true} = State) ->
    {State, []};  %% core line 467 - persist permanently blocked
```

- Replicas never learn about new fragments (manifest broadcasts happen at persist time)
- On node restart, all fragments uploaded since the last persist are orphaned in S3
- `bytes_in_persist` gauge grows without bound
- Consumers reading from the remote tier see the stream stop growing

## Suggested fix

Replace `spawn_link` with `spawn_monitor` and add a `{'DOWN', ...}` handler that calls `group_upload_failed`, matching the pattern already used for the persist task (lines 679 + 406):

- On `{group_upload_result, ...}` - `demonitor(Ref, [flush])`, proceed normally
- On `{'DOWN', Ref, ...}` matching the group monitor - treat as `group_upload_failed` with a non-retriable reason

This clears `rebalance_in_flight`, allowing persist to proceed. The next `drain_completions` cycle will re-detect the rebalance threshold and retry.

Plain `spawn` (as used for `delete_manifest_objects` at line 499) removes the exit-signal hazard but does not solve the liveness problem - if the process dies without sending the result message, `rebalance_in_flight` is still stuck forever.

lukebakken:
## Summary

- Replace `spawn_link` with `spawn_monitor` for the group-upload task in the replica reader
- Add a `'DOWN'` handler that routes to `group_upload_failed`, clearing `rebalance_in_flight` so persists can resume
- Matches the pattern already used for the persist task (lines 679 + 406)

## Problem

The group-upload task uses `spawn_link`, but the replica reader traps exits and the catch-all `handle_info` silently discards `EXIT` signals. If the spawned process is killed externally (OOM, `brutal_kill`), no `group_upload_result` message is sent, leaving `rebalance_in_flight` stuck `true` forever and blocking all subsequent manifest persists.

Consequences: replicas never learn about new fragments, fragments uploaded after the last persist are orphaned in S3, and consumers see the stream stop growing.

## Test plan

- [ ] Verify compilation (done - compiles cleanly)
- [ ] Integration test: start a stream with low `rebalance_threshold`, publish enough to trigger group upload, kill the group-upload process externally, verify the replica reader recovers and persists continue

Closes #151.

### DIFF
--- a/src/rabbitmq_stream_s3_replica_reader.erl
+++ b/src/rabbitmq_stream_s3_replica_reader.erl
@@ -223,6 +223,8 @@ returned by the functional core module.
     %% persist_pending_bytes at start_persist; cleared on persist_result.
     %% bytes_in_persist gauge = persist_pending_bytes + persisting_bytes.
     persisting_bytes = 0 :: non_neg_integer(),
+    %% Monitor ref for the in-flight group upload task.
+    group_mon :: reference() | undefined,
     %% Kind of the group upload currently in flight. Cleared when the
     %% group_upload_result message arrives. Only one rebalance is in
     %% flight at a time so a single slot suffices.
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
Looks good 👍

I thought we could get away with the spawn_link since we `try`/`catch` but I agree, a monitor is the right way to do this.

the-mikedavis:
I'm not sure we necessarily need an integration test case for this. I think the fix is straightforward enough on its own

PR #153, opened 2026-05-20
Improve testing and core operations

the-mikedavis:
This is working towards #36 gradually. I wanted to submit this hodge-podge of changes for you to look at @lukebakken since our work might overlap a little bit, like with https://github.com/amazon-mq/rabbitmq-stream-s3/pull/152.

Mainly this adds tests. A few other things:

1. `make fmt` and `make fmt-check`. Conveniences over `erlfmt`.
2. Refactors for the replica reader core so that retention is performed asynchronously when there are group objects. If there are group objects we need to download objects from the remote tier to know what to delete, so it shouldn't block the replica reader.
3. Along with that, fix the edit ordering. There were some bugs with that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,16 @@ include ../../erlang.mk
 
 CT_QUICK_SUITES = api_fs db replica_reader_core replica_reader log_reader fragment_iterator fragment_assembly prop manifest_replica
 
+ERLFMT ?= erlfmt
+ERLFMT_FILES = src/*.erl test/*.erl
+
+.PHONY: fmt fmt-check
+fmt:
+	$(verbose) $(ERLFMT) -w $(ERLFMT_FILES)
+
+fmt-check:
+	$(verbose) $(ERLFMT) -c $(ERLFMT_FILES)
+
 .PHONY: ct-quick
 ct-quick: test-build
 	$(verbose) mkdir -p $(CT_LOGS_DIR)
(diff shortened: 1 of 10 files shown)


lukebakken (approved):
Clean improvement. The `edits_since_persist` simplification eliminates a class of ordering bugs by construction, and the async group retention prevents blocking S3 downloads from stalling the replica reader's hot path. Test coverage is substantially improved - the property test (`broadcast_edits_with_retention`) and the `retention_and_rebalance_same_persist_cycle` interleaving test are particularly valuable.

Two minor follow-ups filed as issues:

- #155 - `edits_since_persist ++ [Edit]` is O(N^2) per persist cycle (trivial at current thresholds, standard prepend-and-reverse idiom would fix it)
- #156 - async retention task has no timeout, can block persist indefinitely if S3 hangs (same pre-existing pattern as the group upload path)

I pushed one small commit (51e9328) that binds the manifest in the `maybe_spawn_group_retention` function head instead of re-reading it from `State#state.core` - the value is identical since no mutation occurs between the two read sites.

the-mikedavis:
Nice, thanks!

The :genie: loves using `++/2`. It blows its mind a little bit when you suggest prepending and reversing to access it :roll_eyes:

the-mikedavis:
I've got some follow-up changes started to address those in a `testing-2` branch :+1:

PR #158, opened 2026-05-21
remote reader: Reply {error, timeout} instead of crashing the caller

### LINKED ISSUE #157: remote reader: Two consecutive S3 timeouts crash consumers via gen_server:call timeout
## Summary

The `rabbitmq_stream_s3_remote_reader` gen_server crashes consumers when S3 latency exceeds 30 seconds for two consecutive requests. The `timeout` error handler retries immediately without backoff, leaving zero margin between two 30-second request timeouts and the caller's 60-second `gen_server:call` budget.

## Mechanism

1. The log reader calls `gen_server:call(Server, #read{...}, 60000)` (`remote_reader.erl:189`). The 60s timeout starts on the caller.

2. The remote reader starts an async S3 request with a 30s `?REQUEST_TIMEOUT_MS` (`remote_reader.erl:455`, `api_aws.erl:799-800`).

3. If S3 does not respond within 30s, `handle_async` returns `{done_cancel, {error, timeout}}` (`api_aws.erl:588-601`), closing the gun connection.

4. `handle_request_error(timeout, ...)` (`remote_reader.erl:362-367`) calls `maybe_reply_pending(State)` directly - no delay, no backoff. This differs from `slow_down`/`internal_error` (lines 354-360) which use exponential backoff via `erlang:send_after`.

5. `maybe_reply_pending` -> `maybe_reply` -> `try_read` -> `maybe_start_current_request` starts a new 30s request immediately.

6. If the second request also times out at 30s, total elapsed is ~60s - exactly matching the caller's `gen_server:call` timeout. The caller exits with `{timeout, {gen_server, call, [Pid, {read, ...}, 60000]}}`.

7. The crash propagates to `rabbit_stream_connection_sup` which hits `reached_max_restart_intensity`, killing the entire stream connection (all consumers and producers on it).

## Evidence from 2-hour soak test

- 10,747 `"S3 request timed out"` warnings on node0 (sustained 18-80/minute for the entire run)
- 119 `reached_max_restart_intensity` events on `rabbit_stream_connection_sup`
- Consumer rate dropped from ~400k msg/s to 0 at ~27 minutes into the test
- All crash stack traces show `rabbitmq_stream_s3_remote_reader:read/5` -> `gen_server:call/3` timeout
- Replay after the test consumed only 984 of 1,429,916 messages (0%)

## The design comment acknowledges the constraint

Lines 36-38 of `remote_reader.erl`:

```
%% Cancel and retry an in-flight S3 request if no response arrives within
%% this window. Must be less than ?GEN_SERVER_CALL_TIMEOUT (60s) to allow
%% recovery before the caller times out.
```

This assumes one timeout + one successful retry within 60s. Two consecutive timeouts break the invariant.

## Suggested fix

Add backoff to the `timeout` case in `handle_request_error`, matching the pattern used for `slow_down`/`internal_error`:

```erlang
{Transient, _} when
    Transient =:= stream_error;
    Transient =:= connection_error;
    Transient =:= timeout
->
    erlang:send_after(RetryDelay0, self(), retry_requests),
    RetryDelay = min(RetryDelay0 * 2, ?MAX_RETRY_DELAY_MS),
    maybe_reply_pending(State#?MODULE{retry_delay = RetryDelay});
```

This alone does not fix the fundamental 30s + 30s = 60s race. Additional options:

- Reduce `?REQUEST_TIMEOUT_MS` to 15s (allows 3 retries within 60s)
- Increase `?GEN_SERVER_CALL_TIMEOUT` (e.g. 120s)
- Reply with an error to the caller after N consecutive timeouts instead of holding `pending_read` indefinitely
- Combine: reduce request timeout AND add backoff, giving multiple retry attempts with margin

## Related

- #156 (async retention task has no timeout - same class of liveness issue)

### LINKED ISSUE #161: remote reader: Buffer compaction after deadline reply causes assertion failure on retry
## Summary

After the remote reader replies `{error, timeout}` to a pending read, the caller (`rabbit_stream_reader`) reverts the consumer to its pre-`send_chunks`-loop state. The log reader retries `read(Pid, Position, ...)` at a position from before the loop, but the remote reader's internal buffer has been compacted past that position, causing `?assert(Offset >= StartPos)` at `remote_reader.erl:759` to fail.

## Mechanism

1. `send_chunks` calls `send_file` in a recursive loop. Each successful iteration advances the remote reader's `current_pos` (via `try_read` returning `{ok, State#?MODULE{current_pos = Offset}, Data}`). The updated `Log` state is passed to the next recursion.

2. Eventually a `send_file` data-read times out. The 50-second `?PENDING_READ_DEADLINE_MS` fires, the remote reader replies `{error, timeout}`, and clears `pending_read`. The gen_server stays alive with in-flight async requests.

3. `send_chunks` returns `{error, Reason}` immediately. The caller (e.g. `handle_frame_post_auth`) returns the consumer **unchanged** - the consumer's `Log` state reverts to before the `send_chunks` loop started. All position advances from N successful `send_file` iterations are discarded.

4. Async S3 data arrives at the remote reader. `add_data_current` is called. Since `current_pos` (advanced by the successful reads in step 1) differs from `start_pos`, buffer compaction fires: `start_pos = current_pos`.

5. On retry (44ms later, same event loop), the log reader calls `read(Pid, OldPosition, ...)` where `OldPosition` is from before the loop. The remote reader's `start_pos` is now far past `OldPosition`. Assertion `Offset >= StartPos` fails.

## Evidence

From the 2-hour soak test with the `?PENDING_READ_DEADLINE_MS` fix deployed:

```
2026-05-21 17:15:02.980015+00:00 [info] <0.21114.0> Error while sending chunks: timeout
2026-05-21 17:15:03.024015+00:00 [error] <0.21114.0>     exception exit: {{{assert,
    [{module,rabbitmq_stream_s3_remote_reader},
     {line,759},
     {expression,"Offset >= StartPos"},
     {expected,true},
     {value,false}]},
    [{rabbitmq_stream_s3_remote_reader,try_read,3,...},
     {rabbitmq_stream_s3_remote_reader,maybe_reply,2,...},
     {rabbitmq_stream_s3_remote_reader,handle_call,3,...},
```

Same PID (`<0.21114.0>`) logs the graceful timeout handling then crashes 44ms later on the retry. The crash pattern `{read, 8, 303, chunk_boundary}` (Position = 8 = `?SEGMENT_HEADER_B`) indicates the log reader is retrying from the start of the fragment, but the remote reader's `current_pos` (and after compaction, `start_pos`) has advanced well past 8 from the successful reads in the `send_chunks` loop.

3 assertion failures observed in this run, each correlated with an "Error while sending chunks: timeout" event.

## Fix

When replying `{error, timeout}` in `maybe_reply_pending`, reset buffer state to position 0 and cancel in-flight requests. Setting `start_pos = 0` ensures any valid fragment offset passes the assertion. The empty buffer means the next read triggers a fresh async request from the caller's position.

## Related

- #157 (the timeout fix that introduced `?PENDING_READ_DEADLINE_MS`)
- #158 (the PR implementing the deadline)

### LINKED ISSUE #163: log reader: `index_data/4` crashes with badmatch when retention deletes a fragment during consumer attach
## Summary

`rabbitmq_stream_s3_log_reader:index_data/4` at line 832 uses a bare pattern match `{ok, Data} = rabbitmq_stream_s3_api:get_range(...)` that crashes with `{badmatch, {error, not_found}}` when a fragment has been deleted by retention. This kills the stream reader and the entire stream connection.

## Mechanism

1. A consumer subscribes to a stream at an offset that maps to a fragment in the remote tier
2. `find_position/3` calls `index_data/4` to fetch the fragment's index section from S3
3. Between the manifest lookup and the S3 fetch, retention deletes the fragment
4. `get_range` returns `{error, not_found}` (S3 404)
5. The bare match `{ok, Data} = ...` crashes with `{badmatch, {error, not_found}}`
6. The crash propagates through the linked log reader to `rabbit_stream_connection_sup`, hitting `reached_max_restart_intensity`

## Evidence

From a 2-hour soak test on a test deployment:

```
rabbit_stream_reader terminating in state 'open' with reason '{badmatch,{error,not_found}}'
** Stacktrace =
**  [{rabbitmq_stream_s3_log_reader,index_data,4,
        [{file,"rabbitmq_stream_s3_log_reader.erl"},{line,832}]},
     {rabbitmq_stream_s3_log_reader,find_position,3,
        [{file,"rabbitmq_stream_s3_log_reader.erl"},{line,731}]},
```

17 crashes on node2, all at the same instant (13:56:09) when retention ran and deleted fragments that consumers were trying to attach to.

## Suggested fix

Handle `{error, not_found}` in `index_data/4` by returning an error that `find_position/3` can propagate. The consumer should treat a deleted fragment the same way the remote reader treats a 404: check the manifest range and jump to the oldest available fragment.

## Related

- #157, #158 (timeout fixes that kept consumers alive long enough to hit this bug)
- PR #160 (allowed retention to delete all fragments, increasing the window for this race)

lukebakken:
## Summary

- Reduce `?REQUEST_TIMEOUT_MS` from 30s to 15s, allowing multiple retries within the caller's budget
- Add exponential backoff to the `timeout`/`stream_error`/`connection_error` error handler
- Add a 50-second `?PENDING_READ_DEADLINE_MS` that replies `{error, timeout}` instead of holding the pending read forever
- Reset buffer state (position 0, cancel requests) on deadline to prevent assertion failures when the caller retries at an old position
- Propagate `{error, timeout}` through `read_header` and `send_file` in the log reader
- Handle `{error, not_found}` in `index_data/4` to prevent badmatch crash when retention deletes a fragment during consumer attach

## Problem

Three bugs in the read path:

**Bug 1 (#157):** Two consecutive 30-second S3 request timeouts exhaust the caller's 60-second `gen_server:call` budget. The caller exits with `{timeout, {gen_server, call, ...}}`, crashing the log reader and the stream connection.

**Bug 2 (#161):** After replying `{error, timeout}`, the remote reader's buffer can be compacted by arriving async data (setting `start_pos = current_pos`). The caller retries at its pre-loop position (because `send_chunks` reverts the consumer on error), which is now behind `start_pos`, triggering `?assert(Offset >= StartPos)`.

**Bug 3 (#163):** `index_data/4` uses a bare pattern match `{ok, Data} = get_range(...)` that crashes with `{badmatch, {error, not_found}}` when retention deletes a fragment between the manifest lookup and the S3 fetch during consumer attach.

## Fix

Under sustained S3 latency, the remote reader retries with backoff (15s timeout, 1s/2s/4s/.../30s delays) until the 50-second deadline. At the deadline:

1. Reply `{error, timeout}` to the caller (instead of holding forever)
2. Reset buffer to position 0 and cancel in-flight requests (so any retry position passes the assertion)

The log reader propagates `{error, timeout}` through `send_file`, which the stream reader handles gracefully: logs the error and returns the consumer unchanged. On the next osiris offset event, delivery retries from the same position with a fresh buffer. No crash, no data loss, just a delivery stall until S3 recovers.

For the retention race (#163), `index_data/4` handles `{error, not_found}` and propagates it so the consumer can jump to the oldest available fragment instead of crashing.

## Test plan

- [x] Compiles cleanly (`make -C deps/rabbitmq_stream_s3`)
- [x] No trailing whitespace
- [ ] Deploy to test cluster and run 2-hour high-throughput soak test
- [ ] Verify: no assertion failures, no `case_clause` crashes, no `badmatch` crashes, no `reached_max_restart_intensity`; consumers survive S3 latency and recover

Closes #157. Closes #161. Closes #163.

### DIFF
--- a/src/rabbitmq_stream_s3_log_reader.erl
+++ b/src/rabbitmq_stream_s3_log_reader.erl
@@ -728,20 +728,26 @@ find_position(Spec, #manifest{} = Manifest, StreamId) ->
         find_fragment(Manifest#manifest.entries, Spec, GetGroupFun),
     %% Download the index from the fragment to find the exact chunk position.
     IdxStartPos = ?SEGMENT_HEADER_B + Size,
-    IndexData = index_data(StreamId, FragmentOffset, Uid, IdxStartPos),
-    {ChunkId, _, FragPos} = find_index_position(IndexData, Spec),
-    %% Position within the fragment object (after the 8-byte header).
-    Position = ?SEGMENT_HEADER_B + FragPos,
-    %% Create an iterator positioned at this fragment for forward navigation.
-    Iterator = rabbitmq_stream_s3_fragment_iterator:init(Manifest, FragmentOffset, GetGroupFun),
-    %% Advance past the current entry so the iterator is ready for `next`.
-    Iterator1 = advance_past_current(Iterator),
-    {ok, #remote_location{
-        chunk_id = ChunkId,
-        position = Position,
-        fragment_ref = FragRef,
-        iterator = Iterator1
-    }}.
+    case index_data(StreamId, FragmentOffset, Uid, IdxStartPos) of
+        {ok, IndexData} ->
+            {ChunkId, _, FragPos} = find_index_position(IndexData, Spec),
+            %% Position within the fragment object (after the 8-byte header).
+            Position = ?SEGMENT_HEADER_B + FragPos,
+            %% Create an iterator positioned at this fragment for forward navigation.
+            Iterator = rabbitmq_stream_s3_fragment_iterator:init(
+                Manifest, FragmentOffset, GetGroupFun
+            ),
+            %% Advance past the current entry so the iterator is ready for `next`.
+            Iterator1 = advance_past_current(Iterator),
+            {ok, #remote_location{
+                chunk_id = ChunkId,
+                position = Position,
+                fragment_ref = FragRef,
+                iterator = Iterator1
+            }};
+        {error, _} = Err ->
+            Err
+    end.
 
 -doc """
 Finds the manifest entry which contains the requested offset or timestamp.
@@ -829,12 +835,11 @@ saturating_decr(N) -> N - 1.
 index_data(StreamId, FragmentOffset, Uid, IdxStartPos) ->
     Key = rabbitmq_stream_s3:fragment_key(StreamId, FragmentOffset, Uid),
     ?LOG_DEBUG("Looking up key ~ts (~ts)", [Key, ?FUNCTION_NAME], ?DOMAIN),
-    {ok, Data} = rabbitmq_stream_s3_api:get_range(
+    rabbitmq_stream_s3_api:get_range(
         Key,
         {IdxStartPos, undefined},
         #{timeout => ?READ_TIMEOUT}
-    ),
-    Data.
+    ).
 
 %% Advance the iterator past the current entry (so next/1 returns the
 %% entry *after* the one we resolved to).
--- a/src/rabbitmq_stream_s3_log_reader.erl
+++ b/src/rabbitmq_stream_s3_log_reader.erl
@@ -349,7 +349,9 @@ send_file(
                             Err
                     end;
                 end_of_stream ->
-                    {end_of_stream, State0#?MODULE{mode = Remote1}}
+                    {end_of_stream, State0#?MODULE{mode = Remote1}};
+                {error, timeout} = Err ->
+                    Err
             end;
         {become_local, _} ->
             counters:add(counter(), ?C_REMOTE_CLOSE, 1),
@@ -360,7 +362,9 @@ send_file(
                     Err
             end;
         {end_of_stream, Remote} ->
-            {end_of_stream, State0#?MODULE{mode = Remote}}
+            {end_of_stream, State0#?MODULE{mode = Remote}};
+        {error, timeout} = Err ->
+            Err
     end;
 send_file(Socket, #?MODULE{mode = Local0} = State0, Callback) ->
     case osiris_log:send_file(Socket, Local0, Callback) of
@@ -433,7 +437,9 @@ chunk_iterator(
                             Err
                     end;
                 end_of_stream ->
-                    {end_of_stream, State0#?MODULE{mode = Remote1}}
+                    {end_of_stream, State0#?MODULE{mode = Remote1}};
+                {error, timeout} = Err ->
+                    Err
             end;
         {become_local, _} ->
             counters:add(counter(), ?C_REMOTE_CLOSE, 1),
@@ -444,7 +450,9 @@ chunk_iterator(
                     Err
             end;
         {end_of_stream, Remote} ->
-            {end_of_stream, State0#?MODULE{mode = Remote}}
+            {end_of_stream, State0#?MODULE{mode = Remote}};
+        {error, timeout} = Err ->
+            Err
     end;
 chunk_iterator(#?MODULE{mode = Local0} = State0, Credit, PrevIter) ->
     case osiris_log:chunk_iterator(Local0, Credit, PrevIter) of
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):
These changes all look good to me 👍 

It shouldn't interfere to much with the refactor I'm trying out in #162 so I think it's good to merge now, if your testing says so. (The second commit on that branch now (https://github.com/amazon-mq/rabbitmq-stream-s3/commit/9e8c03701ddbcb3e3ce2619cb3bd05b9bb08eefe) has tests for these cases.

PR #162, opened 2026-05-22
Split functional-core/imperative-shell for read path

the-mikedavis:
This is a refactor/redesign of the reading path, specifically the `rabbitmq_stream_s3_remote_reader` process. It is hard to test currently because the side effects (read requests, timers, etc.) are mixed in with the logic. With the `_core` module here we can test tricky situations like sequences of timeouts without the test case needing any timing concerns.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ include ../../erlang.mk
 # CT suites share a node name (erlang.mk limitation), so they cannot run in parallel.
 .NOTPARALLEL:
 
-CT_QUICK_SUITES = api_fs db replica_reader_core replica_reader log_reader fragment_iterator fragment_assembly prop manifest_replica
+CT_QUICK_SUITES = api_fs db replica_reader_core remote_reader_core replica_reader log_reader fragment_iterator fragment_assembly prop manifest_replica
 
 ERLFMT ?= erlfmt
 ERLFMT_FILES = src/*.erl test/*.erl
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -163,8 +163,8 @@ If the task crashes or S3 is unavailable, the orphan detection mechanism eventua
 When a consumer subscribes at an offset that no longer exists locally:
 
 1. `rabbitmq_stream_s3_log_reader` checks the manifest cache. If the offset is below the local range and within the manifest's range, it resolves the starting fragment.
-2. A `rabbitmq_stream_s3_remote_reader` process is spawned for the consumer. It prefetches fragment data from S3 using HTTP range requests.
-3. The remote reader uses the fragment iterator to navigate forward through the manifest tree without knowing its internal structure (groups, kilo-groups, etc.).
+2. A `rabbitmq_stream_s3_remote_reader` process is spawned for the consumer. It delegates all decisions (buffering, retry, fragment transitions) to `rabbitmq_stream_s3_remote_reader_core` and executes the resulting effects (S3 requests, timers, replies).
+3. The remote reader core uses the fragment iterator to navigate forward through the manifest tree without knowing its internal structure (groups, kilo-groups, etc.).
 4. When the consumer catches up to the local tier, the log reader transitions to local reads. The remote reader exits.
 
 If the remote reader encounters a 404 (fragment deleted by retention between iterator creation and fetch), it refreshes the iterator from the manifest cache and repositions at the oldest available offset.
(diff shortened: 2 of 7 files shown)


lukebakken:
## Deadline logic for the new core/shell architecture

PR #158 adds a `?PENDING_READ_DEADLINE_MS` (50 seconds) to the old remote reader. After merging that fix, the deadline logic needs to be ported to this new design. Here's how it fits.

**Problem recap:** The caller's `gen_server:call` has a 60-second timeout. If S3 is persistently slow, the remote reader must reply `{error, timeout}` before 60 seconds to avoid crashing the caller. The old code held `pending_read` indefinitely; the new code has the same issue via the `from`/`since` fields in the shell.

**Where it belongs: the shell, not the core.**

The core is time-free (no system clock access). The `since` timestamp is already tracked in the shell (line 198). The deadline is a wall-clock check that doesn't affect decision logic - it's a safety valve.

**Implementation sketch:**

The shell checks elapsed time before feeding events to the core. If the deadline has passed, it replies `{error, timeout}` directly, cancels in-flight requests, and resets the core's buffer state (to prevent the `Offset >= StartPos` assertion failure on retry - see #161).

```erlang
%% Shell helper:
check_deadline(#state{from = undefined}) -> ok;
check_deadline(#state{since = Since}) ->
    Elapsed = erlang:convert_time_unit(
        erlang:monotonic_time() - Since, native, millisecond
    ),
    case Elapsed >= ?PENDING_READ_DEADLINE_MS of
        true -> expired;
        false -> ok
    end.
```

Check points: `handle_info(retry_requests, ...)` and `handle_async_response` - both are entry points where the core might try to serve a pending read.

On expiry: reply `{error, timeout}`, cancel async requests, and call a `reset/1` function on the core that clears buffer state to position 0. This ensures the next read call (at any position) passes the `Offset >= StartPos` assertion.

The core needs a `reset/1` export:

```erlang
-spec reset(state()) -> state().
reset(State) ->
    State#state{
        buffer = <<>>,
        start_pos = 0,
        current_pos = 0,
        end_pos = 0,
        pending = undefined,
        requests_in_flight = #{},
        retry_delay = ?MIN_RETRY_DELAY_MS
    }.
```

**Testing:** A `remote_reader_core_SUITE` test can verify that after `reset/1`, a read at any offset passes the assertion. The shell deadline logic can be tested by verifying the reply is emitted before 60 seconds under simulated persistent timeouts.

**Related:** #157, #158, #161

the-mikedavis:
I think we want to add the timer to the shell but modify the core with a new `deadline_expired` event, and have the core decide how to reset rather than the shell mutating it from the outside with `reset/1`. Like:

```erlang
%% Core exports:
-spec step(state(), event()) -> {state(), [effect()]}.
%% Add: deadline_expired as an event

step(State, deadline_expired) ->
    State1 = State#state{
        buffer = <<>>,
        start_pos = 0,
        current_pos = 0,
        end_pos = 0,
        pending = undefined,
        requests_in_flight = #{},
        retry_delay = ?MIN_RETRY_DELAY_MS
    },
    {State1, [{reply, {error, timeout}}]}.
```

This way it's like the existing `retry` timer. The shell owns the timers but passes information about timers into the core.

I drafted this idea out in https://github.com/amazon-mq/rabbitmq-stream-s3/commit/9e8c03701ddbcb3e3ce2619cb3bd05b9bb08eefe

lukebakken:
## Review fixes (4 commits after merge from main)

- Ported `jump_to_oldest` loop guard from #167 to the core's `handle_manifest_range` - becomes local when `FirstOffset == CurrentOffset`. Test added.
- Shell exports `start/1` (no link). The monitor is sufficient; the link caused spurious EXIT warnings.
- Corrected `request_timeout_ms` default from 30s to 15s (per #158).
- Removed dead `maps:is_key` branch in `try_fragment_transition` where both arms returned `{await, State}`.
- Added `deadline_expired` to the property test event generator.
- Updated `read-path.md` to reflect the unlinked remote reader.

lukebakken (commented):
## Soak test results - `log-reader-split` branch

Second consecutive 2-hour high-throughput soak test passed on the core/shell refactor branch (tip: 1d345b9). Verified all 3 nodes running the correct commit.

**Config:** 6 streams, 12 producers, 12 consumers, 2h duration, 3-node cluster.

**Results:**
- Replay from `first`: 1.7M messages consumed across 6 streams (exceeded 1.46M expected)
- Consume rate: steady ~30k msg/s sustained, never dropped to zero
- S3 objects: plateaued at 998
- No assertion failures, no `case_clause` crashes, no `EXIT` warnings
- No errors in logs - only expected S3 timeout warnings (handled by retry/deadline logic)
- No memory alarms

**Memory growth (area to watch):**
- Node 1: 3.7x (288MB to 1059MB)
- Node 2: 8.4x (267MB to 2236MB)
- Node 3: 5.3x (286MB to 1509MB)

No memory alarms fired, but the growth on node 2 is steeper than ideal over 2h. Worth profiling separately - likely binary heap accumulation under sustained throughput rather than a leak in the refactored code (first soak test on the `main` branch showed similar patterns).

PR #167, opened 2026-05-22
remote reader: Break jump_to_oldest loop when manifest is stale

### LINKED ISSUE #166: remote reader: `jump_to_oldest` enters infinite loop when manifest references a deleted fragment
## Summary

`jump_to_oldest` enters an infinite loop when the manifest's `first_offset` references a fragment that retention has already deleted from S3. The remote reader jumps to the stale offset, gets a 404, queries the manifest range (which returns the same stale offset), and jumps again - at S3 404 round-trip speed (~10ms per iteration).

## Mechanism

1. Retention deletes a fragment from S3
2. The manifest's `first_offset` still references it (writer's persist cycle hasn't completed)
3. A consumer's remote reader tries to read the next fragment, gets 404 (`not_found`)
4. `handle_request_error(not_found, ...)` sets `current_not_found = true`
5. `try_read` matches `current_not_found = true`, calls `get_range(StreamId)` which returns `{FirstOffset, _}` from the stale manifest
6. `jump_to_oldest(State, FirstOffset)` looks up the manifest, sets `fragment_ref` to the fragment at `FirstOffset`, starts a new request, and replies `{next_fragment, FirstOffset}` to the log reader
7. The log reader repositions and calls `read(Pid, 8, 303, chunk_boundary)` on the remote reader
8. The new request for the fragment at `FirstOffset` returns 404 from S3 (it was deleted)
9. `current_not_found = true` again, `get_range` returns the same `FirstOffset` - goto step 5

The loop runs at S3 404 round-trip speed because a 404 response is immediate (no data to stream). Each iteration takes ~10-20ms, producing millions of log entries ("Jumping to oldest available fragment") and consuming CPU while delivering zero data.

## Evidence

From a 2-hour soak test on a 3-node cluster:

- 1,828,523 "Jumping to oldest available fragment" log entries on a single node
- Loop started at minute 58 of the test and continued for the remainder
- All jumps target the same offset (e.g. `00000000000023634697`)
- Manifest `first_offset` points to a fragment deleted from S3 by retention
- Actual earliest fragment in S3 is at a higher offset than the manifest claims
- Publishing-phase consumers got stuck in this loop; replay consumers also failed because the manifest was still stale

## Why the manifest is stale

Retention evaluates locally and deletes S3 objects immediately, but the manifest update only propagates on the next persist cycle. Between the S3 delete and the manifest persist, any consumer that needs to jump to the "oldest" fragment will hit a 404 on the manifest's `first_offset`.

## Both architectures are affected

The current code (`remote_reader.erl`) and the PR #162 core/shell refactor (`remote_reader_core.erl`) have the same pattern:

- Current: `try_read` -> `current_not_found` -> `get_range` -> `jump_to_oldest` -> 404 -> loop
- PR #162 core: `step({request_error, ..., not_found})` -> `{lookup_manifest_range}` -> `handle_manifest_range` -> `{jump_to_oldest, FirstOffset}` -> `{jumped, ...}` -> 404 -> loop

Neither has loop detection or a fallback for "the fragment we just jumped to is also gone."

## Suggested fix

Detect when `jump_to_oldest` would jump to the same fragment offset that just returned 404. Options:

1. **Track the last-jumped-to offset**: if the next jump targets the same offset, return `end_of_stream` or `become_local` instead of looping
2. **Skip to the next manifest entry**: if the first entry is 404, try the second entry (the next fragment that might still exist in S3)
3. **Add a backoff/retry limit to jump_to_oldest**: after N consecutive jumps to the same offset, give up

Option 2 is the most robust - it finds the next available fragment rather than giving up entirely. The consumer can resume reading from there.

## Related

- #157, #158 (timeout fixes that kept consumers alive long enough to hit this)
- #163 (`index_data` 404 handling during attach - same retention race, different code path)
- PR #162 (core/shell refactor - also affected)

lukebakken:
## Summary

- Add a guard clause to `jump_to_oldest` that detects when the manifest's `first_offset` points to the fragment we're already on (and got a 404 for)
- Transition to `become_local` instead of looping, allowing the consumer to stall briefly until the manifest is updated

## Problem

When retention deletes a fragment from S3 before the manifest persist cycle updates `first_offset`, the remote reader enters an infinite loop:

1. Jump to the manifest's first fragment
2. S3 returns 404
3. Query manifest range (returns the same stale offset)
4. Jump again - goto 1

The loop runs at S3 404 round-trip speed (~10ms/iteration), generating millions of log entries and consuming CPU while delivering zero data. Observed 1,828,523 "Jumping to oldest available fragment" events targeting the same offset in a single 2-hour test run.

## Fix

A guard clause on `jump_to_oldest` checks if `FirstOffset =:= CurrentOffset` (the fragment we're currently on and already know is gone). If so, cancel requests and transition to `become_local` instead of looping. The consumer stalls briefly until the manifest is updated on the next persist cycle, then resumes normally on the next attach.

## Test plan

- [x] Compiles cleanly
- [x] No trailing whitespace
- [ ] Deploy to test cluster and run 2-hour high-throughput soak test
- [ ] Verify: no "Jumping to oldest" loop, consumers recover after retention fires, replay succeeds

Closes #166.

### DIFF
--- a/src/rabbitmq_stream_s3_remote_reader.erl
+++ b/src/rabbitmq_stream_s3_remote_reader.erl
@@ -830,6 +830,25 @@ cancel_requests(Requests) ->
     counters:put(counter(), ?C_REQUESTS_IN_FLIGHT, 0).
 
 jump_to_oldest(
+    #?MODULE{
+        fragment_ref = #fragment_ref{offset = CurrentOffset},
+        requests = Requests,
+        cancelled_requests = Cancelled0
+    } = State0,
+    FirstOffset
+) when FirstOffset =:= CurrentOffset ->
+    %% The manifest's first_offset points to the fragment we already know
+    %% is gone (404). The manifest is stale - retention deleted the
+    %% fragment but the persist cycle hasn't updated the manifest yet.
+    %% Fall through to the local tier to avoid an infinite loop.
+    cancel_requests(Requests),
+    Cancelled = maps:merge(Cancelled0, #{Req => ok || Req := _ <- Requests}),
+    State = goto_next_fragment(State0#?MODULE{
+        requests = #{},
+        cancelled_requests = Cancelled
+    }),
+    {become_local, State};
+jump_to_oldest(
     #?MODULE{stream = StreamId, requests = Requests, cancelled_requests = Cancelled0} = State0,
     FirstOffset
 ) ->


src/rabbitmq_stream_s3_remote_reader.erl
```diff
@@ -829,6 +829,25 @@ cancel_requests(Requests) ->
...
+    %% The manifest's first_offset points to the fragment we already know
+    %% is gone (404). The manifest is stale - retention deleted the
+    %% fragment but the persist cycle hasn't updated the manifest yet.
+    %% Fall through to the local tier to avoid an infinite loop.
+    cancel_requests(Requests),
+    Cancelled = maps:merge(Cancelled0, #{Req => ok || Req := _ <- Requests}),
+    State = goto_next_fragment(State0#?MODULE{
+        requests = #{},
+        cancelled_requests = Cancelled
+    }),
```
the-mikedavis, line 849: I think we do this exact sequence in a few places of cancel_requests/1, then maps:merge adding the new requests, then do something with the state. I think we could have a cancel_all helper that takes the current state and does the same

lukebakken: Good call — filed #168 to extract a `cancel_all_requests/1` helper for this. We'll get it in a follow-up.

PR #169, opened 2026-05-27
Defer retention reaping until manifest is persisted

the-mikedavis:
Connects #166

This is a bit of a behavior change for retention and the remote reader. The main change is that the replica reader only triggers deletion of objects from retention _after_ it has persisted the newly updated manifest with the retention edits. This reduces the window where a reader could get a 404, re-read the manifest, and still see the old data. (It doesn't eliminate the window when reading from replica nodes, though.)

With this change I've also updated the reader's behavior when it gets a 404. The "Refresh fragment iterator in 404s" commit has details and a scenario to consider. By refreshing the fragment iterator on a 404, we get a bit of delayed retry on re-reading the manifest and can sometimes catch up to where fragments actually exist. Overall this lets a lagging reader which is slower than retention read the parts of the remote tier that still exist.

### DIFF
--- a/src/rabbitmq_stream_s3.erl
+++ b/src/rabbitmq_stream_s3.erl
@@ -73,6 +73,7 @@ efficiently using the `rabbitmq_stream_s3_array` module.
     uid/0,
     format_uid/1,
     offset_filename/2,
+    ref_key/2,
     manifest_key/2,
     group_key/2,
     group_name/1,
@@ -118,6 +119,12 @@ pad_zeroes(Offset) ->
         false ->
             Num
     end.
+
+-doc "Returns the S3 key for any ref type.".
+-spec ref_key(stream_id(), #fragment_ref{} | #group_ref{} | #manifest_ref{}) -> key().
+ref_key(StreamId, #fragment_ref{} = Ref) -> fragment_key(StreamId, Ref);
+ref_key(StreamId, #group_ref{} = Ref) -> group_key(StreamId, Ref);
+ref_key(StreamId, #manifest_ref{} = Ref) -> manifest_key(StreamId, Ref).
 
 -doc "Creates the key for the given stream and manifest ref".
 -spec manifest_key(stream_id(), #manifest_ref{}) -> key().
--- a/src/rabbitmq_stream_s3_remote_reader.erl
+++ b/src/rabbitmq_stream_s3_remote_reader.erl
@@ -362,57 +362,34 @@ execute_effect(
 execute_effect({set_timer, DelayMs}, State) ->
     erlang:send_after(DelayMs, self(), retry_requests),
     State;
-execute_effect(lookup_manifest_range, #state{stream = StreamId, core = Core0} = State) ->
-    %% Synchronous local lookup — feed result back to core immediately.
-    Range = rabbitmq_stream_s3_manifest_replica:get_range(StreamId),
-    {Core1, Effects} = rabbitmq_stream_s3_remote_reader_core:step(Core0, {manifest_range, Range}),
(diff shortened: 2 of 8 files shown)


lukebakken (commented):
## Review

The four commits each tell a clear story and the test coverage is appropriate. CI is green. Bottom line: I'd approve.

### Commit notes

**dae0eae replica reader: Defer key reaping until retention change is persisted**

The fix is the right shape. Queue refs on `deferred_deletions` in `on_remote_retention/4`, flush in `on_persist_completed/2`. Both happen in the gen_server process, so no concurrency issues.

The `reinitialize` path correctly clears `deferred_deletions` because the unpersisted retention edits are also discarded with the rest of the in-memory state. The new core re-evaluates retention and produces its own deletions. A one-line comment on the `deferred_deletions = []` reset would help future readers; the implication that the remote objects are intentionally leaked-then-recovered is not obvious from the surrounding code.

One subtle point worth confirming: a single `on_persist_completed` flushes ALL deferred deletions, including ones tied to retention edits that may still be in `edits_since_persist` (un-persisted). I think this is still safe because `start_persist` snapshots `in_persist_count = since_persist`, but the `last_persisted_manifest` advanced by `update_range` reflects the snapshotted manifest, which already has all those edits applied (`apply_edit` runs in `retention_complete` immediately). So readers fetching the cached manifest after `update_range` see the post-retention range, even for edits that are still queued for the next persist. If persist N covers [E1] but not [E2], and we flush both before persisting E2, then a process crash before persist N+1 lands leaves a window where the persisted manifest still lists E2's fragments but they are gone from S3. The 404 path in 98ac271 handles this gracefully. The two commits are paired correctly.

**14429b0 Add a helper to create a key from any ref type**

Pure refactor. Spec covers the three ref types, three call sites updated. Nothing to flag.

**98ac271 log reader: Refresh fragment iterator in 404s**

The meatiest change. Replacing `lookup_manifest_range` plus `jump_to_oldest` with a single `{refresh_iterator, NotFoundOffset}` effect simplifies the state machine considerably and is naturally immune to the stale-manifest infinite-loop bug from #166: the iterator is unconditionally advanced past `NotFoundOffset` in the shell, so there is no way to cycle back to the same offset.

The core's `step({iterator_refreshed, Iterator}, ...)` rebuilds a brand-new `#state{}` record. This drops `current_not_found`, `pending`, `requests_in_flight` correctly. It also resets the AIMD `read_size` to `initial_read_size`. That's a deliberate and reasonable choice (treating a 404-driven jump as a fresh start) but worth flagging.

`next_fragment_offset/1` (line 407) is a partial pattern match. Its only caller is `try_serve` when `try_read` returns `{not_found_check_range, _}`, which only happens when `next = not_found`, set by the 404 path on the next fragment. At that moment the iterator is positioned at the 404'd entry, so `next/1` always succeeds with `{ok, ...}`. The invariant is sound but tightly coupled. A short comment on the function or a tagged result would document the assumption.

Test renames and the prop_SUITE generator switch all match the new flow.

**a4a32ab replica reader: Evaluate remote tier retention when updated**

Symmetric with what `osiris_writer` does for the local tier. The cleaned-up `read_repositions_on_fragment_not_found/1` is now a much stronger test, exercising a live reader hitting 404s after retention deletes fragments under it, which is the exact scenario this PR targets.

### Process notes

I have not run a soak test on this branch yet. Worth doing since this changes the read path under retention pressure.

### Optional nits

- Comment on `deferred_deletions = []` in `reinitialize` documenting the leak-and-recover behavior.
- Comment or tagged result on `next_fragment_offset/1` documenting the caller-side invariant.

lukebakken (approved):

PR #171, opened 2026-05-28
Handle group download function errors

the-mikedavis:
This adds handling for 404s, 500s and 503s in the `GetGroupFun` functions. We do a small amount of sleeping for the transient errors (500, 503) and advance past a group which is 404, since it must've been reclaimed by retention. Conceptually we apply the same strategies for group objects as we do for fragments.

This is based on top of #169.


lukebakken (approved):

the-mikedavis:
(Force-push was a rebase + rewording the commit message)

Issue #178, opened 2026-05-29
Make rabbitmq_stream_s3_api_aws_pool testable without a real S3 endpoint


lukebakken:
## Summary

`rabbitmq_stream_s3_api_aws_pool` is hard to unit-test because `open/0` directly calls `gun:open(Host, 443, Opts)` with hostname/port baked in. State-machine bugs in the pool (such as #177) cannot easily be regression-tested without a real S3 endpoint or a refactor that lets tests inject connections.

## Why this matters

The fix for #177 is in the pool's `'DOWN'` handler. The natural regression test is "spawn a process, have it check out a connection, kill the process, assert the connection is not on the available list." That requires the pool to have a connection in its `monitors` and `checkouts` maps, which currently requires `gun:open` to succeed against a real host.

The existing `rabbitmq_stream_s3_api_aws_SUITE` only contains `kick_the_tires` and skips entirely without real AWS credentials. The pool has no dedicated suite.

## Suggested approach

Extract the connection-opening into a callback or option so tests can inject a fake. Sketch:

```erlang
init(#{min_size := MinSize, max_size := MaxSize, name := Name} = Config) ->
    OpenFun = maps:get(open_fun, Config, fun open/0),
    ...
    {ok, #?MODULE{
        min_size = MinSize,
        max_size = MaxSize,
        counter = Cnt,
        open_fun = OpenFun
    }}.
```

Tests pass `open_fun => fun fake_open/0` where `fake_open/0` spawns a plain process that just sleeps. The pool monitors it the same way it monitors a gun process, and the test can `exit(Pid, kill)` to simulate any failure mode. `gun:close/1` calls in the pool would also need to go through a callback (or the fake process needs to handle the close cast), but that is a small extension.

A new `pool_SUITE` could then cover at least:

- A clean checkout/checkin cycle leaves the pool in a consistent state
- A caller dying while holding a checked-out connection causes the connection to be closed and a replacement grown (the #177 regression)
- A connection process dying (gun crash) causes the connection to be removed and replaced
- `try_checkout` returns `busy` when the pool is at max-size and all are checked out
- `try_checkout` queues and waits when the pool has room to grow (relevant to #174)
- Idle timer reclaims an idle connection without disturbing checked-out ones

## Origin

Filed as a follow-up while implementing the fix for #177. The fix itself is being landed without a unit test because adding one requires this refactor; the soak test in `retention-stress-test.sh` (private repo) is the regression test until this is resolved.

[...]

PR #179, opened 2026-05-29
Drop, do not reuse, a connection whose caller died holding it

### LINKED ISSUE #177: Writer node OOMs ~15s after osiris:update_retention/2 drop, due to gun connection-reuse race
## Summary

The stream's writer node OOMs and is killed by the kernel within ~15 seconds of a sharp `osiris:update_retention/2` drop, on the current `main` branch. Reproduced twice with identical fingerprints. The replica nodes are unaffected; only the writer dies.

The root cause is a `gun` connection-reuse race in our S3 upload path: a `headers` cast is sent to a `gun` `gen_statem` that is still mid-body on a chunked PUT, producing `error:function_clause` in `gun_http:headers/12`. The crashing process holds a 67 MB pending request state plus the TLS option bag including a full CA bundle. The Erlang error logger then formats that state for the SASL crash report (25,888 lines / 3.3 MB of formatted output for one crash). That formatting allocates several GB transiently, crosses the high-watermark, and the kernel OOM-killer terminates `beam.smp` shortly after.

The fix should be on the gun-usage side. Once gun does not crash, the logger does not have a giant state to format and the OOM does not occur.

## Reproduction

The failure reproduces with `retention-stress-test.sh` (private repo).

Workload, per the test harness:

- 1 stream, 3 nodes (`m7g.large`, 8 GB RAM, 4660 MiB high-watermark)
- 1 producer at unlimited rate, 1 consumer at `--consumer-rate 5000` from offset `first` (so the consumer pins itself far behind the head)
- Initial retention: `max-length-bytes 10gb`
- Run for 15 minutes; producer steadily fills the remote tier
- Then call `osiris:update_retention(WriterPid, [{max_bytes, 209715200}])` to drop retention to 200 MB
- Continue running for another 15 minutes

State just before the drop on the second run:

- ~316 fragment objects in S3
- Manifest range `[173511113, 183378202)` (~9.87M offsets)
- Producer was sustaining ~170 MB/s per the manifest growth rate
- Consumer was holding at 5000 msg/s (lagging far behind, by design)

## Timeline

Times from the second run (the first run had identical structure but slightly different timestamps):

- `14:24:39.794` writer applies `update_retention` from 10 GB to 200 MB
- `14:24:39.797` first retention re-evaluation completes in 2.4 ms
- `14:24:43` test harness samples manifest: range collapsed to `[188037219, 188296763)` (~260K offsets), S3 fragment count dropped from 316 to 168. So ~9.6M offsets / ~148 fragments were marked for deletion in the manifest by retention
- `14:24:46.881` another retention pass (1.5 ms)
- `14:24:47.101` writer opens new local segment (publishing continues)
- `14:24:48.886` gun gen_statem crashes with `function_clause` (see Stacktrace section)
- `14:24:54.260` `vm_memory_high_watermark set. Memory used:5976432640 allowed:4887080140` (writer holding 5.57 GiB, threshold 4.55 GiB, total RAM 7.59 GiB)
- shortly after, kernel OOM-killer terminates beam.smp. The log file has no further entries.

The replica nodes log only the writer's memory alarm and then notice the writer is dead two seconds later. Their own memory stays normal throughout.

The first run hit the same fingerprint:

- Drop at 21:44:28.284, gun crash at 21:44:37.410 (9 s later), watermark crossed at 21:44:42.101 (14 s after drop)

The two runs are visibly the same failure.

## gun stacktrace

```
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0> ** State machine <0.11605.0> terminating
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0> ** Last event = {cast,{headers,<0.12021.0>,
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0>                                #Ref<0.3517766163.3633315842.48310>,<<"PUT">>,
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0>                                <<"/rabbitmq/stream/__retention-stress-test-20260529-140926_.../data/00000000000188620810.e627a4fa.fragment">>,
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0> ** Reason for termination = error:function_clause
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0> ** Stacktrace =
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0> **  [{gun_http,headers,
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0>          [{file,"src/gun_http.erl"},{line,568}]},
2026-05-29 14:24:48.886331+00:00 [error] <0.11605.0>      {gun,connected,3,[{file,"src/gun.erl"},{line,1441}]},
```

`gun_http:headers/12` has only two clauses:

```erlang
headers(State, StreamRef, ReplyTo, _, _, _, _, _, _, CookieStore, _, EvHandlerState)
        when is_list(StreamRef) -> ...                                  %% line 568
headers(State=#http_state{opts=Opts, out=head}, ...) -> ...             %% line 573
```

The dumped state has `out = {body, 67233417}`, not `head` (the 11th field of `#http_state{}`). So gun received a new `{cast, {headers, ...}}` while in the middle of streaming a chunked PUT body. The function_clause is correct for that input. The bug is upstream in our code: something is sending a `headers` cast on a connection that is mid-body in a chunked PUT.

## Notes on which node dies

- `rabbit_stream_coordinator: started writer __retention-stress-test-... on rabbit@ip-10-0-1-162 in 1` -- node0 was the original writer.
- node0 OOMs and dies. Replicas (node1, node2) remain healthy.
- After failover, node2's `rabbitmq_stream_s3_reaper:objects_deleted=3343` and `replica_reader:fragments_deleted=2655`, suggesting it took over writer duties and processed deletions post-failover.

So the failure is specifically on the **writer node** at the moment retention is dropped.

## Why the gun crash is the trigger

The single gun crash event in the log spans:

```
$ awk '/2026-05-29 14:24:48.886331/' node0.log | wc
   25888  ...  3306229
```

25,888 lines and 3.3 MB of formatted output for one `gen_statem terminating` report. The state being formatted contains:

- The chunked PUT request state, with `out = {body, 67233417}` (67 MB body still pending)
- Full request headers including a multi-KB AWS STS session token
- The TLS option bag including `cacerts` (a CA bundle of multiple 1-2 KB DER blobs)
- The streams list, the http_state record, the gen_statem state tuple

The error logger walks all of that and produces a printable representation. For the binaries it dumps each byte as a comma-separated integer literal inside `<<...>>`. Counting only lines containing 4-or-more comma-separated decimal integers in the dump gives 17,582 lines on this one report. The temporary iolist plus integer terms used during `io_lib:write/1` traversal easily exceeds an order of magnitude beyond the source data.

The 14.5 second window between drop and OOM lines up: gun crashes ~9 seconds after the drop, the error logger then formats the report over the next ~5 seconds, and the watermark crosses.

The chain of evidence:

- The only error-level event in the death window is the single gun crash
- That event's log output is ~25 k lines / ~3 MB
- OOM follows the crash by 5.4 s (`14:24:48.886` -> `14:24:54.260`)
- No other process on the writer reports memory pressure or crashes
- Replica nodes (which do not run the upload pipeline and do not see this gun crash) stay healthy

To strengthen the diagnosis, a future run could capture `recon:proc_count(memory, 20)` at `--retention-drop-after - 30s` and again 5 s after the drop. I have not done that yet.

## Branch

Reproduced on `main` (725013b "Merge pull request #171 from amazon-mq/group-fun-error", which includes PR #169 deferred-deletion + PR #171 group-fun-error handling), no local modifications.

## Saved evidence

Logs from both runs, including the full crash trace, are saved locally at `/tmp/rstest-pre-reset-20260529/` (run 1) and `/tmp/rstest-run3-20260529/` (run 2). I can attach or quote any of it on request.

lukebakken:
## Summary

Stop reusing an S3 upload pool connection whose caller died holding it. The old behavior put the connection back on the available list, which crashed gun's gen_statem on the next checkout (when the connection was still mid-PUT on a chunked body) and OOM'd the writer node.

Closes #177.

## Changes

- In the pool's `'DOWN'` handler (caller-down branch), close the connection and call `grow/1` to replace it instead of `make_available/2`
- Extract `close_connection/2` helper for the demonitor + `gun:close` + `monitors` cleanup pattern; the `idle_timeout` handler now uses it too

## Test plan

- [x] `gmake` compiles cleanly
- [x] `gmake dialyze` passes
- [x] All existing CT suites pass (`replica_reader_SUITE` 21/21, `replica_reader_core_SUITE` 60/60, `remote_reader_core_SUITE` 25/25, `prop_SUITE` 29/29, others)
- [x] Re-run the retention-stress soak test to confirm the writer no longer OOMs on a sharp `osiris:update_retention/2` drop. Reproduced cleanly twice on the parent commit; expect the fix to make it pass

A unit-test regression check is not included because exercising the pool's `'DOWN'` handler without a real S3 endpoint requires injecting a fake connection. Tracked as a follow-up in #178.

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws_pool.erl
+++ b/src/rabbitmq_stream_s3_api_aws_pool.erl
@@ -295,13 +295,20 @@ handle_info(
         _ ->
             case CheckoutsRev0 of
                 #{MRef := Conn} ->
-                    %% A caller process is down which has a conn checked out.
-                    %% Return the conn.
+                    %% A caller process is down while holding a checked-out
+                    %% connection. The connection may be in any state on the
+                    %% wire (e.g. mid-body in a chunked PUT if the caller
+                    %% died between `gun:headers/4` and `gun:data/4 fin`).
+                    %% Reusing it would land a `headers` cast on a connection
+                    %% whose `out` field is not `head`, crashing gun's
+                    %% gen_statem with `function_clause`. Drop the connection
+                    %% and grow a replacement.
+                    %% See: https://github.com/amazon-mq/rabbitmq-stream-s3/issues/177
                     State1 = State0#?MODULE{
                         checkouts = maps:remove(Conn, Checkouts0),
                         checkouts_rev = maps:remove(MRef, CheckoutsRev0)
                     },
-                    {noreply, make_available(Conn, State1)};
+                    {noreply, grow(close_connection(Conn, State1))};
                 _ ->
                     %% A pending caller process is down. Remove it from the
                     %% pending queue.
@@ -312,21 +319,14 @@ handle_info(
                     {noreply, State}
             end
     end;
-handle_info({idle_timeout, Conn}, #?MODULE{idle_timers = Timers, monitors = Monitors} = State0) ->
+handle_info({idle_timeout, Conn}, #?MODULE{idle_timers = Timers} = State0) ->
     case Timers of
         #{Conn := _} ->
             %% Timer is still active (wasn't cancelled by a checkout), so the
             %% connection has been idle too long. Close it.
             State1 = State0#?MODULE{idle_timers = maps:remove(Conn, Timers)},
             State2 = cancel(Conn, State1),
-            case Monitors of
-                #{Conn := ConnMRef} ->
-                    erlang:demonitor(ConnMRef, [flush]),
-                    gun:close(Conn),
-                    {noreply, State2#?MODULE{monitors = maps:remove(Conn, Monitors)}};
-                _ ->
-                    {noreply, State2}
-            end;
+            {noreply, close_connection(Conn, State2)};
         _ ->
             %% Timer was already cancelled (stale message). Ignore.
             {noreply, State0}
@@ -489,6 +489,19 @@ make_available(Conn, #?MODULE{available = Available, idle_timers = Timers} = Sta
         available = [Conn | Available],
         idle_timers = Timers#{Conn => TRef}
     }).
+
+%% Close a connection and stop monitoring it. Used when the connection has
+%% been idle too long, or when its caller died holding it (and we cannot
+%% trust its on-wire state). A no-op if the connection is not in `monitors`.
+close_connection(Conn, #?MODULE{monitors = Monitors} = State) ->
+    case Monitors of
+        #{Conn := ConnMRef} ->
+            erlang:demonitor(ConnMRef, [flush]),
+            gun:close(Conn),
+            State#?MODULE{monitors = maps:remove(Conn, Monitors)};
+        _ ->
+            State
+    end.
 
 cancel_idle_timer(Conn, #?MODULE{idle_timers = Timers0} = State) ->
     case Timers0 of


the-mikedavis (approved):
This strategy makes sense to me 👍

Issue #172, opened 2026-05-28
rabbitmq_stream_s3_reaper crashes on pool checkout timeout under S3 throttling


lukebakken:
## Summary

`rabbitmq_stream_s3_reaper` crashes with a 5-second `gen_server:call` timeout when the S3 connection pool is saturated by slow or timing-out requests. The supervisor restarts it, so it is recoverable, but the in-flight delete batch is lost and the crash is alarming in logs.

## Reproduction

Observed during a 2-hour high-throughput soak test on the `streams-tiered-storage-combined` branch (6 streams, 12 producers, 12 consumers).

When S3 starts throttling, many `S3 request timed out` and `Slow remote tier read` warnings appear (reads taking 11-16 seconds). The reaper, which uses the same `rabbitmq_stream_s3_general_pool` as everything else, then crashes at the pool checkout step because the default timeout is 5 seconds and all pool workers are tied up on the slow requests.

Two crashes in this run, on two of the three nodes, within roughly 30 seconds of each other:

- `rabbit@ip-10-0-1-21` at 15:16:31
- `rabbit@ip-10-0-1-12` at 15:17:02

Both have identical signatures.

## Crash trace

```
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>   crasher:
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>     initial call: rabbitmq_stream_s3_reaper:init/1
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>     pid: <0.545.0>
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>     registered_name: rabbitmq_stream_s3_reaper
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>     exception exit: {timeout,
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>                         {gen_server,call,
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>                             [rabbitmq_stream_s3_general_pool,
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>                              {checkout,#Ref<0.1817810223.3150708738.164407>},
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>                              5000]}}
2026-05-28 15:16:31.537278+00:00 [error] <0.545.0>       in function  gen_server:call/3 (gen_server.erl, line 1222)
[...]

PR #180, opened 2026-05-29
remote reader: Wire up the previously dead read-path metrics

### LINKED ISSUE #175: Remote reader metrics defined but never incremented
## Summary

Several metrics in `rabbitmq_stream_s3_remote_reader` are exported by the Prometheus endpoint but never incremented. Operators reading them will see permanent zeros and assume nothing is happening on the read path, when in fact the underlying activity is occurring.

In addition, the `await_duration_ms` description is misleading: it claims to measure "total milliseconds spent awaiting async data" but actually counts the total round-trip duration of every `read/4,5` reply, including reads that the buffer could have served instantly.

## Affected metrics

These are defined in `?COUNTERS` and registered via `seshat`/the histogram module, but no source file calls `counters:add` or `histogram:observe` for them:

| Metric | Type | Where defined | Where incremented |
|---|---|---|---|
| `buffer_hit` | counter | `rabbitmq_stream_s3_remote_reader.erl` line 46 | nowhere |
| `buffer_miss` | counter | `rabbitmq_stream_s3_remote_reader.erl` line 47 | nowhere |
| `fragment_transition` | counter | `rabbitmq_stream_s3_remote_reader.erl` line 48 | nowhere |
| `read_size_bytes` | histogram | `rabbitmq_stream_s3_remote_reader.erl` line 121 | nowhere |

`grep -rn 'C_BUFFER_HIT\|C_BUFFER_MISS\|C_FRAGMENT_TRANSITION' src/` only shows the definitions, no callers. Same for `read_size_bytes`: only `histogram:new` (registration) and `histogram:prometheus_format` (reporting), no `histogram:observe`.

## How this happened

`8ee55f4` "reader: Add instrumentation" added the counter increments alongside the metric definitions. `ee2d93a` "Split functional-core/imperative-shell for read path" relocated the decision logic into `rabbitmq_stream_s3_remote_reader_core` and the shell now drives effects rather than running the buffer logic directly. The increments were not carried over.

The core knows when each event happens (it returns `{ok, Data, State1}` for a buffer hit in `try_serve/1`, returns `{await, State1}` for a miss, and emits `{next_fragment, NextOffset, State}` for a fragment transition). The shell sees the resulting effects but the increments never made it into the new code.

`read_size_bytes` was meant to record the AIMD prefetch window. The core adjusts `read_size` in `adjust_read_size/2` but never tells the shell to record the value.

## Misleading description

`await_duration_ms` and `await` are incremented in `execute_effect({reply, Result}, ...)` (line 340-341), once per reply. This means every read call increments `await`, regardless of whether the data came from buffer or from a fresh S3 fetch.

In the soak test of #169 + #171, all three nodes show `read = await` exactly (e.g. node 1: 10039004 = 10039004). The metric is not a buffer-miss-await counter; it is a read-duration counter under a misleading name. Either rename it (and its description) or fix it to fire only when an S3 fetch actually blocked the reply.

## Impact

We have no way to distinguish, from metrics alone:

- buffer hits vs. misses
- the prefetch-size distribution (which would let us tune AIMD)
- fragment transitions (e.g. how often consumers cross fragment boundaries vs. read within one)

This blindspot was visible in the 2-hour soak test on the combined branch of PR #169 and PR #171: with millions of reads served, all four metrics read 0.

## Suggested fix

For the four dead metrics: have the core return either an enriched effect (e.g. `{reply, Result, hit}` / `{reply, Result, miss}`) or a separate event the shell can use to bump counters without re-deriving state. The histogram needs the shell to call `histogram:observe(?MODULE, NewReadSize)` whenever the core changes its prefetch window.

For `await_duration_ms`: either rename to `read_duration_ms` (which already exists as a separate counter on `read/4,5`) and remove the duplicate, or fire it only from the `{await, State1}` branch in `try_serve/1`.

## Origin

Found while reviewing metrics from a 2-hour high-throughput soak test on the combined branch of PR #169 and PR #171.

lukebakken:
## Summary

Wire up the four read-path metrics that have been dead since the core/shell split. The core now emits `{observe, hit | miss | fragment_transition, ReadSize}` from the three branches of `try_serve/1` that matter; the shell increments the matching counter and observes the AIMD read_size into the histogram.

Closes #175.

## Changes

- New `{observe, Kind, ReadSize}` effect emitted from `try_serve/1` after the `{reply, ...}` effect, so existing `[{reply, ...} | _]` test assertions still match
- Shell handles the three observe variants: increments `buffer_hit`, `buffer_miss`, or `fragment_transition`, and calls `rabbitmq_stream_s3_histogram:observe/2` for `read_size_bytes`
- Remove `await` and `await_duration_ms` counters along with the now-unused `since` field. They were duplicates of `read` and `read_duration_ms` (set in `read/5`), and the issue called the description misleading
- Add `observe_effects_emitted_for_hit_miss_and_transition` to `remote_reader_core_SUITE` covering all three kinds end-to-end

## Test plan

- [x] `gmake` compiles cleanly
- [x] `gmake dialyze` passes
- [x] `remote_reader_core_SUITE` passes (26/26 with the new test)
- [x] `prop_SUITE` 29/29, `replica_reader_SUITE` 21/21, `replica_reader_core_SUITE` 60/60, `log_reader_SUITE` 9/9 still pass
- [x] Confirm post-merge that the metrics actually populate on a soak run (existing `retention-stress-test.sh` will exercise hits, misses, and at least one fragment transition)


lukebakken:
Doing a test run now...

lukebakken:
## Soak test results

Re-ran the 30-minute retention-stress test on dfc0f47 (this branch's tip) end to end. Single stream, producer firehose, consumer at `--consumer-rate 5000` from offset `first`, retention drop 10gb -> 200mb at the 15-minute mark.

The four previously-dead metrics now populate correctly. From the consumer's connection node (the only one running the remote_reader for this test):

| Metric | Final value |
|---|---|
| `buffer_hit` | 245223 |
| `buffer_miss` | 655 |
| `fragment_transition` | 67 |
| `read_size_bytes_count` | 245945 |
| `read_size_bytes_sum` | 15180782764032 |

Sanity checks:

- `read_size_bytes_count` (245945) equals `buffer_hit + buffer_miss + fragment_transition` (245945) as expected, since the histogram is observed once per observe effect
- Hit rate: 245223 / (245223 + 655) = 99.7%, indicating the AIMD prefetch is highly effective for this workload
- Mean prefetch window from the histogram: 15.18 TB / 245945 = ~61.7 MB, close to the 64 MB AIMD ceiling, consistent with sustained hit-driven growth
- Histogram bucket distribution skewed toward the 8 MB+ bucket as the consumer pulled ahead; the smaller buckets stayed populated from the early-warmup phase

Other observations from the same run:

- Test's strict `verify_consumer_kept_up` check passed: 8423280 messages consumed (>= 95% of the 324042 surviving manifest range)
- No node OOMs, no gun function_clause crashes, no remote_reader crashes (PR #179's fix held)
- One reaper crash at 18:17:01 with the same signature as #172 (pool checkout timeout under S3 throttling); reaper supervisor restarted it. Unrelated to this PR

the-mikedavis (approved):

PR #183, opened 2026-06-01
Fix consumer crash/wedge from premature Fd close, poisoned pool, and pool-growth regression

### LINKED ISSUE #182: rabbit_stream_reader crashes on unhandled {error,einval} from osiris_log:read_header_with_ra/1
## Summary

`rabbit_stream_reader` crashes with `{case_clause, {error, einval}}` during local-tier reads, or wedges permanently with the consumer unable to make forward progress. Three bugs contribute:

### Bug 1: `maybe_become_remote` closes local Fd before remote path succeeds

`rabbitmq_stream_s3_log_reader:maybe_become_remote/2` (line 710) calls `osiris_log:close(Local)`, which closes the local-tier segment Fd. If the subsequent remote-tier read times out, `{error, timeout}` propagates up through `send_file` back to `rabbit_stream_reader:send_chunks/8`. The `send_chunks` error path (rabbit_stream_reader.erl line 2112-2118) preserves the **original Consumer unchanged** - including the now-closed local Log state. On the next retry, `osiris_log:send_file` calls `read_header_with_ra`, which calls `file:pread` on the closed Fd, returning `{error, einval}`.

On the fork's `tiered-storage-abstractions` branch, `read_header_with_ra` has no clause for `{error, einval}`, so this propagates as a `case_clause` crash, killing the consumer connection via `reached_max_restart_intensity`.

### Bug 2: `cancel_async` checks in a connection still draining a large response

When the remote reader's 50-second `pending_read_deadline_ms` fires, `cancel_all_requests` calls `rabbitmq_stream_s3_api:cancel_async/2`. This does:
- `gun:cancel(Conn, StreamRef)` - marks the stream as `is_alive=false` in gun; gun continues reading (and discarding) the response body from the socket
- `finish_async(State)` - cancels the 15s request timer AND **checks the connection back into the pool** via `rabbitmq_stream_s3_api_aws_pool:checkin/2`

The pool is LIFO (`take_available` pops `[Conn | Available]` at api_aws_pool.erl:416-417). The next `try_checkout` immediately pops this same connection. A new `gun:request` is issued, but on HTTP/1.1 responses arrive in order. The new response cannot be parsed until the old (cancelled, partially-delivered) response body fully drains from the TCP stream. Gun IS actively reading and discarding (`send_data(_, State, _) -> [{state, State}, {active, true}]` at gun_http.erl:486-487), but the drain is slow.

Observed throughput: 6 MB received in 50s = ~126 KB/s. At that rate, draining the remaining ~62 MB of a 68 MB fragment takes ~490 seconds - far exceeding the 50s deadline window. Each deadline cycle cancels, checks in, checks out the same poisoned connection, pipelines another request behind the drain, and waits another 50s. The consumer never makes progress.

Contrast: `handle_async({request_timeout, StreamRef}, ...)` (api_aws.erl:599) does the right thing - `gun:close(Conn)` kills the connection, the pool's `'DOWN'` handler removes it and grows a fresh replacement. But `cancel_async` (used by the deadline path) optimistically assumes the connection is reusable.

### Bug 3: `pool_busy` silently drops request but core thinks it is in flight

When `rabbitmq_stream_s3_api_aws_pool:try_checkout/1` returns `busy` (all pool connections are checked out), `execute_effect({start_request, ...})` in the remote reader shell returns `State` unchanged - no request is added to the shell's `requests` map. However, the **core** has already added the fragment offset to its `requests_in_flight` map (remote_reader_core.erl line 585) before emitting the `{start_request, ...}` effect.

The core's guard at line 581 (`not maps:is_key(Fragment, Reqs)`) prevents issuing another `start_request` for the same fragment. Since no real request is in flight, no data will ever arrive, no timeout will fire, and no retry will be attempted. The consumer is permanently wedged.

This is the **primary trigger** observed in testing: the consumer enters remote mode, the general pool is empty (min_size=0, and `try_checkout` never successfully grew it due to the demand-calculation bug), and every S3 request attempt fails with `pool_busy`. The core never retries because it believes a request is already in flight.

### How they interact

The three bugs create a cascade:
1. Consumer in local mode hits `offset_not_found` (segment rotated by retention).
2. `maybe_become_remote` transitions to remote, closing the local Fd (Bug 1 setup).
3. Remote reader tries `get_range_async` -> `pool_busy` (Bug 3). Core thinks request is in flight. Nothing happens for 50s.
4. Deadline fires. Core replies `{error, timeout}`. `cancel_all_requests` runs (finds nothing to cancel since no real request exists in the shell's `requests` map).
5. Caller preserves old Consumer with the closed-Fd local state.
6. Next retry: local pread -> einval -> crash (Bug 1).
7. After restart, same cycle. If a connection IS available: partial data, deadline, cancel checks in poisoned connection (Bug 2), next checkout gets the same stalled conn.

## Reproduction

Reproduced on every `retention-stress-test.sh` run (a single stream, 1 producer firehose, 1 consumer at 5000 msg/s from offset `first`). Consumer wedges within ~2 minutes in all runs.

The trigger is: consumer at offset `first` transitions to remote mode, the general pool has no connections (min_size=0, grow never fires due to the demand-calculation regression), every `try_checkout` returns `busy`, and local retention rotates segments under the consumer's stale local state.

### Bug 3 (pool_busy) - primary wedge trigger

From run with full diagnostic logging:
```
16:59:46 [warning] <0.906.0> remote_reader start_request: pool_busy
         key=rabbitmq/stream/.../data/00000000000006377137.9a1c2de6.fragment
17:00:36 [warning] <0.906.0> remote_reader pending-read deadline expired:
         core=#{in_flight_requests => 1, in_flight_fragment_offsets => [6377137],
                buffer_bytes => 0, start_pos => 8, end_pos => 8, ...}
17:00:36 [warning] <0.966.0> remote_reader start_request: pool_busy
         key=rabbitmq/stream/.../data/00000000000015297081.e65604af.fragment
```

The core reports `in_flight_requests => 1` but no actual request was issued (pool_busy). Different fragment keys each cycle confirm the consumer advances through the manifest but never successfully fetches data.

After the retry fix but before the pool-growth fix, the pool state on all three nodes showed: `available=[], monitors=#{}, checkouts=#{}` (zero connections total, pool never grew). `try_checkout` triggered `grow(State0)` which calculated `N=0` because `queue:len(Pending) = 0` for non-blocking callers.

### Bug 1 (einval) - from earlier runs without the deferred-close fix

```
osiris_log: pread returned einval at pos=499740960 next_offset=482266
            last_chunk_id=25441921 result=offset_not_found
```

Same `pos=499740960` every cycle. `file:pread/3` returns `einval` on Linux when the Fd has been closed by the same process. The Fd was explicitly closed by `osiris_log:close(Local)` in `maybe_become_remote`.

### Bug 2 (HTTP/1.1 head-of-line blocking) - from runs where pool had availability

First deadline showed partial data delivery (6 MB buffer), then nothing for 50s. Same fragment key across all subsequent deadlines for the stuck pid. S3 responds to the same range request in <1s via aws cli from the same node.

## Crash trace (from runs without Bug 1 fix)

```
exception error: no case clause matching {'EXIT',
    {{case_clause,{error,einval}},
     [{osiris_log,read_header_with_ra,1,[{file,"src/osiris_log.erl"},{line,3128}]},
      {osiris_log,send_file,3,[{file,"src/osiris_log.erl"},{line,1822}]},
      {rabbitmq_stream_s3_log_reader,send_file,3,[{file,"rabbitmq_stream_s3_log_reader.erl"},{line,370}]},
      {osiris_log_reader,send_file,3,[{file,"src/osiris_log_reader.erl"},{line,86}]},
      {rabbit_stream_reader,send_chunks,8,[{file,"rabbit_stream_reader.erl"},{line,3812}]}]}}
```

### Bug 1 fix (implemented in PR #183)

`maybe_become_remote` returns the local state to the caller instead of closing it. `send_file/4` and `chunk_iterator/4` thread it as a `DeferredClose` argument through recursion, closing only on terminal non-error returns (success, end_of_stream, become_local). On `{error, _}` the Fd remains valid for retry.

### Bug 2 fix (implemented in PR #183)

`cancel_async` closes the connection (same as the `request_timeout` path: `gun:close(Conn)` + `finish_async_close`) rather than optimistically checking it back into the pool.

### Bug 3 fix (implemented in PR #183)

Two parts:

1. When `get_range_async` returns `{error, pool_busy}`, the shell feeds `{request_error, _, FragOffset, pool_busy}` back into the core. `pool_busy` is added to the transient-error clause so the core clears `requests_in_flight` and schedules a retry timer with exponential backoff.

2. `try_checkout` in the pool called `grow(State0)` on the `empty when size < MaxSize` branch, but `grow/1` calculates demand from `queue:len(Pending)` which is always 0 for non-blocking `try_checkout` callers. Result: `N=0`, no connection ever opened. Fixed by calling `grow(1, State0)` directly. This was a regression from 3da765a which made `try_checkout` non-blocking without accounting for the demand-based growth calculation.

### osiris-side change

No change needed in osiris. The `{error, einval}` from `pread` is a consequence of the Fd being closed by `rabbitmq_stream_s3` code. An earlier draft PR (rabbitmq/osiris#222, now closed) proposed handling einval in `read_header_with_ra` but that was treating the symptom. The fix belongs entirely in `rabbitmq_stream_s3`.

lukebakken:
## Summary

Fix three bugs that together cause a consumer crash or permanent wedge when the remote reader's pending-read deadline fires during a large fragment download. Also fix a pool-growth regression that prevented connections from ever being created for non-blocking callers.

Closes #182.

## Changes

**`cancel_async`: Close connection instead of checking it back in (Bug 2)**

On HTTP/1.1, `gun:cancel` only marks a stream dead internally; the response body continues draining on the wire and blocks pipelined requests. `cancel_async` was checking the connection back into the pool via `finish_async`, allowing the next caller to immediately pop it (LIFO) and pipeline a request behind the still-draining response. For large fragments (68 MB at ~126 KB/s observed throughput), the drain takes minutes, far exceeding the 50-second deadline window, creating a permanent wedge.

Fix: close the connection (matching the `request_timeout` handler), letting the pool's `'DOWN'` handler grow a fresh replacement.

**log reader: Defer local Fd close until remote path succeeds (Bug 1)**

`maybe_become_remote` closed the local osiris log state (and its segment Fd) before the recursive remote-mode `send_file` returned. If that call failed with `{error, timeout}`, the error propagated to `rabbit_stream_reader:send_chunks` which preserved the original Consumer unchanged, including the now-closed local Log. The next retry called `file:pread` on the closed Fd, producing `{error, einval}` and crashing the reader.

Fix: `maybe_become_remote` returns the local state to the caller instead of closing it. `send_file/4` and `chunk_iterator/4` thread it as a `DeferredClose` argument through recursion, closing only on terminal non-error returns (success, end_of_stream, become_local). On `{error, _}` the Fd remains valid for retry.

**remote reader: Retry on `pool_busy` instead of silently wedging (Bug 3)**

When `get_range_async` returned `{error, pool_busy}`, the shell returned State unchanged. But the core had already added the fragment to its `requests_in_flight` map before emitting the `start_request` effect. The guard in `maybe_start_current_request` prevented any further attempt for that fragment. No data arrives, no timeout fires, no retry occurs. The consumer is permanently wedged.

Fix: feed `{request_error, _, FragOffset, pool_busy}` back into the core from the shell. `pool_busy` is added to the transient-error clause so the core clears `requests_in_flight` and schedules a retry timer with exponential backoff.

**pool: Fix `try_checkout` never growing the pool**

`try_checkout` returns `busy` immediately (non-blocking) and called `grow(State0)` to open a connection in the background. But `grow/1` calculates demand from `queue:len(Pending)`, which is always 0 for `try_checkout` callers (they never queue). The result: `N = 0`, no connection is opened, the pool stays empty forever. This is a regression from 3da765a which made `try_checkout` non-blocking.

Fix: call `grow(1, State0)` directly to unconditionally open one connection when the pool has room.

## Test plan

- [x] `remote_reader_core_SUITE` passes 26/26
- [x] `prop_SUITE` passes 29/29
- [x] `replica_reader_core_SUITE` passes 60/60
- [x] `log_reader_SUITE` passes 9/9
- [x] Dialyzer passes
- [x] `retention-stress-test.sh` 30-minute soak: consumer at 5000+ msg/s the entire run, survived retention drop from 10 GB to 200 MB, 9M messages consumed, zero errors
- [x] `high-throughput-test.sh` 45-minute soak: 6 streams, 12 producers, 12 consumers, sustained 250-450K msg/s with S3 recv at 30-50 MiB/s, full replay verified

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -602,8 +602,11 @@ handle_async(
 
 -spec cancel_async(async_req(), async_state()) -> ok.
 cancel_async(StreamRef, #{conn := Conn, stream_ref := StreamRef} = State) ->
-    gun:cancel(Conn, StreamRef),
-    ok = finish_async(State).
+    %% On HTTP/1.1, gun:cancel only marks the stream dead; the response body
+    %% continues draining on the wire, blocking subsequent pipelined requests.
+    %% Close the connection so the pool replaces it with a fresh one.
+    gun:close(Conn),
+    ok = finish_async_close(State).
 
 %% Cancel the request timer without checking in the connection or
 %% decrementing the active-request counter. Used when the response
--- a/src/rabbitmq_stream_s3_api_aws_pool.erl
+++ b/src/rabbitmq_stream_s3_api_aws_pool.erl
@@ -190,11 +190,11 @@ handle_call(
             counters:add(Cnt, ?C_CHECKOUTS, 1),
             {reply, Conn, State};
         empty when map_size(Monitors) < MaxSize ->
-            %% Pool has room to grow. Kick off a grow in the background
-            %% and return `busy` immediately so the caller is not held
-            %% past the gen_server:call deadline waiting for a new
-            %% connection. The caller will retry on a later event.
-            {reply, busy, grow(State0)};
+            %% Pool has room to grow. Open one connection in the
+            %% background and return `busy` so the caller retries on a
+            %% later event. grow/1 calculates demand from Pending, which
+            %% is empty for try_checkout callers, so call grow/2 directly.
+            {reply, busy, grow(1, State0)};
         empty ->
             {reply, busy, State0}
     end;
--- a/src/rabbitmq_stream_s3_api_aws_pool.erl
+++ b/src/rabbitmq_stream_s3_api_aws_pool.erl
@@ -190,10 +190,6 @@ handle_call(
             counters:add(Cnt, ?C_CHECKOUTS, 1),
             {reply, Conn, State};
         empty when map_size(Monitors) < MaxSize ->
-            %% Pool has room to grow. Open one connection in the
-            %% background and return `busy` so the caller retries on a
-            %% later event. grow/1 calculates demand from Pending, which
-            %% is empty for try_checkout callers, so call grow/2 directly.
             {reply, busy, grow(1, State0)};
         empty ->
             {reply, busy, State0}
--- a/src/rabbitmq_stream_s3_log_reader.erl
+++ b/src/rabbitmq_stream_s3_log_reader.erl
@@ -25,7 +25,7 @@ tier.
 %% and osiris_log:chunk_iterator/3 than their specs declare, causing false
 %% positives for the {offset_not_found, ...} branches and maybe_become_remote/2.
 %% Remove these suppressions once OTP 28 is required.
--dialyzer({no_match, [send_file/3, chunk_iterator/3]}).
+-dialyzer({no_match, [send_file/4, chunk_iterator/4]}).
 -dialyzer({no_unused, maybe_become_remote/2}).
 
 -define(READ_TIMEOUT, 10000).
@@ -295,10 +295,17 @@ close(#?MODULE{mode = Local}) ->
     counters:add(counter(), ?C_LOCAL_CLOSE, 1),
     ok = osiris_log:close(Local).
 
+close_deferred(undefined) -> ok;
+close_deferred(Local) -> osiris_log:close(Local).
+
+send_file(Socket, State, Callback) ->
+    send_file(Socket, State, Callback, undefined).
+
 send_file(
     Socket,
     #?MODULE{config = Config, mode = #remote{} = Remote0, verify_crc = VerifyCrc} = State0,
-    Callback
+    Callback,
+    DeferredClose
 ) ->
     case read_header(Remote0) of
         {ok,
@@ -324,6 +331,7 @@ send_file(
                     PrefixData = Callback(Header, ToSend + byte_size(HeaderData)),
                     case send(Transport, Socket, [PrefixData, HeaderData, Data]) of
                         ok ->
+                            close_deferred(DeferredClose),
                             Remote = Remote1#remote{
                                 next_offset = ChId + NumRecords,
                                 position = NextPosition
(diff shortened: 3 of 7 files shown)


lukebakken:
@the-mikedavis I have no idea what's up with the OTP install in GHA - `dialyze` passes just fine locally.

the-mikedavis (approved):

PR #189, opened 2026-06-03
Fix management UI message count for tiered streams

### LINKED ISSUE #185: Management UI message count only reflects local tier, not full stream range
## Summary

The RabbitMQ management UI reports incorrect (low, oscillating) message counts for streams with tiered storage active. The count reflects only the local tier window (~120K messages) instead of the full readable range (~9.8M messages).

## Root cause

The writer's `trigger_retention_eval` in `osiris_log.erl` (line 3293-3311) sets the `?C_FIRST_OFFSET` counter to the first offset of remaining local segments after retention evaluation:

```erlang
EvalFun = fun ({{FstOff, _}, FstTs, NumSegLeft})
                when is_integer(FstOff), is_integer(FstTs) ->
              osiris_log_shared:set_first_chunk_id(Shared, FstOff),
              counters:put(Cnt, ?C_FIRST_OFFSET, FstOff),
              ...
          end,
ok = osiris_retention:eval(Name, Dir, RetentionSpec, EvalFun),
```

The `RetentionSpec` includes the tiered storage plugin's `local_retention_fun` (prepended by `rabbitmq_stream_s3_hooks:on_retention_updated/2`). This function aggressively trims all local segments that have been fully uploaded to S3. After evaluation, only the current active segment remains, so `FstOff` reflects the local tier boundary (e.g. offset 24,224,475) rather than the stream's true first readable offset (manifest `first_offset` = 14,509,675).

`rabbit_osiris_metrics:handle_info(tick, ...)` reads `first_offset` from `osiris_counters:overview()` and computes:

```erlang
COffs = Offs + 1 - FstOffs,
rabbit_core_metrics:queue_stats(QName, COffs, 0, COffs, 0),
```

This produces the management UI's `messages` count: `24,345,132 - 24,224,475 + 1 = ~120K` instead of the correct `24,345,132 - 14,509,675 + 1 = ~9.8M`.

## Verified data from running cluster

| Metric | Value | Source |
|--------|-------|--------|
| `osiris_counters` `first_offset` | 24,224,475 | `osiris_counters:overview()` on writer node |
| Local segment file on disk | `00000000000024224475.segment` | Only local segment present |
| Manifest `first_offset` | 14,509,675 | `rabbitmq_stream_s3_manifest_first_offset` Prometheus metric |
| Manifest `next_offset` | 24,334,398 | `rabbitmq_stream_s3_manifest_next_offset` Prometheus metric |
| Management UI `messages` | ~120K | `offset + 1 - first_offset` (local window only) |
| Correct `messages` | ~9.8M | `offset + 1 - manifest_first_offset` (full stream) |

## Why `first_chunk_id` cannot simply be changed

`osiris_log_shared:first_chunk_id` (set alongside `?C_FIRST_OFFSET`) is used by `rabbitmq_stream_s3_log_reader:resolve_remote_location` to decide whether a consumer should read from the local or remote tier:

```erlang
resolve_remote_location(first, #{name := StreamId, shared := Shared}) ->
    LocalFirstOffset = osiris_log_shared:first_chunk_id(Shared),
    case rabbitmq_stream_s3_manifest_replica:get_manifest(StreamId) of
        #manifest{first_offset = RemoteFirstOffset} when RemoteFirstOffset < LocalFirstOffset ->
            %% Remote tier has older data: attach remote reader
            ...
```

If `first_chunk_id` were set to the manifest's `first_offset`, then `RemoteFirstOffset < LocalFirstOffset` would always be false, and consumers could never enter remote mode.

## Fix direction

The fix must separate the two uses of `first_offset`:

1. **Reader routing** (`first_chunk_id` / `osiris_log_shared`): must remain as the local tier's actual first offset so `resolve_remote_location` can decide local vs remote
2. **Management UI message count** (`?C_FIRST_OFFSET` / `osiris_counters`): should reflect the stream's overall first readable offset (manifest `first_offset`)

Fix: the writer's EvalFun (or an `on_retention_evaluated` hook in osiris) sets `?C_FIRST_OFFSET` to the manifest's `first_offset` instead of the local tier's first offset, while keeping `set_first_chunk_id` unchanged. This is plugin-side work: either the existing `local_retention_fun` hook is extended to also overwrite the counter, or a new osiris hook fires after retention evaluation and the plugin uses it to correct `?C_FIRST_OFFSET`.

lukebakken:
## Summary

Fixes #185. The management UI message count was reporting only the local segment window (~120K messages) instead of the full readable range (~9.8M) for streams with tiered storage active.

- Override `?C_FIRST_OFFSET` with the manifest's `first_offset` when the remote tier holds older data than the local tier
- Add `on_retention_evaluated/2` hook to osiris so the plugin gets the last word after `trigger_retention_eval` sets counters
- Writer-side and replica-side `maybe_evaluate_retention` use `min(FstOff, ManifestFirstOffset)` for their own EvalFuns
- `first_chunk_id` (reader routing) is unchanged everywhere

## Osiris dependency

Requires amazon-mq/upstream-to-osiris@b16b80d (`tiered-storage-abstractions` branch) which adds the `on_retention_evaluated/2` optional callback to `osiris_log_hooks`.

## Test plan

- [ ] Deploy to test cluster, verify management UI shows correct message count (~9.8M not ~120K)
- [ ] Verify `osiris_counters:overview()` `first_offset` matches manifest `first_offset` after retention fires
- [ ] Verify consumers still enter remote mode (reader routing unaffected)
- [ ] Run 4-hour high-throughput soak test, confirm no regressions

### DIFF
--- a/src/rabbitmq_stream_s3_hooks.erl
+++ b/src/rabbitmq_stream_s3_hooks.erl
@@ -17,6 +17,7 @@ retention is updated.
 -export([
     on_init/3,
     on_retention_updated/2,
+    on_retention_evaluated/2,
     local_retention_fun/1,
     discover/0
 ]).
@@ -83,6 +84,29 @@ on_retention_updated(Retention, #{name := Name}) ->
         Pid -> gen_server:cast(Pid, {retention_updated, Retention})
     end,
     [{'fun', local_retention_fun(StreamId)} | Retention].
+
+-doc """
+Called after retention evaluation sets the first_offset counter.
+
+Overrides `?C_FIRST_OFFSET` with the manifest's first offset when the
+remote tier holds older data than the local tier. Without this, the
+management UI reports only the local segment window as the message count.
+""".
+-spec on_retention_evaluated(counters:counters_ref(), map()) -> ok.
+on_retention_evaluated(Cnt, #{name := Name}) ->
+    StreamId = iolist_to_binary(Name),
+    case rabbitmq_stream_s3_manifest_replica:get_manifest(StreamId) of
+        #manifest{first_offset = ManifestFirst} when ManifestFirst > 0 ->
+            LocalFirst = counters:get(Cnt, ?C_OSIRIS_LOG_FIRST_OFFSET),
+            case ManifestFirst < LocalFirst of
+                true ->
+                    counters:put(Cnt, ?C_OSIRIS_LOG_FIRST_OFFSET, ManifestFirst);
+                false ->
+                    ok
+            end;
+        _ ->
+            ok
+    end.
 
 -doc """
 Discover existing osiris writers and replicas on this node and attach
--- a/src/rabbitmq_stream_s3_manifest_replica.erl
+++ b/src/rabbitmq_stream_s3_manifest_replica.erl
@@ -267,7 +267,7 @@ request_resync(StreamId, WriterNode) ->
 maybe_evaluate_retention(
     StreamId,
     #manifest{next_offset = OldManifestNextOffset},
-    #manifest{next_offset = NewManifestNextOffset},
+    #manifest{first_offset = ManifestFirstOffset, next_offset = NewManifestNextOffset},
     #state{contexts = Ctxs}
 ) when NewManifestNextOffset > OldManifestNextOffset ->
     case maps:get(StreamId, Ctxs, undefined) of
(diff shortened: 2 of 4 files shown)


the-mikedavis (changes requested):
I hate testing counters but I think the message count is so crucial that it deserves a test, probably in replica_reader_SUITE. We can emulate counting messages like the server does by subtracting last offset from first offset. (The server might do something slightly different, that's just what I have off the top of my head.)

the-mikedavis:
Other than a test I think this looks good. But I think we do need a test here to guard against future regression.

lukebakken (commented):
I looked at using `seed_log` but it bypasses the writer process and plugin pipeline entirely (writes raw segments to disk, no seshat counter registration, no replica reader, no uploads, no retention evaluation). This test needs the full pipeline running to exercise the counter being set correctly after upload -> local retention -> hook fires. The `osiris_writer:write` + `await_offset` + `?awaitMatch` pattern is the lightest way to get there with real retention firing.

Issue #191, opened 2026-06-04
Consumer close blocked by synchronous S3 reads in send_chunks loop


lukebakken:
## Summary

Closing a consumer takes 10-20+ seconds when the `rabbit_stream_reader` gen_statem is blocked in `send_chunks` serving other consumers on the same connection. With 8 parallel replay consumers, total close time reaches ~80 seconds.

## Root cause

`rabbit_stream_reader` is a `gen_statem` that processes all protocol frames sequentially. The `send_chunks` loop calls:

```
send_chunks → osiris_log_reader:send_file → rabbitmq_stream_s3_log_reader:send_file → read(RemoteReaderPid, ...) → gen_server:call(RemoteReader, #read{}, 60000)
```

Each `read` to the remote reader is a synchronous `gen_server:call` with a 60-second timeout (`?GEN_SERVER_CALL_TIMEOUT` from `rabbit.hrl`). While blocked waiting for S3 data, the gen_statem cannot process other frames — including `unsubscribe` requests that arrive on the same TCP connection.

The original `send_chunks` design assumed sub-millisecond local reads (kernel zero-copy `sendfile`). The S3 plugin introduced multi-second blocking calls into this path.

## Observed behavior

```
00:39:23.408 [rabbitmq-stream-consumer-connection-0] WARN  ServerFrameHandler - Could not find outstanding request with correlation ID 10 (ResponseFrameHandler)
00:39:23.411 [main] WARN  ConsumersCoordinator - Unexpected response code when unsubscribing from s3-high-throughput: 4 (SUBSCRIPTION_ID_DOES_NOT_EXIST) (subscription ID 1)
```

Timeline:
1. Client sends `unsubscribe` (correlationId=10) for subscription 1
2. Server gen_statem is blocked in `send_chunks` for another consumer → frame sits in TCP buffer
3. Client's 10s RPC timeout fires → removes correlationId 10 from outstanding requests
4. Server finally processes the unsubscribe, responds on correlationId 10
5. Client can't find the correlation ID → "Could not find outstanding request"
6. Client retries → server already removed the subscription → `SUBSCRIPTION_ID_DOES_NOT_EXIST`

## Impact

- ~20 seconds per affected consumer during close (10s timeout + retry)
- 80+ seconds to close 8 consumers in a replay test
- Affects production scaling events, rebalances, and graceful shutdowns
[...]

PR #194, opened 2026-06-05
Fix `{badmatch, end_of_manifest}` crash when last fragment is deleted

### LINKED ISSUE #193: Remote reader crashes with `{badmatch, end_of_manifest}` when last fragment is deleted
## Summary

The remote reader crashes with `{badmatch, end_of_manifest}` in `rabbitmq_stream_s3_remote_reader_core:next_fragment_offset/1` when the 404'd fragment is the last entry in the manifest.

## Stack trace

```
exception exit: {{{badmatch,end_of_manifest},
    [{rabbitmq_stream_s3_remote_reader_core,next_fragment_offset,1,
        [{file,"rabbitmq_stream_s3_remote_reader_core.erl"},{line,438}]},
     {rabbitmq_stream_s3_remote_reader_core,try_serve,1,[...]},
     {rabbitmq_stream_s3_remote_reader,handle_call,3,...},
     ...]},
  {gen_server,call,[<0.75514.0>,{read,63739640,...},60000]}}
```

This crashes the `rabbit_stream_reader` gen_statem (the stream protocol connection process).

## Root cause

In `remote_reader_core.erl`, the `not_found_check_range` path (line 339-342) calls `next_fragment_offset(State1)` which assumes the fragment iterator has a next entry:

```erlang
next_fragment_offset(#state{iterator = Iterator}) ->
    {ok, #fragment_ref{offset = Offset}, _} = rabbitmq_stream_s3_fragment_iterator:next(Iterator),
    Offset.
```

When the 404'd fragment is the **last** entry in the manifest, the iterator is exhausted and `next/1` returns `end_of_manifest`, causing the badmatch.

## Scenario

1. A consumer is reading from the remote tier
2. Retention deletes the last fragment in the manifest
3. The remote reader's prefetch of that fragment gets a 404
4. `try_fragment_transition` sees `next = not_found` → returns `{not_found_check_range, State}`
5. The handler calls `next_fragment_offset(State1)` to find the next fragment's offset to refresh past
6. The iterator is exhausted → `end_of_manifest` → crash

## Fix

Handle `end_of_manifest` in the `not_found_check_range` branch by falling back to the same logic as the `refresh_iterator` path — refresh past the current fragment's offset:

```erlang
{not_found_check_range, State1} ->
    case next_fragment_offset(State1) of
        {ok, NotFoundOffset} ->
            {State1, [{refresh_iterator, NotFoundOffset}]};
        end_of_manifest ->
            CurrentOffset = (State1#state.fragment_ref)#fragment_ref.offset,
            {State1, [{refresh_iterator, CurrentOffset}]}
    end;
```

This requires `next_fragment_offset/1` to return `{ok, Offset} | end_of_manifest` instead of crashing on the badmatch.

## Reproduction

Run `make DURATION=600 WARMUP=120 multi-offset-consumer` which has a rate-limited lagging consumer reading deep in the remote tier while retention actively deletes fragments. The crash was observed at 17:26:24 on `ip-10-0-1-12` during a test run.

lukebakken:
Closes #193

## Summary

- Fix crash in `remote_reader_core:next_fragment_offset/1` when the iterator is exhausted
- Handle `end_of_manifest` by refreshing past the current fragment offset, which correctly transitions to local mode downstream
- Replace all nested accessor `(State#state.fragment_ref)#fragment_ref.offset` with the existing `current_fragment_offset/1` helper
- Remove a no-op field reassignment in `try_fragment_transition`'s `{error, _}` clause

## Root cause

When the current fragment is the last in the manifest and 404s while no read is pending, `current_not_found` is set but no `refresh_iterator` is emitted. A subsequent read triggers `not_found_check_range`, which calls `next_fragment_offset/1`. That function assumed the iterator always had a next entry — when exhausted, it crashed with `{badmatch, end_of_manifest}`.

## Test plan

- [x] New test case: `last_fragment_404_no_pending_read_refreshes_past_current`
- [x] All 28 `remote_reader_core_SUITE` cases pass
- [x] Dialyzer passes
- [x] 10-minute `multi-offset-consumer` integration test (rate-limited lagging consumer + retention) — no crashes observed in server logs

### DIFF
--- a/src/rabbitmq_stream_s3_remote_reader_core.erl
+++ b/src/rabbitmq_stream_s3_remote_reader_core.erl
@@ -197,7 +197,7 @@ step(State0, {data, _RequestId, Fragment, Data, DoneOrContinue}) ->
     {State5, Effects2} = try_serve(State4),
     {State5, Effects1 ++ Effects2};
 step(State0, {request_error, _RequestId, Fragment, not_found}) ->
-    case Fragment =:= (State0#state.fragment_ref)#fragment_ref.offset of
+    case Fragment =:= current_fragment_offset(State0) of
         true ->
             %% Current fragment 404. Refresh the iterator past this offset.
             State = State0#state{current_not_found = true, requests_in_flight = #{}},
@@ -261,9 +261,8 @@ step(#state{cfg = #cfg{min_retry_delay_ms = MinDelay}} = State0, deadline_expire
     {State, [{reply, {error, timeout}}]};
 step(State0, {iterator_refreshed, end_of_manifest}) ->
     %% No new entries. Become local.
-    FragOffset = (State0#state.fragment_ref)#fragment_ref.offset,
     State = goto_next_fragment(State0),
-    {State, [{reply, {become_local, FragOffset}}]};
+    {State, [{reply, {become_local, current_fragment_offset(State0)}}]};
 step(State0, {iterator_refreshed, Iterator}) ->
     %% Iterator has been refreshed past the 404'd fragment. Reinitialize
     %% at the next available fragment.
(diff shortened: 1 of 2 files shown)


lukebakken:
## Test results (tmp/gh-191-gh-193 combined branch)

10-minute `multi-offset-consumer` test passed. Zero `badmatch` crashes, zero `end_of_manifest` errors. The fix is validated under sustained load with retention actively deleting fragments while the lagging consumer reads from the remote tier.

the-mikedavis (approved):
Nice fix & cleanup 👍

PR #195, opened 2026-06-05
Reduce remote reader deadline to 18s to unblock gen_statem (#191)

lukebakken:
Mitigates #191

## Summary

- Lower `pending_read_deadline_ms` from 50s to 18s so the remote reader's internal deadline fires before the `gen_server:call` timeout (20s). This guarantees the caller receives `{error, timeout}` via a normal reply, keeping `from` cleared before the next read arrives.
- Add a 20s `gen_server:call` timeout with `try/catch exit:{timeout, _}` as a safety net (should never fire under normal operation since the internal deadline fires first at 18s).
- The `rabbit_stream_reader` gen_statem unblocks after 18s instead of blocking for up to 60s, allowing it to process pending subscribe/unsubscribe/credit frames.

## Approach

The previous implementation attempted to handle overlapping reads (caller times out at the `gen_server:call` level while `from` is still set) by resetting buffer state (`discard_stale_pending`). This caused data corruption on the wire - chunks already sent before the timeout were re-sent from the reverted log position.

The correct fix is simpler: ensure the internal deadline always fires first. The server replies `{error, timeout}` to the caller within the call timeout window, so the caller never exits via `exit:{timeout, _}`, `from` is always cleared, and the existing `?assertEqual(undefined, State0#state.from)` assertion remains valid. No buffer resets, no overlapping reads, no duplicate delivery.

## Test plan

- [x] `remote_reader_core_SUITE` passes
- [x] Property-based tests pass
- [ ] 45-minute high-throughput integration test (pending)

### DIFF
--- a/src/rabbitmq_stream_s3_remote_reader.erl
+++ b/src/rabbitmq_stream_s3_remote_reader.erl
@@ -130,10 +130,11 @@ read_size_prometheus_format() ->
 start(Config) ->
     gen_server:start(?MODULE, Config, []).
 
-%% The gen_server:call timeout must exceed pending_read_deadline_ms so the
+%% The gen_server:call timeout must exceed PENDING_READ_DEADLINE_MS so the
 %% internal deadline always fires first and replies {error, timeout} to the
 %% caller. This avoids overlapping reads (caller times out, new read arrives
 %% while from is still set) which require unsafe buffer resets.
+-define(PENDING_READ_DEADLINE_MS, 18_000).
 -define(SEND_FILE_READ_TIMEOUT_MS, 20_000).
 
 read(Server, Offset, Bytes, Hint) ->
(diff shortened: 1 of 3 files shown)


lukebakken:
## 45-minute high-throughput test results (3-node cluster, commit ae7673e)

Zero errors on all nodes. Zero checksum failures, zero assertion crashes, zero `invalid_chunk_header`.

252M messages published, 8-consumer replay from S3 completed in 101s (96% coverage, 133-177 MiB/s).

Warnings (nodes 1 and 2 only, under pool pressure):
- 77 slow remote tier reads hitting the 18s deadline - working as designed
- 76 S3 HTTP request timeouts (15s `request_timeout_ms`) triggering retry
- 136 `pool_busy` warnings - handled via backoff
- 1 replica fell behind retention - normal resync

The 18s deadline fired on replica readers under sustained S3 pool pressure without any data corruption or assertion failures. The gen_statem unblocked and processed other frames correctly each time.

Known #191 behavior still present: `SUBSCRIPTION_ID_ALREADY_EXISTS` during consumer re-assignment (once) and unsubscribe timeouts during close (twice).

the-mikedavis (approved):

PR #197, opened 2026-06-08
S3: Allow configuring a KMS key ARN

the-mikedavis:
This enables a defense-in-depth mechanism where you can configure a bucket (through CloudFormation for example) to deny an upload which does not specify server-side-encryption (SSE). Generally you should configure your bucket to use SSE, and SSE is enabled by default since 2023 anyways. But you can prevent against the accidental upload of unencrypted data if those measures fall through.


lukebakken (approved):
🚀

PR #198, opened 2026-06-08
CLI: Add a command to trigger local retention

the-mikedavis:
`rabbitmq-streams evaluate_local_retention <stream name> [--vhost vhost]` kicks off the local retention job. This should let us debug situations where the plugin is leaving more in the local tier because of possible race conditions with retention. By evaluating retention and seeing whether or not it results in a segment file being deleted we can know whether or not retention is the issue.


lukebakken (approved):
👍

PR #200, opened 2026-06-09
hooks: Fix condition check for an empty manifest on retention eval

the-mikedavis:
This would cause spurious mismeasurements of message count (based on this first offset counter) while a queue was growing and not yet hitting remote tier retention.

Neither first or next offsets are reliable for detecting an empty manifest - the remote tier is only really empty if the manifest points to zero fragments.

This caused a pattern like so in message count:

Screenshot 2026-06-09 at 19 52 58

### DIFF
--- a/src/rabbitmq_stream_s3_hooks.erl
+++ b/src/rabbitmq_stream_s3_hooks.erl
@@ -95,15 +95,10 @@ management UI reports only the local segment window as the message count.
 -spec on_retention_evaluated(counters:counters_ref(), map()) -> ok.
 on_retention_evaluated(Cnt, #{name := Name}) ->
     StreamId = iolist_to_binary(Name),
-    case rabbitmq_stream_s3_manifest_replica:get_manifest(StreamId) of
-        #manifest{first_offset = ManifestFirst} when ManifestFirst > 0 ->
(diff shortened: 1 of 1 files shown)


src/rabbitmq_stream_s3_hooks.erl
```diff
@@ -95,15 +95,10 @@ management UI reports only the local segment window as the message count.
 -spec on_retention_evaluated(counters:counters_ref(), map()) -> ok.
 on_retention_evaluated(Cnt, #{name := Name}) ->
     StreamId = iolist_to_binary(Name),
-    case rabbitmq_stream_s3_manifest_replica:get_manifest(StreamId) of
-        #manifest{first_offset = ManifestFirst} when ManifestFirst > 0 ->
+    case rabbitmq_stream_s3_manifest_replica:get_range(StreamId) of
```
the-mikedavis, line 98: Note: this helper returns `empty` if `entries =:= <<>>`

lukebakken (approved):

the-mikedavis:
Screenshot 2026-06-12 at 16 49 28

Confirmed fixed with an identical test on a new deployment.

PR #207, opened 2026-06-17
Fix some minor correctness bugs

the-mikedavis:
This is a quick round of correctness fixes and assertions / logging flagged by the 🧞.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #208, opened 2026-06-17
Service transfers larger than the governor burst on credit

the-mikedavis:
Theoretical bug in the governor - we can smooth it out to always ensure progress by the governor.

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -88,6 +88,14 @@ stream_s3.persist_interval_ms = 2000
 # When unlimited, the governor does not pace transfers. Set to a fraction
 # of instance bandwidth on managed deployments to avoid contention with
 # other traffic. Default: unlimited.
+#
+# The ceiling is a long-run average. A single transfer larger than the
+# internal burst allowance (a fragment can exceed its target size when a
+# large chunk lands at the end) is still admitted, on credit, and the
+# governor throttles subsequent transfers to repay it; it never stalls.
+# The `governor_oversized_admissions` metric counts such transfers: a
+# persistently climbing value means the configured rate is low relative
+# to fragment sizes.
 stream_s3.max_transfer_bytes_per_sec = unlimited
 ```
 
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
Reviewed by reading the delta against `audit-fixes` (the branch is now cleanly stacked, so the only commit under review is the governor fix). I did not run the suites; CI will cover them.

The fix is correct. The admission guard `min(Tokens, Burst) =< Available` is identical to the old `Tokens =< Available` for any request at or below burst, and for an oversized request the threshold saturates at `Burst`, so admission requires a full bucket. Since `refill` caps `available` at `burst`, "full" is the only point an oversized request is admitted, after which `available` goes negative (debt); the field is an `integer()` and `refill` repays the debt from the negative value before anything else is admitted, so the long-run rate is preserved exactly. The deficit returned to a waiting caller is updated to `min(Tokens, Burst) - Available`, consistent with the new threshold.

No deadlock, and head-of-line blocking is now bounded. `drain_pending` is strict FIFO and stops on an unadmittable head. Before the fix an oversized head was unadmittable forever (the deadlock); now it is unadmittable only until the bucket refills to full, then admitted, so the item behind it drains once the debt repays. FIFO is preserved, nothing jumps the queue, and an oversized transfer does not starve others beyond the rate it legitimately consumes.

The `unlimited` path is untouched: it has its own `dispatch` clause that never touches the bucket, and `count_oversized`/`info` are only reached on the real-bucket admit branch, so `info/1` is never called on the `unlimited` atom.

The rationale is real, not theoretical: a transfer is a whole fragment, `fragment_target_size` is only a soft cut floor, and a trailing chunk batches records, so a fragment can exceed burst. The docstring captures this well.

Tests are thorough: admit-when-full, blocks-until-debt-repaid, arbitrarily-large, zero-burst-no-deadlock, and the governor end-to-end deadlock test, each of which fails on the base. The rewritten `insufficient_returns_deficit_test` is correct, since a full bucket can no longer report insufficient for an oversized request. The `governor_oversized_admissions` counter is a good operational signal for burst being set too small.

One note rather than a request: this is stacked on #207, so it should merge after it. The docs/operations.md wording ("rate is low relative to fragment sizes") and the counter help text ("burst is smaller than typical fragments") describe the same condition slightly differently; unifying them would be a nice touch but is not necessary.

LGTM.

PR #209, opened 2026-06-17
Position the fragment iterator by the requested offset inside groups

the-mikedavis:
The forward iterator positioned the root index by the caller's requested offset but descended into every group at the manifest's first_offset. A consumer starting a read inside a group, above the group's first surviving child, was therefore handed the group's first child and walked forward from there: after the first fragment boundary it received earlier, already-seen fragments, out-of-order and with duplicate offsets, and no error was raised. This triggered only for a stream large enough to hold a group entry and a subscription landing mid-group (replay from an old offset, or a lagging consumer), which is the core tiered-read case.

Position descent by max(requested_offset, first_offset) at every level, precomputed once as the iterator's start_offset. The first_offset floor still skips leading entries deleted by retention; the requested offset now selects the correct child. Reusing one value at every level is correct because the iterator only moves forward, so a group entirely after the requested offset resolves to its first child.

The fix is confined to the iterator: find_position, refresh_iterator, and resolve_first already pass the correct offset into init/3. Adds four fragment_iterator_SUITE regression tests (mid-group, between children, nested kilo-group, and mid-group with a retention floor); each fails when descent uses first_offset.

### DIFF
--- a/src/rabbitmq_stream_s3_fragment_iterator.erl
+++ b/src/rabbitmq_stream_s3_fragment_iterator.erl
@@ -30,9 +30,15 @@ as it descends into branches.
     index :: non_neg_integer(),
     stack :: [{binary(), non_neg_integer()}],
     get_group_fun :: get_group_fun(),
-    %% The manifest's first_offset. Used to skip deleted entries when
-    %% descending into groups after partial retention.
-    first_offset :: osiris:offset()
+    %% The offset to position at, at every level of the tree: the caller's
+    %% requested offset clamped up to the manifest's first_offset. The clamp
+    %% serves two roles at once. It positions descent at the requested offset
+    %% (so a read starting mid-group lands on the right child, not the group's
+    %% first child), and the first_offset floor skips leading entries that
+    %% retention has deleted (whose objects are gone). Reusing one value at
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
Reviewed by reading the diff and the surrounding iterator code.

The fix is sound. Descent now positions by `max(requested_offset, first_offset)` precomputed once and reused at every nesting level, which is correct for a forward-only iterator: it is a strict superset of the prior behavior (identical when the requested offset is at or below `first_offset`, correct when above). I traced the root, group, and nested kilo-group paths and the `partition_point`/`saturating_decr` boundary handling; offset-on-boundary, between-children, past-end, and retention-floor-above-requested all resolve correctly.

The four regression tests each fail under the old `first_offset`-only descent and pass with the fix, and the nested kilo-group case proves the fix applies at every level rather than only the root.

One minor, non-blocking note: `init_between_group_children_positions_at_containing_child` and `mid_group_descent_respects_retention_floor` stop after two `next/1` calls and do not assert the tail reaches `end_of_manifest`; the other two new tests cover full traversal, so this is not a coverage gap worth blocking on.

LGTM.

PR #210, opened 2026-06-17
Retry fatally-failed fragment uploads instead of dropping them

### LINKED ISSUE #206: Local segment freed while its fragment was never durable in S3
## Summary

A range of confirmed messages can end up in neither tier: its local segment was deleted by the local-tier cleanup (`local_retention_fun`, which frees a local segment once the manifest says it is durably tiered) while the covering fragment was never actually durable in S3. The manifest `next_offset` advanced past the missing range, so the cleanup deleted the local copy and the hole is silent. Consumers replaying the stream skip the range with no error, and the data is permanently lost.

This is not stream retention (`max-length-bytes`). Stream retention trims the oldest contiguous prefix of the whole stream, and trimmed data is legitimately gone. This hole is different: it sits in the middle of the readable range with present, consumable data on both sides, which retention cannot produce. The cause is the local-tier cleanup deleting a segment that was not yet durable in S3.

Observed in a `content-verification` integration test run on the 3-node cluster: a single fragment covering 251,891 confirmed offsets was missing from both tiers.

## Evidence

Stream `__s3-content-verify_1781652415711741774`, single producer, 256-byte messages, `max-length-bytes=500000000`, 5-minute publish then full-stream fan-out replay. Every assigned sequence was confirmed (`confirmed == maxSequence == 173493172`), so offsets 0..173493171 are all expected to be present.

The four replay consumers each independently read exactly 2,960,740 messages from `first`..tail. The readable range is `[170280541, 173493171]` (manifest `first_offset` to the committed tail) = 3,212,631 offsets, so 251,891 confirmed messages are missing.

The missing range is `[171541091, 171792982)`, verified three independent ways that all agree on 251,891.

### 1. S3 fragment chunk chains

Downloading the fragments around the hole and walking their osiris chunk headers:

```
fragment 171037046: span [171037046, 171290407)  records=253361 chunks=242
fragment 171290407: span [171290407, 171541091)  records=250684 chunks=224
fragment 171792982: span [171792982, 172045183)  records=252201 chunks=250
```

Fragment 171290407 ends cleanly at offset 171541091 (its chunk chain terminates at the expected byte boundary, the remainder is the index). The next fragment starts at 171792982. Nothing covers `[171541091, 171792982)`. The fragments on either side are contiguous with their own neighbors, so the boundaries are clean and only the one fragment is absent.

### 2. Manifest entries

Dumping the live manifest entries from the node (decoding the 34-byte entry records):

```
171037046 (fragment, size 67912364)
171290407 (fragment, size 67194064)
171792982 (fragment, size 67601868)
172045183 (fragment, size 67157264)
```

There is no entry between 171290407 and 171792982. The manifest itself does not record a fragment for the missing range, yet `manifest_next_offset` is 173298867, well past the hole. So the manifest advanced as though the range were durable when it was not.

### 3. Consumer-observed gap

The replay verifier counts confirmed sequences that were skipped (a forward jump in the per-message sequence where the skipped sequences were confirmed). It reported 1,007,564, which is 251,891 per consumer across 4 consumers. This matches the offset arithmetic and the fragment-chain gap exactly.

### This is not stream retention

Two facts rule out `max-length-bytes` retention as the cause:

- The hole is in the middle of the readable range, with consumable data on both sides: `[170280541, 171541091)` below it and `[171792982, tail]` above it both read back fine. Stream retention trims the oldest contiguous prefix and cannot leave a middle slice missing while older data survives.
- The hole at offset ~171.5M is well above the manifest `first_offset` of 170280541, which is the floor of the retained range. Retention has not reached it.

### Why local recovery is impossible

The hole is below `local_first=171713280` (the local segment on every node starts at 171713280, after the local-tier cleanup freed the earlier segments). So the missing range is gone from local disk and absent from S3: permanent loss.

## Read path behavior

The read path itself behaved correctly given the broken durable state. The `become_local` transition and iterator refresh both worked (`result=ok`, no `deadline_expired`, no `{error, timeout}` on any node). The reader simply delivered what was durably present, which is why the gap is silent rather than an error. This is a durability bug in the upload/retention path, not a read-path bug. It is distinct from #203 (read-path stall on a cold pool), which did not occur in this run.

## Root cause

Not yet determined, but the local-tier cleanup is downstream of the defect, not the cause. `local_retention_fun` deletes a local segment only when its offset is at or below the manifest `next_offset` (the last durably-tiered offset plus one), so it deleted the local copy of the missing range only because the manifest already claimed that range was durable. The defect is that `next_offset` advanced to 173298867, past the fragment for `[171541091, 171792982)`, even though that fragment was never durable in S3 and has no manifest entry.

So the question is why `next_offset` advanced past a fragment that was not uploaded and recorded. Candidate areas to investigate:

- Upload failure handling: a fragment upload that failed or was dropped without being retried, while `next_offset` still advanced past it
- A lost or mis-applied manifest edit for that fragment, with the offset accounting advancing regardless
- An ordering or atomicity gap between recording a fragment in the manifest and advancing `next_offset`

## Reproduction

1. Publish at high throughput for 5 minutes with `max-length-bytes=500000000` (single producer, 256-byte messages, sequential IDs in the body) so fragments are uploaded to S3 and the local-tier cleanup frees the uploaded local segments
2. Stop publishing and wait for confirms to settle
3. Replay the whole stream from `first` and verify every confirmed sequence is delivered exactly once

The `content-verification` integration test does this. Note that its gap counter sums across consumers and so over-reports the magnitude by the consumer count (reported 1,007,564 for a true 251,891); this is a known test reporting bug and does not affect the finding.

## Severity

Silent, permanent loss of confirmed messages. A publisher receives confirmation, the data is later unreadable, and nothing surfaces an error to producer or consumer.

the-mikedavis:
A non-transient fragment upload failure took the fatal branch of transfer_failed/3, which removed the fragment from the in-flight queue and drained the fragments behind it. drain_completions then advanced the manifest next_offset over those later fragments, past an offset range that was never made durable in S3. The manifest recorded the range as tiered, the local-tier cleanup (which frees segments at or below next_offset) removed the only remaining copy, and the result was a silent, permanent hole in the middle of the stream with consumable data on both sides (#206).

A confirmed fragment must never be abandoned. transfer_failed/3 now keeps the fragment at its place in the queue and retries it: immediately for transient errors, and after a backoff delay (upload_retry_delay_ms, default 1000ms) for non-transient ones. The pipeline stalls at the failed offset until the fragment is durable rather than advancing past it, and the delayed retry logs a warning naming the stalled offset so the stall is observable.

Add a defensive assert_contiguous/2 guard in apply_fragment so a fragment that does not begin exactly where the manifest ends fails loudly instead of silently recording a gap; on restart the reader re-resolves the durable manifest rather than persisting the hole.

Update the core suite: the cases that codified the old drop-and-gap behavior now assert that a fatal failure retries without advancing the manifest, and a new end-to-end case proves a fragment that fails then succeeds yields a contiguous manifest.

*Issue #, if available:* Likely fixes #206.

### DIFF
--- a/src/rabbitmq_stream_s3_config.erl
+++ b/src/rabbitmq_stream_s3_config.erl
@@ -36,6 +36,7 @@ lives here. Callers use these functions instead of calling
     task_retry_delay_exponent/0,
     verbose_logging/0,
     segment_upload_timeout/0,
+    upload_retry_delay_ms/0,
     retention_task_timeout/0,
     tick_timeout_milliseconds/0,
     max_transfer_bytes_per_sec/0,
@@ -160,6 +161,13 @@ verbose_logging() ->
 segment_upload_timeout() ->
     application:get_env(?APP, segment_upload_timeout, 45_000).
 
+%% Delay before retrying a fragment upload that failed with a non-transient
+%% error. The upload pipeline stalls at the failed offset until the retry
+%% succeeds, so this bounds how often a persistently failing upload is retried.
+-spec upload_retry_delay_ms() -> non_neg_integer().
+upload_retry_delay_ms() ->
+    application:get_env(?APP, upload_retry_delay_ms, 1000).
+
 -spec retention_task_timeout() -> non_neg_integer().
 retention_task_timeout() ->
     application:get_env(?APP, retention_task_timeout, 60_000).
@@ -213,6 +221,7 @@ defaults_test_() ->
         ?_assertEqual(2, task_retry_delay_exponent()),
         ?_assertEqual(false, verbose_logging()),
         ?_assertEqual(45_000, segment_upload_timeout()),
+        ?_assertEqual(1000, upload_retry_delay_ms()),
         ?_assertEqual(60_000, retention_task_timeout()),
         ?_assertEqual(5000, tick_timeout_milliseconds()),
         ?_assertEqual(false, verify_crc_on_read()),
--- a/src/rabbitmq_stream_s3_replica_reader.erl
+++ b/src/rabbitmq_stream_s3_replica_reader.erl
@@ -835,7 +835,7 @@ execute_effect(
     {resubmit_transfer_delayed, Ref, _StreamId, Dir, Meta, Reason}, #state{cfg = Cfg} = State0
 ) ->
     StreamId = Cfg#cfg.stream,
-    Delay = application:get_env(rabbitmq_stream_s3, upload_retry_delay_ms, 1000),
+    Delay = rabbitmq_stream_s3_config:upload_retry_delay_ms(),
     ?LOG_WARNING(
         "~ts fragment upload at offset ~b failed with non-transient error ~p; "
         "retrying in ~bms. The upload pipeline and local-tier cleanup are "
--- a/src/rabbitmq_stream_s3_replica_reader.erl
+++ b/src/rabbitmq_stream_s3_replica_reader.erl
@@ -455,6 +455,12 @@ handle_info({transfer_result, Ref, {error, Reason}}, #state{core = Core0} = Stat
     State1 = on_transfer_result(Ref, {error, Reason}, State0),
     {Core, Effects} = rabbitmq_stream_s3_replica_reader_core:transfer_failed(Ref, Reason, Core0),
     {noreply, execute_effects(Effects, State1#state{core = Core})};
+handle_info({retry_transfer, Ref, Dir, Meta}, #state{cfg = #cfg{stream = StreamId}} = State) ->
+    %% A previously failed fragment upload is being retried after its backoff
+    %% delay. The in-flight gauges were already restored when the delayed
+    %% resubmit effect was executed, so only re-submit the upload here.
+    submit_upload(Ref, Dir, StreamId, Meta),
+    {noreply, State};
 handle_info({group_upload_result, {ok, Uid}}, #state{core = Core0, group_mon = Mon} = State0) ->
     demonitor(Mon, [flush]),
     State1 = on_group_upload_completed(State0),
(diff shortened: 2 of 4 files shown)


lukebakken:
The AI and I reviewed the fix against the read/upload paths and ran the core suite locally (61/61). The central fix is correct: keeping the failed fragment at its place in the in-flight queue means `drain_completions` stalls at the incomplete head and `next_offset` cannot advance past a non-durable range. That closes #206, and the new test cases (`fatal_failure_does_not_drain_subsequent`, `fatal_failure_retry_then_complete_no_gap`, `fatal_front_does_not_drain_buffered_completions`) encode the regression well. The `assert_contiguous/2` guard is a good belt-and-suspenders addition.

A nice property worth calling out: because the stalled fragment holds the head of the queue, `next_offset` does not advance, so the local-tier cleanup (which only frees segments at or below `next_offset`) cannot delete the stalled fragment's local segment. The stall protects its own source data, so a retry can always re-read it.

That said, I think the retry/backoff strategy needs work before this is production-sound. None of these affect the data-loss fix itself.

## The transient path retries with no delay

`transfer_failed/3` resubmits transient errors (`{http, 500}`, `{http, 503}`, `timeout`) immediately via `resubmit_transfer`, with no delay. Under the default governor rate of `unlimited`, submissions execute immediately, so a persistently failing upload becomes a tight loop: fail, resubmit, execute, fail, bounded only by the S3 round-trip. A 503 is specifically "slow down", and an immediate resubmit is the opposite of what it asks for.

This is also a regression in spin behavior: before this change the fatal branch dropped the fragment and stopped; now nothing escapes the queue, so a persistent transient error spins unthrottled. And it spins silently, since the warning log is only on the delayed (non-transient) path. Each attempt re-reads and re-uploads the whole fragment (~67 MiB at the sizes in the repro), so this is sustained S3 PUT and disk-read cost, not a cheap busy-loop.

## No exponential backoff or cap on either path

Both paths are constant-rate: transient is 0ms, non-transient is a flat `upload_retry_delay_ms`. The plugin already has exponential-backoff machinery in config (`task_retry_delay_constant`, `task_retry_delay_exponent`, `task_retry_delay_max_ms`, default cap 5000ms) used for task retries. I would reuse it for both upload-retry paths rather than introducing a separate flat delay, so there is one retry convention and a sane ceiling under a sustained outage.

## `upload_retry_delay_ms` default of 1000ms looks too low for the path it governs

This value governs only the non-transient path. By the PR's own classification those are errors that do not self-heal quickly (`checksum_mismatch`, `enoent`, a 403, etc.), yet it retries them every second, each retry re-reading and re-uploading the full fragment. If the cause is genuinely persistent, that is ~3600 full ~67 MiB re-upload attempts per hour that cannot succeed. For a non-transient classification the delay should be longer than the transient case and should back off, not sit at a flat 1s, which is even below the existing `task_retry_delay_max_ms` ceiling of 5000ms. The 0ms-transient / 1000ms-non-transient split also reads as arbitrary. If a fixed value is kept, I would raise the default to at least match the existing 5000ms cap and document why fixed is acceptable here.

## Classification is narrow

`is_retriable/1` treats only 500, 503, and timeout as transient. `{http, 429}` (Slow Down / too many requests) is a canonical S3 throttling response and currently falls into the `_ -> false` catch-all, so it takes the non-transient backoff path. That is safe now that nothing is dropped, but it is mislabeled, and many genuinely transient conditions (connection reset, DNS blips, gateway errors that do not match exactly) will also be treated as non-transient. Worth widening, or at least a comment on why the set is intentionally minimal.

## Observability

An indefinitely stalled upload surfaces only as a warning log line per retry. A counter (e.g. fatal upload retries, or a gauge for "stalled at offset") would make a stuck pipeline visible in Prometheus rather than only in logs.

the-mikedavis:
I'll merge this down now so we can build on top of it, and make an issue for improving retry in the upload path

PR #211, opened 2026-06-17
Make remote group retention idempotent across cycles

the-mikedavis:
A group object is immutable: once fragments are factored into a group and uploaded, the object always lists all of its original children. When remote retention partially consumes a group, build_group_retention_result leaves the root group entry and the group object untouched (it only advances first_offset and total_size), so the already-removed children remain in the object.

On the next retention cycle eval_group_entries re-downloaded the full group and counted from the first child again, with no reference to first_offset. It therefore re-processed the already-removed leading children every cycle, which:

  * double-subtracted their sizes from total_size, drifting it below the true durable size (silently disabling max_bytes retention) or negative (which crashes the persist, since total_size is serialized as an unsigned field);
  * regressed first_offset below the real data floor, pointing it at children whose objects had already been deleted.

eval_group_entries now drops the already-consumed prefix (children whose offset is below the manifest's first_offset) before evaluating, so each cycle sizes, deletes and advances over only the surviving children. First-cycle behavior is unchanged because first_offset equals the group's first child offset until a partial consumption advances it.

Add manifest_SUITE covering multi-cycle group retention (previously uncovered): the basic partial case, idempotent max_bytes that does not regress first_offset, max_age that no longer drives total_size negative, and a partial followed by a full consumption that removes the group entry. The three multi-cycle cases fail without the fix.

### DIFF
--- a/src/rabbitmq_stream_s3_manifest.erl
+++ b/src/rabbitmq_stream_s3_manifest.erl
@@ -344,7 +344,17 @@ maybe_eval_group_retention(#manifest{entries = Entries} = Manifest, Specs, Now, 
     end.
 
 %% Evaluate retention on fragment entries within a group.
-eval_group_entries(Manifest, GroupEntries, GroupRef, Specs, Now) ->
+%%
+%% A group object is immutable, so after an earlier cycle partially consumed
+%% this group it still lists the children those cycles already removed. Skip
+%% that already-consumed prefix (children below the manifest's first_offset)
+%% before evaluating. Otherwise retention re-counts and re-deletes them and
+%% double-subtracts their sizes from total_size on every subsequent cycle,
+%% which drifts total_size below the true durable size (silently disabling
+%% max_bytes retention) or negative (crashing the persist under max_age), and
+%% regresses first_offset below the real data floor.
+eval_group_entries(Manifest, GroupEntries0, GroupRef, Specs, Now) ->
+    GroupEntries = drop_consumed_children(GroupEntries0, Manifest#manifest.first_offset),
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
Reviewed by reading the diff and the surrounding manifest/retention code.

The fix is correct. Dropping children whose offset is strictly below the manifest `first_offset` before evaluating removes the per-cycle re-processing of already-consumed leading children, which was the source of the `total_size` double-subtraction and the `first_offset` regression. The `<` boundary is right: `first_offset` is the first surviving offset, so a child at exactly `first_offset` must be kept. The negative-`total_size` crash path is genuinely closed (the field is serialized as a 70-bit unsigned value, so a negative would crash the persist), and first-cycle behavior is provably unchanged because `first_offset` equals the group's first child offset until a partial consumption advances it.

The multi-cycle tests cover the basic partial case, idempotent max_bytes with no first_offset regression, max_age with no negative total_size, and partial-then-full consumption; the three multi-cycle cases fail without the fix. The property test cross-checks accounting against an independent live-fragment model.

Non-blocking: nested kilo/mega-group multi-cycle retention is not exercised by the new tests. This PR does not change nested behavior, so it is not a regression, but a multi-level case would be worth adding in a follow-up.

LGTM.

the-mikedavis:
> Non-blocking: nested kilo/mega-group multi-cycle retention is not exercised by the new tests. This PR does not change nested behavior, so it is not a regression, but a multi-level case would be worth adding in a follow-up.

PR coming soon with fairly large changes for recursive retention :wink:

PR #213, opened 2026-06-17
Resolve a timestamp seek to the fragment holding the boundary chunk

the-mikedavis:
Fragment selection for a {timestamp, Ts} spec used the non-strict predicate Ts >= last_ts, which treats a fragment whose last_ts equals Ts as wholly before the target and skips it. The chunk at that timestamp is the fragment's last chunk and lives there, so the seek resolved to the next fragment and silently dropped the offsets at the boundary.

The chunk-level resolver in find_index_position/2 already uses the strict Ts > T, snapping a timestamp to the first chunk at or after it. The fragment level is only routing to the fragment that contains that chunk, the first fragment whose last_ts >= Ts, which requires the strict skip predicate Ts > last_ts. This also matches osiris, whose segment search treats a segment as too old only under strict Ts > EndTs.

The new find_fragment_timestamp_boundary_test fails on the old predicate and passes on the fix. The prop_find_fragment_timestamp property encoded the old behaviour: its generator separates fragments by +1, so its in-gap probe degenerated onto last_ts and asserted the skipped fragment. It now probes a strictly-interior gap point and asserts both fragment boundaries resolve to the containing fragment.

### DIFF
--- a/src/rabbitmq_stream_s3_log_reader.erl
+++ b/src/rabbitmq_stream_s3_log_reader.erl
@@ -797,7 +797,12 @@ find_fragment(Entries, Spec, GetGroup) ->
             {offset, Offset} ->
                 fun(?ENTRY(O, _FTs, _LTs, _K, _Sz, _Uid, _)) -> Offset >= O end;
             {timestamp, Ts} ->
-                fun(?ENTRY(_O, _FTs, LTs, _K, _Sz, _Uid, _)) -> Ts >= LTs end
+                %% Strict, like the chunk-level predicate in
+                %% `find_index_position/2'. A fragment whose last_ts equals Ts
+                %% contains the chunk at that timestamp, so it must not be
+                %% skipped. Compare Osiris, which treats a segment as too old
+                %% only when Ts is strictly greater than its end timestamp.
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
Reviewed by reading the diff, the surrounding fragment-selection code, and the osiris segment search.

The fix is correct. Changing the fragment-level skip predicate from `Ts >= last_ts` to `Ts > last_ts` makes selection resolve to the first fragment whose `last_ts >= Ts`, which is the fragment that actually contains the boundary chunk (that chunk is the fragment's last chunk). This matches the strict `Ts > T` already used by the chunk-level resolver in `find_index_position/2`, and matches osiris, whose segment search treats a segment as too old only under strict `Ts > EndTs`. Before-first, after-last, on-first_ts, and single-fragment cases all resolve correctly through the clamp.

On the modified property test: I checked that the old property was encoding the bug rather than asserting correct behavior. The generator separates fragments by at least +1, so the old in-gap probe degenerated onto `last_ts` and asserted the buggy skip to the next fragment. The new property re-gates the in-gap branch on a genuine gap and probes a strictly-interior point, and the within-range branch now probes the boundary timestamps directly. The modification is justified. The new EUnit boundary test fails on the old predicate and passes on the fix.

LGTM.

Issue #214, opened 2026-06-17
Audit and standardize log levels before 1.0.0


lukebakken:
## Summary

Before the 1.0.0 release, audit every log statement in the plugin and settle on a consistent, deliberate level for each. Logging has accreted during development and bug investigations, and the levels are not consistent: some operationally useful events are at debug, some noisy or investigation-only events were at info or warning, and there is no agreed policy for what belongs at each level.

## Background

While investigating the cold-pool read stall (#203) and the durability hole (#206), a fair amount of diagnostic logging was added across the read and upload paths (`refresh_iterator`, `become_local` transitions, `pool_busy` retries, deadline expiry, read-retry attempts, pool `try_checkout`/`gun_up` state). That logging was moved to debug level so it does not spam the default level, but the choice was uniform rather than considered per statement. Some of it is genuinely useful operational signal and might warrant info or a metric; some is pure investigation scaffolding that could be removed.

This issue is the deliberate pass over all of it.

## Scope

- Inventory every `?LOG_DEBUG`, `?LOG_INFO`, `?LOG_WARNING`, `?LOG_ERROR` (and any `?LOG_NOTICE`) call in `src/`
- Decide the right level for each against a written policy, for example:
  - error: an operator must act; data or availability is at risk
  - warning: unexpected but handled; worth noticing if it recurs
  - info: significant lifecycle events at a rate that will not flood logs
  - debug: per-request or per-retry detail useful only when investigating
- Remove statements that are pure investigation scaffolding with no lasting value
- Promote anything that is really an operational signal to a metric/counter rather than a log line, where a metric fits better (the plugin already has per-stream counters)
- Check message quality: format strings match arguments, no `~ts` on non-string terms, stream IDs included where useful, no unbounded terms logged
- Confirm hot paths do not log per-request at info or above

## Notes

- The debug-level diagnostic logging mentioned above lands via the #203 read-path fix branch. This issue covers the whole plugin, not only those lines.
- Worth doing alongside or after the metrics are finalized, since some log lines should become counters instead.

lukebakken:
Concrete data for this audit, from a 2-hour high-throughput run against
`streams-tiered-storage-combined` (3 x
`m7g.8xlarge`, plugin active, broker log level debug, finished 2026-08-03
19:34:19).

The run was clean: zero `[error]`, `[critical]`, `[alert]` and `[emergency]`
lines and no Erlang crash atoms on any node. But it produced 3153-3505
[...]

PR #215, opened 2026-06-17
Harden remote tier reads against a cold connection pool

### LINKED ISSUE #203: Remote tier reads stall permanently on cold connection pool
## Summary

Consumers reading from the remote tier could stall permanently when the HTTP connection pool was empty (for example after the idle timeout drained all connections, or at the start of a replay against a cold pool). A burst of `pool_busy` retries consumed the read deadline, the read timed out, and the consumer was then parked waiting for an event that never arrived, so it never recovered even though S3 was healthy.

This is fixed. The fix has landed on `feature/integration-test-2` and been validated on the 3-node cluster: a 5-minute publish followed by a full-stream fan-out replay (4 consumers each independently replaying `first`..tail over the S3-resident history) completes with every consumer reaching the tail, zero `deadline_expired`, and zero `{error, timeout}` propagated to `send_chunks`.

## Root cause

Two compounding issues.

### 1. `pool_busy` used the same exponential backoff as network errors

In `rabbitmq_stream_s3_remote_reader_core`, `pool_busy` shared a clause with `timeout`/`stream_error`/`connection_error`, so the retry sequence was 1s, 2s, 4s, 8s, 16s. But `pool_busy` means "the pool is growing, a TLS handshake is in progress, try again in ~100-200ms." A single `pool_busy` at a fragment transition consumed 1s+ of the read deadline waiting for a retry that should have taken 100ms. A fragment transition issues several prefetch requests, so the cumulative delay quickly exceeded the deadline and it fired.

### 2. After a timeout, nothing retried without an external event

When `send_file` returned `{error, timeout}` to the `rabbit_stream_reader` gen_statem, the handler logged "Error while sending chunks: timeout" and returned the consumer unchanged. The gen_statem then waited for either a new `osiris_offset` notification (requires new data to be committed) or a client CREDIT frame (requires the client to have received messages first). If publishing had stopped, neither event occurred and the consumer was stuck permanently.

## Fix

1. **`pool_busy` no longer shares the exponential backoff clause.** It has its own handler in `rabbitmq_stream_s3_remote_reader_core` with a mild capped backoff (`pool_busy_delay`, starting at 100ms, growing to a small cap, reset on success) instead of the network-error sequence. Pool growth completes in ~100-200ms, so many retries now fit inside the read deadline.

2. **The log reader retries remote reads transparently.** `rabbitmq_stream_s3_log_reader` retries a remote read up to `READ_RETRY_ATTEMPTS` (3) with a `READ_RETRY_DELAY_MS` (500ms) pause before propagating `{error, timeout}`, so a transient timeout no longer parks the consumer waiting for an offset notification that never arrives.

3. **Supporting tuning:**
   - `general_pool_min_size` default 0 to 2, so an idle broker keeps warm connections and a replay does not start against a fully cold pool
   - `pending_read_deadline_ms` 18s to 40s
   - `SEND_FILE_READ_TIMEOUT_MS` 20s to 45s (must exceed the pending read deadline so the internal deadline fires first and replies `{error, timeout}` rather than the caller's `gen_server:call` timing out)

## Validation evidence

Latest run, per node, for the replay stream:

```
node-162: deadline_expired=0 timeout_reply=0 pool_busy=0
node-21:  deadline_expired=0 timeout_reply=0 pool_busy=6
node-12:  deadline_expired=0 timeout_reply=0 pool_busy=19
```

`pool_busy` still occurs under the concurrent read pressure of multiple consumers re-reading the same fragments. That is expected: it signals "the pool is growing, retry shortly," not saturation. The fix absorbs these events with the mild backoff and read retry rather than letting them cascade into a deadline expiry and a permanent stall. Zero `deadline_expired` and zero `{error, timeout}` replies across all nodes confirm the consumers complete cleanly.

## Original reproduction (for the record)

1. Publish at high throughput for 5 minutes with `max-length-bytes=500MB` so local retention trims segments that were uploaded to S3
2. Stop publishing
3. Start replay consumers at offsets that require remote tier reads
4. Consumers received some initial data then stalled permanently

This was surfaced by the `content-verification` integration test, which also had two unrelated bugs of its own: it sliced the offset range across consumers as if a stream were a work queue, and it gated pass/fail on the management message count, which lags retention. Those were fixed separately so the test now exercises a real independent-fan-out replay, and the read-path behavior above was confirmed against the corrected test.

Earlier log evidence of the stall, before the fix:

```
remote_reader start_request: pool_busy key=.../00000000000172781846... frag=172781846
[multiple consecutive pool_busy for the same fragment]
remote_reader deadline_expired for stream ... (requests_in_flight=0)
remote_reader replying {error, timeout} for stream ...
send_file: data read returned timeout (next_offset=172949993 position=45070844)
Error while sending chunks: timeout
```

## Still open

The design question stands: `rabbit_stream_reader` assumes `send_file` never blocks for meaningful time (local sendfile is effectively instant), and the remote tier violates that assumption. Fix 2 is a bounded workaround (retry within the synchronous call), not the asynchronous `send_file` redesign tracked in #191. While a read is retrying, the gen_statem is still blocked and cannot process heartbeats, credits, or other consumers on the same connection, and if S3 is genuinely down the consumer blocks for the full deadline times retry count before giving up.

The diagnostic logging added during this investigation is currently at INFO and should be downgraded or removed before release.

## Related

- #191 `rabbit_stream_reader` gen_statem blocks during synchronous S3 reads
- #195 Reduce `pending_read_deadline_ms` to 18s (the change that originally surfaced this issue; the deadline has since been raised to 40s)

lukebakken:
## Summary

Consumers reading from the remote tier could stall permanently when the HTTP connection pool was empty (after the idle timeout drained all connections, or at the start of a replay against a cold pool). A burst of `pool_busy` retries consumed the read deadline, the read timed out, and the consumer was then parked waiting for an offset notification that never arrived. Fixes #203.

## Changes

- **Separate `pool_busy` from the network-error exponential backoff.** `pool_busy` means the pool is growing and a TLS handshake is in progress, which completes in ~100-200ms. It now uses a mild capped backoff (100ms growing to a small cap, reset on success) instead of the 1s/2s/4s/8s sequence, so many retries fit inside the read deadline.
- **Retry remote reads transparently in the log reader.** A read that times out is retried a few times with a short pause before the error reaches `send_chunks`, so a transient timeout no longer parks the consumer.
- **Keep a warm pool.** `general_pool_min_size` defaults to 2 so a replay does not start against a fully cold pool.
- **Raise the read timeouts.** `pending_read_deadline_ms` to 40s and the `send_file` read timeout to 45s (the latter must exceed the former so the internal deadline fires first and replies `{error, timeout}` rather than the caller's `gen_server:call` timing out).
- **Add debug-level diagnostics** across the read and pool paths so pool contention and tier transitions are observable without noise at the default level.

## Validation

`gmake test-build`, `gmake fmt-check`, config eunit (30/30), and the `remote_reader_core` (28/28) and `log_reader` (10/10) suites pass. Validated end to end on a 3-node cluster: a 5-minute publish followed by a full-stream fan-out replay completes with every consumer reaching the tail, zero `deadline_expired`, and zero `{error, timeout}` propagated to `send_chunks`. `pool_busy` still occurs under concurrent read pressure but is absorbed by the backoff and read retry rather than cascading into a stall.

## Notes

The log levels added here are uniformly debug; a deliberate per-statement audit across the whole plugin is tracked in #214.

The broader design question, that `rabbit_stream_reader` assumes `send_file` never blocks and the remote tier violates that, is not addressed here. The read retry is a bounded workaround, not the asynchronous `send_file` redesign tracked in #191.

### DIFF
--- a/src/rabbitmq_stream_s3_log_reader.erl
+++ b/src/rabbitmq_stream_s3_log_reader.erl
@@ -341,6 +341,11 @@ send_file(
                             Err
                     end;
                 {next_fragment, Offset} ->
+                    ?LOG_DEBUG(
+                        "send_file: data read returned next_fragment at ~b",
+                        [Offset],
+                        ?DOMAIN
+                    ),
                     State = State0#?MODULE{
                         mode = Remote1#remote{
                             next_offset = Offset,
@@ -351,22 +356,34 @@ send_file(
                 {become_local, _} ->
                     close_deferred(DeferredClose),
                     counters:add(counter(), ?C_REMOTE_CLOSE, 1),
-                    case init_local_reader(Remote1#remote.next_offset, Config) of
+                    case become_local_init(Remote1#remote.next_offset, Config) of
                         {ok, State} ->
                             send_file(Socket, State, Callback, undefined);
                         {error, _} = Err ->
                             Err
                     end;
                 end_of_stream ->
+                    ?LOG_DEBUG(
+                        "send_file: data read returned end_of_stream"
+                        " (next_offset=~b position=~b)",
+                        [Remote1#remote.next_offset, Position],
+                        ?DOMAIN
+                    ),
                     close_deferred(DeferredClose),
                     {end_of_stream, State0#?MODULE{mode = Remote1}};
                 {error, timeout} = Err ->
+                    ?LOG_DEBUG(
+                        "send_file: data read returned timeout"
+                        " (next_offset=~b position=~b)",
+                        [Remote1#remote.next_offset, Position],
+                        ?DOMAIN
+                    ),
                     Err
             end;
         {become_local, _} ->
             close_deferred(DeferredClose),
             counters:add(counter(), ?C_REMOTE_CLOSE, 1),
-            case init_local_reader(Remote0#remote.next_offset, Config) of
+            case become_local_init(Remote0#remote.next_offset, Config) of
                 {ok, State} ->
                     send_file(Socket, State, Callback, undefined);
                 {error, _} = Err ->
@@ -453,7 +470,7 @@ chunk_iterator(
                 {become_local, _} ->
                     close_deferred(DeferredClose),
                     counters:add(counter(), ?C_REMOTE_CLOSE, 1),
-                    case init_local_reader(Remote1#remote.next_offset, Config) of
+                    case become_local_init(Remote1#remote.next_offset, Config) of
                         {ok, State} ->
                             chunk_iterator(State, Credit, undefined, undefined);
                         {error, _} = Err ->
@@ -468,7 +485,7 @@ chunk_iterator(
         {become_local, _} ->
             close_deferred(DeferredClose),
             counters:add(counter(), ?C_REMOTE_CLOSE, 1),
-            case init_local_reader(Remote0#remote.next_offset, Config) of
+            case become_local_init(Remote0#remote.next_offset, Config) of
                 {ok, State} ->
                     chunk_iterator(State, Credit, undefined, undefined);
                 {error, _} = Err ->
@@ -550,13 +567,21 @@ send(ssl, Socket, Data) ->
     | {end_of_stream, #remote{}}
     | {error, timeout}.
 read_header(#remote{shared = Shared, next_offset = NextOffset} = Remote) ->
+    LastChunkId = osiris_log_shared:last_chunk_id(Shared),
+    CommittedChunkId = osiris_log_shared:committed_chunk_id(Shared),
     CanReadNext =
-        osiris_log_shared:last_chunk_id(Shared) >= NextOffset andalso
-            osiris_log_shared:committed_chunk_id(Shared) >= NextOffset,
+        LastChunkId >= NextOffset andalso
+            CommittedChunkId >= NextOffset,
     case CanReadNext of
         true ->
             read_header1(Remote);
         false ->
+            ?LOG_WARNING(
+                "Remote read_header returning end_of_stream:"
+                " next_offset=~b last_chunk_id=~b committed_chunk_id=~b",
+                [NextOffset, LastChunkId, CommittedChunkId],
+                ?DOMAIN
+            ),
             {end_of_stream, Remote}
     end.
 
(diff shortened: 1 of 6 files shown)


the-mikedavis:
> a TLS handshake is in progress, which completes in ~100-200ms

This seems a bit long to me, no? Especially when running on EC2 I would expect a super quick handshake and very high throughput to/from S3

src/rabbitmq_stream_s3_log_reader.erl
```diff
@@ -550,13 +567,21 @@ send(ssl, Socket, Data) ->
...
     case CanReadNext of
         true ->
             read_header1(Remote);
         false ->
+            ?LOG_WARNING(
+                "Remote read_header returning end_of_stream:"
+                " next_offset=~b last_chunk_id=~b committed_chunk_id=~b",
+                [NextOffset, LastChunkId, CommittedChunkId],
+                ?DOMAIN
+            ),
```
the-mikedavis: Should this be debug or info level instead? This should be a normal condition during catch-up, right?

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -224,16 +228,26 @@ step(
     NextDelay = min(RetryDelay * 2, MaxDelay),
     State = State0#state{retry_delay = NextDelay, requests_in_flight = #{}},
     {State, [{set_timer, RetryDelay}]};
+step(
+    #state{pool_busy_delay = Delay} = State0,
+    {request_error, _RequestId, _Fragment, pool_busy}
+) ->
+    %% Pool is growing — a connection will be ready within ~100-200ms (TLS
+    %% handshake time). Use a mild backoff (100, 200, 500, 500...) so we
```
the-mikedavis:
```suggestion
    %% handshake time). Use a mild backoff (100, 200, 400, 500, 500...) so we
```

I think `min(Delay * 2, 500)` should hit 400 and then cap out at 500, no?

the-mikedavis:
I pushed up some core tests covering the backoff behavior since we can run those cheaply. I wouldn't be offended if you force-pushed to drop that change though. These tests will probably see updates soon as we work through #191.

src/rabbitmq_stream_s3_log_reader.erl
lukebakken: Yeah debug, I'll change it.

the-mikedavis (approved):
Looks good to me 👍

lukebakken:
Good catch, you are right. The "~100-200ms" was an unmeasured estimate, and on same-region EC2 to S3 the handshake should be much faster than that. I have lowered the backoff floor to 25ms (same doubling and 500ms cap, so the sequence is now 25, 50, 100, 200, 400, 500, 500...) and reworded the comment to stop asserting a specific handshake duration. Starting at 25ms catches a freshly grown connection sooner on a cold-pool read while still backing off enough to avoid spinning if the pool genuinely cannot grow.

On the tests: thanks, I am keeping them. They cover the backoff behavior directly and were cheap to update for the new floor, so I would rather keep the coverage in place until #191 reworks this area than drop it now.

PR #216, opened 2026-06-17
Add content-verification integration test and fan-out replay

lukebakken:
## Summary

Add a `content-verification` integration test and correct both replay tests to model real-world fan-out consumption.

## Changes

- **`content-verification` test.** Publishes messages with sequential IDs, replays the whole stream from `first` independently from every consumer, and verifies ordering, no duplicates, and no gaps in the confirmed sequences. Catches corruption, reordering, and silent message skips. This is the test that surfaced #206.
- **`ConfirmedSequences`.** A paged bitset recording which sequences the broker confirmed, so a real gap is distinguished from a hole left by an unconfirmed send. A settle-wait before closing producers ensures the bitset is complete before replay reads it.
- **Fan-out correction to both replay tests.** Every consumer now reads the whole stream from `first` to the tail rather than a sliced offset range. A stream is a replayable log, not a work queue, so a plain stream cannot be partitioned across cooperating consumers by offset slice; that is what super streams are for. The previous slicing made `content-verification` stall and let `high-throughput` pass for the wrong reason.
- **`ReplayProgress`.** Shared per-consumer progress helpers extracted from the two tests.
- **Docs.** A `test/integration/README.md` documenting the consumption patterns so the distinction is not re-derived, and the plugin `AGENTS.md` now points at the subsystem docs, the glossary, the investigations directory, and the integration-test README.

## Validation

`mvn compile` is clean. The `content-verification` test passes against current `main` (which includes the #206 durability fix from #210).

## Notes

This branch is test and documentation only; it shares no files with the #203 read-path fix.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,9 +8,20 @@ Read the relevant documentation before modifying any module:
 
 - `docs/README.md` — overview and reading guide
 - `docs/DEVELOPMENT.md` — building, testing, formatting
+- `docs/GLOSSARY.md` — terminology used throughout the plugin and its docs
 - `docs/concepts.md` — streaming primitives and how the plugin extends them
 - `docs/architecture.md` — how the pieces fit together
 - `docs/conventions.md` — patterns and conventions for writing code here
+
+When working on a specific subsystem, also read its dedicated document first:
+
+- `docs/read-path.md` — the consumer read path, including the remote-to-local tier transition (`become_local`)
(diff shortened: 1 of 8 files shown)


the-mikedavis (approved):
👍

Also I added a doc to make the invariants explicit in https://github.com/amazon-mq/rabbitmq-stream-s3/blob/main/docs/invariants.md, like the one covered by this test

PR #217, opened 2026-06-17
Recurse remote retention through nested group objects

the-mikedavis:
Remote retention assumed the oldest root entry was a group of fragments. When a stream grew enough to form kilo-groups or mega-groups (groups of groups), the oldest entry could be one of those, and the deletion-ref collection hard-matched fragment entries, so it crashed on the group children it found instead. The crash was caught by the retention task, so retention silently made no progress against a deeply nested prefix and the remote tier grew without bound. Group entries also carry no size, so max_bytes could not measure how much a group represented.

Replace the single-level group evaluation with a recursive one that walks the manifest tree to fragment granularity. It computes the new first_offset f' as a leaf boundary (descending groups, summing leaf sizes for max_bytes and using the contiguity of timestamps for max_age), then deletes exactly the objects whose interval lies in [f, f') -- the fragments below f' and every group object all of whose descendants are below f' -- skipping the prefix below the current first_offset so retention stays idempotent at every level, and splices only the wholly-consumed leading root entries. Group objects remain immutable. A group fetch error aborts the cycle cleanly rather than crashing.

Extend manifest_SUITE with kilo- and mega-group cases (partial, max_bytes, full consumption, and multi-cycle idempotence) and make the prop_SUITE retention property also generate kilo-group trees; the property fails without this change.

Note: review for https://github.com/amazon-mq/rabbitmq-stream-s3/pull/211 caught this indirectly.

### DIFF
--- a/include/rabbitmq_stream_s3.hrl
+++ b/include/rabbitmq_stream_s3.hrl
@@ -106,7 +106,7 @@
 %% * first timestamp (8)
 %% * first last timestamp (8)
 %% * size (9)
-%% = 41 bytes
+%% = 49 bytes
 -define(MANIFEST_HEADER_SIZE, 49).
 
 -define(MANIFEST(FirstOffset, NextOffset, FirstTs, FirstLastTs, TotalSize, Entries), <<
--- a/src/rabbitmq_stream_s3_manifest.erl
+++ b/src/rabbitmq_stream_s3_manifest.erl
@@ -324,152 +324,261 @@ build_retention_result(#manifest{entries = Entries} = Manifest, NumEntries) ->
     },
     {Edit, Refs}.
 
-%% Evaluate retention within the first group entry, if present.
+%% Evaluate retention when the oldest root entry is a group of any kind.
 maybe_eval_group_retention(_Manifest, _Specs, _Now, undefined) ->
     unchanged;
 maybe_eval_group_retention(#manifest{entries = Entries} = Manifest, Specs, Now, GetGroupFun) ->
     case Entries of
-        <<Offset:64/unsigned, _FTs:64/signed, _LTs:64/signed, Kind:8/unsigned, _Size:40/unsigned,
-            Uid:32/unsigned,
-            _Rest/binary>> when Kind =/= ?MANIFEST_KIND_FRAGMENT ->
-            GroupRef = #group_ref{offset = Offset, kind = Kind, uid = Uid},
-            case GetGroupFun(GroupRef) of
-                {ok, GroupEntries} ->
-                    eval_group_entries(Manifest, GroupEntries, GroupRef, Specs, Now);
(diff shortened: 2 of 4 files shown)


lukebakken (approved):
Reviewed the algorithm by hand, ran an adversarial pass over the deletion logic, and ran the suites locally (manifest 9/9, prop 30/30, including the new nested cases). This is a careful, correct extension of the single-level fix from #211 to arbitrary tree depth.

The structure is sound: compute the new first_offset `Hi` as a leaf boundary (most-aggressive policy wins, with the timestamps taken from the winning boundary), then delete exactly the objects whose interval lies in `[Lo, Hi)` and splice the wholly-consumed leading root entries. I checked the parts that, if wrong, would lose data or leak storage:

- Idempotency (the #211 bug class) holds at every level: `boundary_age`, `boundary_bytes`, and `collect_deletions` all gate fragments on `Offset >= Lo`, so an already-consumed prefix is never re-counted or re-deleted, and `total_size` cannot drift or go negative. `multi_tier_idempotent` pins this on a kilo-group: cycle two removes only `[200]`, not `[0, 100, 200]`, and `total_size` lands at 200 rather than drifting.
- Bytes accounting is exact: `Bytes` is summed in the same clause that adds a fragment to `Refs`, group entries contribute zero, and the deleted set is a subset of fragments at or above `Lo`, so `total_size + (-Bytes)` stays non-negative.
- Group deletion is safe: a group object is deleted only when all its descendants are below `Hi` (`CAll`) and at least one was newly deleted this cycle (`CRefs` non-empty), so a group with a surviving descendant is never deleted, and a now-empty group is not re-deleted.
- `count_whole_root` and `collect_deletions` use the same predicate over the same immutable group data via the same `Get`, so the spliced root prefix and the deleted objects cannot disagree (no dangling root entry, the #206 class). A group fetch error throws and aborts the whole cycle to `unchanged` rather than recording a partial result.

Two non-blocking notes:

1. The cycle descends the tree twice per evaluation (`collect_deletions` then `count_whole_root`), each re-fetching intermediate kilo/mega group objects via `Get`, and the first-group ETS cache only caches plain groups. Correctness is fine; this is extra S3 fetches per retention cycle on deep trees, worth a follow-up if it shows up in cost.

2. `include/rabbitmq_stream_s3.hrl` has a stale comment (around lines 108-110) describing the group header as 41 bytes while the code uses 49 consistently. Pre-existing, not from this PR, but adjacent and worth a one-line fix while here.

One thing worth keeping in mind rather than a request: the whole scheme rests on `first_offset` always equaling a real surviving fragment offset and group objects being immutable and complete. That invariant holds for retention itself; if any future non-retention path could set `first_offset` off a real boundary, the `Offset >= Lo` reasoning that protects both bytes accounting and group deletion would need re-checking.

LGTM.

PR #218, opened 2026-06-17
Use double-layer supervision for replica reader process

the-mikedavis:
Replica readers were direct children of a single simple_one_for_one supervisor, sharing one restart-intensity budget across the whole node. That shared budget couples unrelated streams: one reader that crash-loops on bad state, or several readers crashing in the same window, can exhaust the budget and take down every replica reader on the node.

Interpose a per-stream supervisor (rabbitmq_stream_s3_stream_sup) between the factory and the worker:

    rabbitmq_stream_s3_replica_reader_sup   (simple_one_for_one factory)
    └── rabbitmq_stream_s3_stream_sup        (one per stream)
          └── rabbitmq_stream_s3_replica_reader   (the worker)

Each stream now owns its own restart budget, so a poison stream parks alone and its siblings keep uploading.

The per-stream supervisors are temporary, so the factory never restarts them and never spends its own budget on their termination. The worker is transient and significant, and the per-stream supervisor sets auto_shutdown => any_significant: when the reader stops normally (its osiris writer went down on leadership transfer or stream deletion) the now-childless supervisor shuts itself down instead of idling. A genuine crash is restarted within the per-stream budget; only once that budget is exhausted does the stream park, awaiting reconciliation by discovery or membership reconciliation.

start_child still returns the reader worker pid and still reports {error, {already_started, Pid}} for a duplicate stream, preserving the single-layer contract. stop_child maps a reader pid back to its enclosing per-stream supervisor and terminates that.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -35,14 +35,30 @@ rabbitmq_stream_s3_sup (one_for_one)
 │     Per-node transfer pacing. Accepts fragment upload submissions from
 │     replica readers, paces them via a token bucket, spawns tasks, and
 │     reports completions back.
-└── rabbitmq_stream_s3_replica_reader_sup (simple_one_for_one, temporary)
-      └── rabbitmq_stream_s3_replica_reader (per-stream)
-            Owns the full upload lifecycle for one stream: drain committed
-            chunks, assemble fragments, submit to governor, apply completions
-            in order, persist manifest, broadcast edits, evaluate retention.
+└── rabbitmq_stream_s3_replica_reader_sup (simple_one_for_one)
+      Factory. Its dynamic children are per-stream supervisors, started
+      `temporary` so the factory never restarts them.
+      └── rabbitmq_stream_s3_stream_sup (one per stream)
+            Per-stream supervisor. Owns its own restart-intensity budget
+            and auto-shuts-down when its reader exits normally.
+            └── rabbitmq_stream_s3_replica_reader (per-stream worker)
+                  Owns the full upload lifecycle for one stream: drain
+                  committed chunks, assemble fragments, submit to governor,
+                  apply completions in order, persist manifest, broadcast
+                  edits, evaluate retention.
 ```
 
 The supervisor init also performs one-time setup: creates the seshat counter group, initializes the API backend, sets the osiris hooks (`log_hooks` and `log_reader`), registers the Khepri deletion trigger, and creates the process registry ETS table.
+
+### Replica reader fault isolation
+
+Replica readers are supervised in two layers so that one stream's failure stays confined to that stream.
+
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
Reviewed the supervision design against the OTP 27 semantics and traced the start/stop and failure paths, including an adversarial pass over the races. The design is sound and the documentation is excellent. I have no blocking issues, but I would like a couple of things addressed or at least acknowledged before merge.

What I verified:

- OTP legality: the per-stream supervisor is `one_for_one` with a `transient` + `significant` child and `auto_shutdown => any_significant` (legal: significant requires a non-permanent restart, and significant children are only forbidden under simple_one_for_one). The factory is `simple_one_for_one` with a `temporary`, non-significant `type => supervisor` child. The `start_link/1` arity chain (factory appends `[Args]` -> stream_sup `start_link(Args)` -> reader `start_link([Args])`) is correct.
- start_child no-orphan: a concurrent duplicate start loses at the registry's `insert_new`, the loser reader exits `normal` before its init runs (so it does not clobber the winner's registry entry), and the loser's per-stream supervisor self-terminates during its own init rather than leaking. The registry re-check correctly maps the nested start error back to `{error, {already_started, Pid}}`, preserving the single-layer contract. Both production callers (`on_init`, `attach_writer`) match `{ok, _}` and discard the pid, so the `reader_pid/1` lookup is never trusted to be non-undefined on a path that matters.
- Budget exhaustion cleans up: the reader traps exits and its `terminate/2` unregisters from the registry; shutdown is a 5000ms timeout (trappable), not brutal_kill, so on budget exhaustion the registry entry is already clear and a later discovery/on_init can re-create the stream. No stale-entry deadlock.
- The factory's 3/10 is effectively inert: simple_one_for_one start failures do not consume restart intensity, and temporary children are never restarted, so a node-wide start-failure storm cannot cascade to the factory.

Non-blocking comments:

1. No tests. The diff adds no test coverage, and the existing tests only exercise the unchanged external contract (`start_child`/`stop_child`/already_started), not the failure-isolation behavior that is the whole point of the change. I would like at least one test that proves isolation: start two streams, drive one reader to crash past 5/10 so its per-stream supervisor parks (assert its registry entry is cleared and the factory survives), and assert the sibling stream's reader is untouched and still registered. A second cheap test that two concurrent same-stream starts yield exactly one `{ok, _}` and one `{already_started, _}` with no orphaned per-stream supervisor would lock in the race analysis.

2. Doc inaccuracy. `docs/architecture.md` says a parked stream is re-attached by "discovery (on plugin start) or membership reconciliation." Reader re-creation is driven only by the osiris writer `on_init` hook and plugin-start `discover` (`start_child`/`attach_writer` are referenced only in `hooks.erl`); the membership reconciliation path does not re-create readers. Please drop the membership-reconciliation half or correct it.

3. Behavioral point to confirm (tied to 2): once a stream parks or auto-shuts-down, there is no active reconciler. Re-tiering happens only when the writer's `osiris_log:init/2` runs again or the plugin is re-enabled. If a writer incarnation can keep running on this node after its reader parked, that stream silently stops tiering until the next writer re-init. The PR documents this as intended ("awaiting reconciliation"); I just want to confirm the team is comfortable that the only triggers are writer re-init and plugin enable, with no periodic sweep.

The two-layer structure itself is the right call and the module docs are clear and accurate (apart from comment 2). Approving since the design is correct and the contract is preserved; please follow up on the isolation tests (1) and fix the doc line (2), and confirm the re-attach behavior (3).

PR #219, opened 2026-06-17
Make remote retention and root rebalance mutually exclusive

the-mikedavis:
Remote retention and a root rebalance both rewrite the manifest's leading entries, so they must never run concurrently (the Single mutator invariant). The code already enforced one direction: persist_complete does not evaluate retention while a rebalance is in flight. The other direction was missing.

maybe_start_persist short-circuits on both rebalance_in_flight and retention_in_flight, but maybe_start_rebalance only checked the former. So an async remote-retention evaluation (the group-download path) could be in flight when a drain reached the rebalance threshold, starting a rebalance alongside it. When the retention edit then landed it rewrote the leading prefix the rebalance was factoring, and group_upload_complete reconstructed its {Kind, Pos, Len} from the now-shifted manifest, crashing in pending_rebalance.

The manual CLI trigger (evaluate_remote_retention) made it worse: it calls maybe_evaluate_remote_retention directly, bypassing the persist_complete gate, so an operator could start retention in the middle of a rebalance.

Close both directions:

- maybe_start_rebalance now defers while retention_in_flight, mirroring maybe_start_persist. An async retention in flight blocks a rebalance from starting, so the two never overlap and a retention edit is always applied to a manifest no rebalance has changed underneath it
- maybe_evaluate_remote_retention skips evaluation while a rebalance is in flight, gating the manual CLI trigger the same way the automatic post-persist path is already gated. Nothing is computed or deleted; retention is re-evaluated by the next persist_complete after the rebalance finishes
- retention_complete and retention_failed re-check rebalance once the flag is clear, so a rebalance deferred while retention was in flight is not stranded until the next drain

Add a rebalance_in_flight/1 accessor for the shell guard, and core tests for the deferral and both re-check paths.

### DIFF
--- a/src/rabbitmq_stream_s3_replica_reader.erl
+++ b/src/rabbitmq_stream_s3_replica_reader.erl
@@ -1023,16 +1023,35 @@ update_counter(Cnt, FstOff, NumSegLeft) ->
 maybe_evaluate_remote_retention(_Manifest, [], _StreamId, State) ->
     State;
 maybe_evaluate_remote_retention(Manifest, Retention, StreamId, #state{core = Core0} = State) ->
-    inc(State, ?C_REMOTE_TIER_RETENTION_EVALUATIONS, 1),
-    Now = erlang:system_time(millisecond),
-    %% First try without group download (handles fragments-only case synchronously).
-    case rabbitmq_stream_s3_manifest:evaluate_remote_retention(Manifest, Retention, Now) of
-        unchanged ->
-            %% No leading fragments to remove. If the first entry is a group,
-            %% spawn an async task to download it and evaluate retention within.
-            maybe_spawn_group_retention(Manifest, Retention, Now, StreamId, State);
-        {Edit, Refs} ->
-            on_remote_retention(Edit, Refs, StreamId, Core0, State)
+    case rabbitmq_stream_s3_replica_reader_core:rebalance_in_flight(Core0) of
+        true ->
+            %% A rebalance is rewriting the manifest's leading entries. Remote
+            %% retention rewrites the same prefix, so evaluating it now would
+            %% compute an edit against a manifest the rebalance is about to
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
Reviewed the diff, traced the concurrency invariant in both directions against the gate functions, and ran the core suite (67/67, including the three new cases).

The fix is correct and symmetric. The single-mutator invariant over the manifest's leading entries was enforced only in the persist direction (`persist_complete` skips retention while a rebalance is in flight); this closes the other direction:

- `maybe_start_rebalance` now defers while `retention_in_flight`. I checked all four callers, including the `drain_completions` path that produced the original `pending_rebalance` crash, and they all funnel through this gated function, so the gate is comprehensive.
- `maybe_evaluate_remote_retention` skips while `rebalance_in_flight`, closing the manual CLI trigger that bypassed the `persist_complete` gate.
- `retention_complete` and `retention_failed` both re-check rebalance once the flag clears, so a deferred rebalance is not stranded until the next drain.

I confirmed the gate actually arms when the race needs it: `maybe_spawn_group_retention` calls `retention_started`, so `retention_in_flight` is true for the whole async group-download window, which is exactly the interval during which a drain could otherwise start a concurrent rebalance.

The tests cover all three core behaviors (defer while in flight, re-check on complete, re-check on fail), each asserting both the emitted effect and `rebalance_in_flight`, and each would fail without the fix. The `rebalance_deferred_while_retention_in_flight` case is a clean mirror of the existing `persist_complete_defers_retention_during_rebalance`.

One non-blocking note: there is no test for the shell-side `maybe_evaluate_remote_retention` CLI gate (the core suite cannot reach it; it would need a `replica_reader_SUITE` integration case). The core gating it relies on is covered, so this is a minor coverage gap rather than a concern.

LGTM.

PR #221, opened 2026-06-17
Follow-up to #218: correct re-attach doc and test two-layer supervision

lukebakken:
## Summary

Follow-up to the review of #218 (two-layer replica reader supervision). Addresses the two non-blocking items raised there.

## Changes

- **Doc fix.** `docs/architecture.md` said a parked replica reader is re-attached by "discovery or membership reconciliation", but membership reconciliation does not re-create readers. Re-attachment happens only via the writer `on_init` hook (its `osiris_log` is re-initialized) or `discover/0` at plugin start. The doc now states those two triggers and notes there is no periodic reconciler.
- **Tests.** Two cases covering the new supervision shape:
  - `two_layer_supervision_structure` asserts the factory's child is a per-stream supervisor and the reader worker lives one level below it.
  - `duplicate_start_child_no_orphan_supervisor` asserts a second `start_child` for an already-registered stream returns `{error, {already_started, Pid}}` and leaves no orphaned per-stream supervisor under the factory.

## Not included

The per-stream restart-budget isolation test (one stream parks, sibling survives) needs a deterministic reader crash that runs `terminate/2`. Inducing that requires test-only crash injection in the production reader, so it is left for a follow-up rather than adding crash scaffolding to the reader.

## Validation

`gmake test-build` and `gmake fmt-check` pass; both new cases pass (`ct-replica_reader`).


the-mikedavis (approved):
Nice 👍

PR #224, opened 2026-06-17
Retry a transient group fetch instead of becoming local

the-mikedavis:
On the remote read path, advancing the fragment iterator can descend into a group object, which is fetched synchronously from S3. The iterator already distinguishes a transient fetch error (`{group_fetch_failed, _}`) from a group deleted by retention (`not_found`, which it skips) and from the genuine end of the manifest (`end_of_manifest`). The core ignored that distinction: three sites collapsed `{error, {group_fetch_failed, _}}` to `become_local` (or to `end_of_manifest`, which then becomes local).

That is a Tier overlap violation. The group is part of the remote tier the consumer is reading, so a transient S3 error must not route the consumer to the local tier, which may have already trimmed that range. The fallback is silent: the consumer reads missing or wrong data with no error.

Route `{group_fetch_failed, _}` to a backoff retry instead, at all three sites:

- the forward fragment transition (`try_fragment_transition`)
- probing the next entry after a 404 (`next_fragment_offset`)
- advancing a refreshed iterator (`step({iterator_refreshed, _})`)

The retry keeps the pending read, so the existing pending-read deadline bounds the loop and surfaces `{error, timeout}` if S3 stays unavailable, rather than an unbounded stall or a silent local fallback. With the group-fetch error no longer producing `become_local` from `try_serve`, that branch becomes dead and is removed; `become_local` now comes only from a genuinely exhausted manifest.

Add core tests for the transition and refreshed-iterator sites and for the backoff-and-re-attempt behaviour.

### DIFF
--- a/src/rabbitmq_stream_s3_remote_reader_core.erl
+++ b/src/rabbitmq_stream_s3_remote_reader_core.erl
@@ -310,9 +310,15 @@ step(State0, {iterator_refreshed, Iterator}) ->
             },
             {State1, Effects} = start_current_request(State),
             {State1, [{reply, {next_fragment, Offset}} | Effects]};
-        _ ->
+        end_of_manifest ->
             %% Iterator exhausted after refresh. Become local.
-            {State0, [{reply, {become_local, current_fragment_offset(State0)}}]}
+            {State0, [{reply, {become_local, current_fragment_offset(State0)}}]};
+        {error, {group_fetch_failed, _Reason}} ->
+            %% A group object could not be fetched while advancing the
+            %% refreshed iterator. Transient S3 error, not end of manifest:
+            %% retry rather than routing to a local tier that may lack the
+            %% data.
+            retry_group_fetch(State0)
     end.
 
 %% @doc Returns the pending read, if any.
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
Reviewed the three call sites and the backoff. Correct and well-tested.

The fix rests on the fragment iterator distinguishing a retention-deleted group (not_found, which next/1 skips) from a transient fetch error ({error, {group_fetch_failed, _}}), and it does: fragment_iterator next/1 skips on not_found and surfaces {group_fetch_failed, Reason} for any other error. Routing that error to a local tier was the bug, since the group is part of the remote tier the consumer is reading, so a local fallback could serve missing or wrong data.

All three sites that previously collapsed the group-fetch error to become_local or end_of_manifest now route to retry_group_fetch: the forward fragment transition, the next-entry probe after a 404, and advancing a refreshed iterator. retry_group_fetch mirrors the existing network-error backoff exactly (emit the current delay, store the doubled delay, clear requests_in_flight), so it starts at min_retry_delay_ms and caps at max_retry_delay_ms. The pending read is kept, so the read deadline bounds the loop and surfaces {error, timeout} if S3 stays unavailable, rather than an unbounded stall or a silent local fallback. Removing the now-dead become_local branch from try_serve is correct: become_local now comes only from a genuinely exhausted manifest.

The three new core tests cover the transition site, the refreshed-iterator site, and the exponential backoff progression. CI is green.

PR #226, opened 2026-06-17
Retry manifest resolution instead of resolving empty on a transient error

the-mikedavis:
On a manifest-replica cache miss, resolve_manifest read the metadata store and the object store, and collapsed any failure to an empty manifest. Two of those collapses are dangerous: when Khepri references a manifest at some revision but the object-store GET fails with a transient error (not a 404), and when the metadata-store lookup itself errors. In both cases the stream is not empty; we just could not read it. Returning an empty manifest makes the writer believe the remote tier is empty and re-tier from the local log, orphaning the real remote objects (a Local authority hazard). The collapse was also silent: the manifests_resolved_empty counter ticked for it exactly as for a genuinely new stream, so it could not be told apart, and there was no log.

Distinguish the cases and, for the dangerous ones, retry rather than resolve empty:

- a genuinely absent metadata node (Khepri not_found) or a manifest object that has not been written yet (object-store 404) still resolves to empty: these are correct for a new stream
- a transient metadata-store error, a caught store exception, or a non-404 object-store error now returns {retry, Reason}: resolve_and_start logs a warning, increments a new manifest_resolution_failures counter, and reschedules via the existing retry_resolve self-message, leaving the remote tier untouched

retry_resolve already existed for transient init_data_reader failures, so this reuses a non-crashing retry path rather than crashing the reader, which on a shared supervisor could cascade across streams during a store blip. With the dangerous collapse gone, manifests_resolved_empty now means a genuinely empty stream.

The store/object classification is split into pure classify_store_result/2 and classify_object_result/3 functions, covered by inline eunit tests. The three resolution call sites (initial continue, retry_resolve, reinitialize) share the new resolve_and_start/1.

### DIFF
--- a/src/rabbitmq_stream_s3_replica_reader.erl
+++ b/src/rabbitmq_stream_s3_replica_reader.erl
@@ -95,19 +95,20 @@ returned by the functional core module.
 -define(C_BYTES_DRAINED, 20).
 -define(C_BYTES_PERSISTED, 21).
 -define(C_REMOTE_TIER_RETENTION_FAILURES, 22).
+-define(C_MANIFEST_RESOLUTION_FAILURES, 23).
 %% Counters above; gauges below. The macro ?NODE_COUNTERS filters by
 %% type, and seshat requires counter indexes to be 1..N sequential. Add
 %% new counters before this divider and renumber the gauges that follow.
--define(C_TRANSFERS_IN_FLIGHT, 23).
--define(C_LAST_PERSIST_TIMESTAMP_MS, 24).
--define(C_MANIFEST_FIRST_OFFSET, 25).
--define(C_MANIFEST_NEXT_OFFSET, 26).
--define(C_MANIFEST_FIRST_TIMESTAMP_MS, 27).
--define(C_REMOTE_BYTES, 28).
--define(C_REMOTE_MESSAGES, 29).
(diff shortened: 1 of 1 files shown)


lukebakken (approved):
Reviewed the classification logic and all three resolution call sites. This is a correct and well-scoped fix.

The five-way classification is sound:

- cache hit returns the manifest, refreshing the revision from Khepri but tolerating a failed refresh (the persist CAS catches a stale revision, so keeping the cached one is safe)
- Khepri not_found and object-store 404 both resolve to an empty manifest, which is correct for a genuinely new stream or one that has not persisted a manifest yet
- a Khepri error, a caught store exception, and a non-404 object-store error all return {retry, Reason} rather than collapsing to empty, which is the actual fix: treating a hidden manifest as empty would let the writer re-tier from the local log and orphan the real remote objects

Reusing the existing retry_resolve self-message means a store blip reschedules rather than crashing the reader, which matters on the shared supervisor. Logging at WARNING (not ERROR) with the dedicated manifest_resolution_failures counter as the alertable signal is the right call given the one-second retry cadence. Unifying the three call sites (initial continue, retry_resolve, reinitialize) into resolve_and_start/1 is a nice cleanup, and the manifests_resolved_empty semantic now means what it says.

The five inline eunit tests cover every classification branch. CI is green.

PR #228, opened 2026-06-17
Recover from an upload stall when retention trims an un-uploaded segment

### LINKED ISSUE #225: User retention trimming a segment before upload permanently stalls the manifest
## Summary

When the user's stream retention (`max-length-bytes`) deletes a local segment before the plugin has uploaded the covering fragment to S3, the upload pipeline stalls permanently at that offset. The fragment's source segment returns `enoent`, the non-transient retry loop introduced by #210 retries it forever, and the manifest `next_offset` never advances past the missing segment. Every fragment behind the stall point uploads to S3 but is orphaned (never recorded in the manifest), and consumers can read only the small prefix below the stall before `become_local` jumps them to the local tail, skipping the entire range in between.

This is a distinct failure mode from #206, and in a sense the consequence of its fix. #206 was silent, permanent loss caused by `next_offset` advancing *past* a non-durable range. PR #210 correctly prevents that advancement, so the manifest is now honest. But the new behavior on a permanently missing segment is an unrecoverable stall plus a large inaccessible range, rather than the "accept the gap and jump forward" behavior the architecture doc describes for user retention outrunning uploads.

## Environment

- 3-node cluster, plugin commit `07b7d4c` (latest main, includes the #210 fix `28bdcaa`)
- `content-verification` integration test: single producer, 256-byte messages, `max-length-bytes=500000000`, 300s publish then full-stream fan-out replay with 4 consumers

## Observed result

The test confirmed 150,246,340 messages. Each of the 4 replay consumers read exactly 2,686,204 messages from `first`..tail, then stopped:

```
Replay complete: each consumer read 2686204 messages in 66s
Integrity: corrupt=0 duplicates=0 gaps=590240544 out-of-order=0
GAP DETECTED: 590240544 confirmed sequences were never delivered
```

The per-consumer gap is `150246340 - 2686204 = 147560136`, and `147560136 * 4 = 590240544`, matching the reported total (the gap counter sums across consumers; the true gap is 147,560,136).

### Manifest is stalled, not advanced

Live manifest on the writer node, queried via `rabbitmq_stream_s3_manifest_replica:get_manifest/1`:

```
first_offset=0 next_offset=1764541 total_size=473019292 num_entries=7
entry_offsets=[0, 251278, 505353, 759301, 1011348, 1262496, 1512980]
```

`next_offset` is frozen at 1,764,541 with only 7 fragment entries. This is the opposite of #206, where `next_offset` advanced past a non-durable hole. Here it correctly refuses to advance.

### The stalled fragment's segment was deleted by user retention

The upload at offset 1,764,541 fails continuously, beginning seconds into the run and still failing 11 minutes after the test ended:

```
18:39:18 [warning] __s3-content-verify_... fragment upload at offset 1764541 failed with non-transient error
  {open_failed, ".../00000000000000000000.segment", enoent};
  retrying in 1000ms. The upload pipeline and local-tier cleanup are stalled at this offset until the fragment is durable in S3.
```

The segment `00000000000000000000.segment` no longer exists on any node. The lowest local segment on all three nodes is `00000000000149324677.segment`. Because the plugin's `local_retention_fun` only frees segments at or below the manifest `next_offset` (1,764,541), it cannot have deleted the segment covering offset 1,764,541. The deletion was the user's `max-length-bytes=500000000` retention, which trims the oldest contiguous prefix independent of upload progress. At ~150 MB/s ingress (550k msg/s * 256B) against a ~500 MB local window (~3 seconds), retention trimmed the segment before its fragment could upload.

### Fragments behind the stall are orphaned in S3

S3 holds 515 fragments for the stream, spanning offsets 0 up to 149,929,435, but only the first 7 (offsets 0..1,512,980) are referenced by the manifest. The fragment that would start at the stalled offset 1,764,541 is absent (S3 jumps 1,512,980 -> 2,016,226), and everything above it is durable but unreferenced. Because completions apply in cut order, the ~508 later fragments can never be applied to the manifest while 1,764,541 is stuck.

### Consumer read path behaves correctly given the stalled manifest

The remote reader served fragments through the manifest extent, then transitioned to local:

```
become_local transition for stream '__s3-content-verify_...':
  next_offset=1764541 local_first=149324677 local_committed=150246165 local_last=150246165 result=ok
```

So each consumer read [0, ~1,764,541) from the remote tier and then jumped to `local_first=149324677`, skipping offsets [1,764,541, 149,324,677). No read-path error; the gap is silent from the consumer's perspective. This is a durability/upload-path bug, not a read-path bug.

### Spread across the run

82 distinct offsets logged `open_failed`/`enoent` upload failures over the run (the pipeline kept attempting fragments behind the stall as new segments rolled and were themselves trimmed), but the manifest stall is anchored at the first one, 1,764,541.

## Why this is not #206

- #206: `next_offset` advanced past a non-durable range, producing a silent hole in the *middle* of the remote tier with consumable data on both sides and false durability in the manifest. That signature is verifiably absent here.
- This issue: `next_offset` is frozen *behind* durable data. The manifest is honest about what is durable. The loss of access comes from a permanent stall plus orphaned fragments, not from false durability.

PR #210 fixed #206 by never abandoning a confirmed fragment and stalling at the failed offset instead. That is correct when the segment is still present and the failure is transient. It is not correct when user retention has *permanently* deleted the source segment: there is no copy of offset 1,764,541 left anywhere, so retrying forever cannot make progress, and the stall makes the bulk of the stream inaccessible and strands hundreds of uploaded fragments.

## Code trace

The mechanism is confirmed end to end in the source (`rabbitmq_stream_s3_replica_reader.erl` and `rabbitmq_stream_s3_replica_reader_core.erl`).

1. The upload re-reads the segment from disk on every attempt. `submit_upload/4` builds `Fun = fun() -> upload_fragment(Dir, StreamId, Meta) end`, and `upload_fragment` -> `stream_spans/4` opens the segment file by path each time. `Meta` carries a path plus byte positions, not the bytes themselves, so once the file is deleted, `file:open` returns `{error, enoent}`, wrapped as `{error, {open_failed, SegFile, enoent}}`. Retrying re-opens the same missing path and can never succeed.

2. The core classifies `enoent` as non-transient and retries forever. `transfer_failed/3` calls `is_retriable(Reason)`, which returns `true` only for `{http, 500}`, `{http, 503}`, and `timeout`. `{open_failed, _, enoent}` is therefore non-transient and takes the `resubmit_transfer_delayed` branch, which keeps the fragment at the head of the in-flight queue and re-arms `{retry_transfer, ...}` after `upload_retry_delay_ms` (default 1000ms), indefinitely.

3. The manifest is correctly pinned. Because the failed fragment stays at the head of the in-flight queue, `drain_completions/1` peeks the front, finds it incomplete, and stops, so `next_offset` never advances. This is #210 working as intended (the `assert_contiguous/2` guard is the belt-and-suspenders backstop). The cost is that an unrecoverable failure becomes a permanent stall.

### The correct recovery already exists, but only at reader init

`start_reading0/1` already implements the "accept the gap and jump forward" behaviour the architecture doc describes, in the data-reader open path:

```erlang
{error, {offset_out_of_range, {LocalFirst, _}}} when LocalFirst > StartOffset ->
    %% Local retention deleted data the remote tier never received.
    %% Discard the remote manifest ... and restart from the local log's first offset.
```

That branch discards the manifest, jumps `next_offset` to `LocalFirst`, and resumes. The problem is that it only fires when *opening* the data reader at reader startup. The failure in this run happened mid-stream during *upload*, while the log was already open and draining, so the trimmed-segment condition surfaces as `open_failed`/`enoent` in `stream_spans/4` (the upload path), which has no equivalent recovery and simply loops.

There is also a `missing_file` catch in `start_reading0/1` that retries immediately, for the genuinely transient case of a segment deleted between listing and opening during a roll. Any fix must preserve that distinction: a bare `enoent` can be transient during a segment roll, so the permanent case must be identified by the offset being below the live local log's first offset, not by `enoent` alone.

## Possible directions

- Route the upload-path failure to the same recovery `start_reading0/1` already performs: when the segment is permanently gone (its offset is below the live local log's first offset), advance `next_offset` past the trimmed range and resume, accepting the lost range, rather than retrying forever. Distinguish this from a transient `enoent` during a segment roll, which must still retry.
- Surface the permanent stall more loudly than a repeating warning (the pipeline being wedged for the lifetime of the stream is arguably an error/alarm condition).
- Consider whether the plugin should bound how far user retention can outrun uploads, or otherwise coordinate with stream retention so confirmed data is not trimmed before it is durable.

## Reproduction

1. Publish at high throughput (~550k msg/s, 256-byte messages) for several minutes with `max-length-bytes=500000000`, so user retention trims local segments faster than the plugin uploads their fragments
2. Stop publishing, wait for confirms to settle
3. Replay the whole stream from `first` and verify every confirmed sequence is delivered

The `content-verification` integration test does this. Expected once fixed: either no gap, or a bounded gap only at the oldest trimmed prefix (never a frozen manifest with the bulk of the stream inaccessible and uploaded fragments orphaned).

## Severity

A single segment trimmed before upload permanently wedges the entire remote tier for that stream: the manifest stops advancing, the vast majority of confirmed messages become inaccessible to replaying consumers, and hundreds of uploaded fragments are stranded. No error is surfaced to producers or consumers.

lukebakken:
Fixes #225.

When the user's stream retention (max-length-bytes) deletes a local segment before the plugin uploads its covering fragment, the upload fails with `{open_failed, _, enoent}`. The core (correctly, per #210) classifies this as non-transient and keeps the fragment at the head of the in-flight queue, retrying it forever so `next_offset` never advances past a non-durable range. But once the segment is permanently gone, no retry can ever succeed: the manifest wedges at the stalled offset, every fragment uploaded behind it is orphaned, and consumers read only the small prefix below the stall before `become_local` jumps them to the local tail, silently skipping the range in between.

`start_reading0/1` already handles this exact condition at reader init: when `init_data_reader` reports the local log first offset is ahead of the manifest `next_offset`, it discards the remote manifest and restarts from the local first offset, accepting the lost range. The gap was that this recovery only ran when opening the data reader, not when the same trim surfaced mid-stream as an upload failure.

This routes the mid-stream case to the same recovery:

- detect the permanent case in the transfer-result handler: when the live local log first offset (`osiris_log_shared:first_chunk_id`) is ahead of the manifest `next_offset`, the stalled fragment's segment is gone for good, so recover instead of retrying. The contiguity invariant guarantees the head fragment begins exactly at `next_offset`, so comparing the local first offset against `next_offset` is equivalent to comparing it against the stalled fragment's offset, mirroring the check `start_reading0/1` makes at init. A transient `enoent` during a segment roll leaves the local first offset at or below `next_offset` and still retries
- recovery resets the in-flight bookkeeping and delegates to `resolve_and_start/1` (from #226), so a transient store failure during recovery reschedules a retry rather than proceeding with a possibly-empty manifest
- recovery abandons the transfers queued behind the stall. Their late results would otherwise reach a core that no longer tracks their references: a success would append a non-contiguous orphan that trips `assert_contiguous`, and any result would crash `get_meta` on the unknown reference. Drop transfer results for untracked references in `handle_info` before they reach the core

## File-descriptor leak fix

While validating on the cluster, this surfaced a pre-existing leak that the repeated recovery made acute: the reader's restart paths set `log = undefined` without closing the `osiris_log` data reader first, leaking its open segment file descriptors. When retention then unlinks those segments the kernel cannot reclaim the space, and deleted-but-open segments accumulate until the data volume fills and the disk alarm blocks publishers (observed: 11.65 GiB held open after a few hundred recoveries). Added `close_log/1` and call it on every path that drops the log (the `reinitialize` effect, the trimmed-segment recovery, the `missing_file` retry, and `terminate`).

## Durability model

Documents the retention-versus-upload durability model that this recovery follows: the user's local retention bound is a hard limit, tiering never holds segments past it to wait for an upload, and a range trimmed before upload is lost from both tiers by design. Adds the "Durability versus the local retention bound" section to architecture.md and the "Reset safety invariant" to invariants.md, and extracts `reset_manifest/2` so an empty reset manifest carries `first_offset = next_offset = local floor`.

## Validation

Validated on the 3-node cluster with the high-throughput content-verification test (single producer, ~570k msg/s, `max-length-bytes=500000000`): no crashes, no stalls, no FD leak, the disk alarm no longer fires, and the recovery advances `next_offset` forward each time rather than wedging. The recovery fires repeatedly under sustained overload because uploads cannot keep pace with ingest at that rate; that is the intended discard-and-resume behavior. The fragments left in S3 below the reset floor are reclaimed by the existing `below_first_offset` garbage collector; that GC currently runs only on demand, tracked separately in #230.

## Tests

`close_log/1`, the `local_log_ahead/1` boundary (the strict greater-than that distinguishes a permanent trim from a transient segment roll), the stale-transfer-result guard, and `reset_manifest/2` are covered by inline eunit tests. The end-to-end mid-stream stall is a retention-versus-upload race best exercised by the cluster content-verification test rather than a timing-dependent CT case; the shared recovery path it funnels into is already covered by `local_ahead_discards_manifest` in `replica_reader_SUITE`.


the-mikedavis (commented):
This fix direction looks correct to me. The intention is that a period of unavailability in S3 should not make an active stream with a retention like 5 GB, for example, grow in the local tier waiting for upload to S3 before retention. Local and remote retention can act independently and that can lead to this scenario, and the replica reader needs to handle it gracefully.

I'm looking at updating the documentation to make this much more explicit, and maybe a test, if you don't beat me to it!

the-mikedavis (approved):

PR #231, opened 2026-06-18
Improve logging and metrics for edge-cases

the-mikedavis:
This is a small grab-bag of observability improvements - improving logging and metrics for a few cases we will probably want to see. I also took a pass at updating the operations doc to add metrics we've been adding but didn't capture in the docs yet.

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -199,7 +199,8 @@ These metrics decompose disk-resident bytes by their position in the upload pipe
 | Metric                              | Type    | Description                                                       |
 |-------------------------------------|---------|-------------------------------------------------------------------|
 | `rabbitmq_stream_s3_manifests_resolved`                | counter | Times a non-empty manifest was resolved on startup                |
-| `rabbitmq_stream_s3_manifests_resolved_empty`          | counter | Times an empty manifest was resolved on startup                   |
+| `rabbitmq_stream_s3_manifests_resolved_empty`          | counter | Times a genuinely empty manifest was resolved on startup          |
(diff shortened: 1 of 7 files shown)


lukebakken (approved):
Reviewed the full diff, built, and ran the suites locally — 97 eunit, broker_SUITE 6/0, replica_reader_SUITE 26/0, no compiler warnings, no trailing whitespace. Approving.

Walked through each commit:

- **sup + hooks**: replacing the silent `catch _:_ -> ok` swallows with logged warnings is a clear improvement. Confirmed `rabbitmq_stream_s3_db:setup/0` is spec'd to and does return `ok`, so the new `ok = ...setup()` assertion won't spuriously trip the warning path.
- **manifest_replica**: the `resyncs_requested` counter is correctly wired into `rabbitmq_stream_s3_api:init/0`, and `inc/2` safely no-ops before the counter is registered.
- **reaper + aws**: parsing the DeleteObjects 200-body for per-key `<Error>` entries and counting only confirmed deletes is the right fix — a 200 genuinely can hide per-key failures. Good test coverage (all-ok, partial, unexpected-root).
- **docs**: verified all newly-documented metrics exist in code and that the removed `await` / `await_duration_ms` counters are genuinely gone.

One non-blocking observation: the reaper's precise confirmed-vs-failed split only triggers on the AWS backend's `{error, {delete_errors, _}}` shape. The FS backend returns partial failures as `{error, [{Key, Error}]}`, which lands in the reaper's catch-all `{error, Reason}` branch and counts the whole batch as failed. Since the FS backend is test-only, this is harmless over-counting rather than a bug — just noting the asymmetry in case the FS backend is ever used for metrics assertions.

PR #232, opened 2026-06-18
Reclaim orphaned fragments via GC after a manifest reset

### LINKED ISSUE #230: Reset recovery orphans already-uploaded fragments in S3; GC reclaims them but runs only manually
## Summary

Under sustained retention pressure, the trimmed-segment upload-stall recovery (#225) repeatedly resets the remote manifest to the local floor. This is intended behavior (see #229: the local retention bound is a hard limit and a range trimmed before upload is lost from both tiers). The problem this issue tracks is a side effect of that reset: fragments already uploaded to S3 below the reset floor are left behind as orphans, and nothing reclaims them automatically.

The garbage collector (`rabbitmq_stream_s3_gc`) correctly identifies these fragments (offset below the manifest `first_offset`, classified `below_first_offset`) and deletes them in `delete` mode. The gap is that GC runs only on demand via the `stream_s3_gc` CLI command. There is no scheduler or supervised worker that runs it periodically. So under the reset model, orphans accumulate in S3 indefinitely until an operator runs GC by hand.

## How it was observed

A `content-verification` run on the 3-node cluster: single producer, 256-byte messages, `max-length-bytes=500000000`, ~570k msg/s (at or above the S3 upload rate, so user retention trimmed local segments at roughly the rate uploads completed).

The recovery fired 351 times on the writer node over the 5-minute publish phase. At the end:

- Published: offsets 0 to 171160371 (171.16M messages)
- Local tier first offset (all three nodes): 169845578
- Live manifest root: `first_offset=169845578 next_offset=171103351`
- S3 fragments physically present: 187 fragments spanning offsets 1255079 to 170850836, roughly 12 GiB, all below the manifest floor of 169845578 and therefore unreferenced

## Mechanism (verified against the code on the #225 branch, post-#229)

Two deletion paths exist, and the orphans fall between them:

1. The reset path calls `delete_manifest_objects(StreamId, Manifest)`, which deletes only the objects referenced by the in-memory manifest being discarded (`rabbitmq_stream_s3_fragment_iterator:all_refs(Manifest)`). Under sustained recovery the manifest has usually already collapsed to a near-empty root by the time a reset runs, so fragments uploaded and applied between resets were never in a manifest this call enumerates. Their S3 objects are left behind. It is also a best-effort fire-and-forget `spawn` with no retry.

2. The GC path (`rabbitmq_stream_s3_gc`) LISTs `rabbitmq/stream/` and deletes any data fragment whose offset is below the manifest `first_offset`. After the reset, `first_offset = next_offset = local floor`, so all 187 fragments are below the floor and would be reclaimed correctly. The deletion logic is sound.

So this is not a deletion-logic bug. It is a missing automatic reclamation path: the reset model leans on GC to clean up after it, but GC does not run on its own. Confirmed there is no `rabbitmq_stream_s3_gc:run` call anywhere outside the CLI command, and `rabbitmq_stream_s3_sup` has no GC child.

## Not in scope

The manifest reset to the local floor, and the loss of the trimmed range, is intended per #229 and is not a bug. The surgical forward-skip that would have preserved the fragments in the manifest (#227) has been closed as not planned, because it contradicts the #229 durability model and would require the manifest and read path to represent a non-contiguous remote tier.

## Possible directions

- Run GC automatically: a periodic trigger (a supervised worker on an interval, or a hook after N resets) so orphans below the floor are reclaimed without operator action. This is the smallest change and fits the existing GC design.
- Or make the reset self-cleaning: instead of `all_refs` of the (often already-collapsed) in-memory manifest, have the reset delete the stream's data objects below the new floor by listing the prefix. This duplicates GC's classification logic.
- Either way, confirm the two paths do not race or double-delete.

## References

- #229: documents the discard-and-resume durability model that makes the reset intended
- #227: closed as not planned (the surgical forward-skip contradicts #229)
- #225 / PR #228: the recovery this builds on

lukebakken:
Fixes #230.

## Problem

When local retention trims a segment before its fragment is uploaded, the replica reader discards the remote manifest and restarts from the local log first offset (the #225 recovery, on `main` via #228). Fragments already uploaded below the new floor are left in S3 as orphans.

The reset previously called `delete_manifest_objects/2`, which only removed objects referenced by the in-memory manifest being discarded (`all_refs`). Under sustained recovery that manifest has usually already collapsed to a near-empty root by reset time, so fragments uploaded and applied between resets are never enumerated and leak. In the prompting run this left 187 fragments (~12 GiB) in S3.

## Fix

Route reclamation through the garbage collector instead of the manifest-walk, per the direction in #230: GC's epoch and `first_offset` checks are the safe signal, especially on a deposed writer.

- Add `rabbitmq_stream_s3_gc:run_stream/2`: lists only the stream's own S3 prefix and deletes any data object below the manifest `first_offset`, reusing the collector's existing `below_first_offset` / `stale_epoch` classification
- The replica reader spawns it in `delete` mode **after** installing the reset manifest, so the collector observes the new (advanced) floor
- `build_stream_lookup/1` reads `first_offset` from the manifest record directly rather than via `get_range/1`. The reset installs an *empty* manifest, and `get_range/1` reports `empty` whenever the entries array is empty — which would skip the stream and reclaim nothing in exactly the case this targets

## Correctness hardening: `stream_prefix/1`

`stream_prefix/1` now includes the trailing slash. Stream IDs derive from user-controlled vhost and queue names, so one stream ID can be a textual prefix of another; without the delimiter a LIST on the prefix could match a sibling stream's objects. This also hardens the reaper's stream-deletion listing, which already used `stream_prefix/1` without the delimiter.

Stream IDs cannot contain a slash (they are `base64uri`-encoded, and `/` is not in the allowed set), so no double-slash case arises.

## Safety

GC only deletes objects strictly below `first_offset`, which is monotonic (a stale read can only be lower, never higher), so deletion is always conservative. A repeated reset re-runs GC; double-deletes of an already-removed S3 key are no-ops. The lookup degrades to a no-op (`skip`) if Khepri or the manifest replica is unavailable.

## Validation

- 96 eunit tests pass
- `replica_reader_SUITE`: 26 ok, 0 failed — `local_ahead_discards_manifest` is the direct end-to-end test for this path (drives a real discard, the reset installs an empty manifest, and GC reclaims the old fragments through Khepri + the manifest replica)
- `broker_SUITE`: 6 ok, 0 failed — exercises the `stream_prefix/1` change via the reaper's `stream_deletion_cleans_remote_tier` and the listing helpers
- Clean compile, no warnings

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -504,6 +504,13 @@ rabbitmq-streams stream_s3_gc --formatter json
 rabbitmq-streams stream_s3_gc --mode delete
 ```
 
+Pass a stream name to restrict GC to a single stream, which lists only that stream's S3 prefix instead of sweeping every stream:
+
+```bash
+rabbitmq-streams stream_s3_gc my-stream --vhost /
+rabbitmq-streams stream_s3_gc my-stream --mode delete
+```
+
 See [Garbage collection](#garbage-collection) below for the mechanism and safety guarantees.
 
 ### Inspect a replica reader
--- a/src/Elixir.RabbitMQ.CLI.Ctl.Commands.StreamS3GcCommand.erl
+++ b/src/Elixir.RabbitMQ.CLI.Ctl.Commands.StreamS3GcCommand.erl
@@ -29,17 +29,24 @@ help_section() ->
 
 validate([], _Opts) ->
     ok;
+validate([_Stream], _Opts) ->
+    ok;
 validate(_, _Opts) ->
     {validation_failure, too_many_args}.
 
 merge_defaults(Args, Opts) ->
-    {Args, maps:merge(#{mode => <<"dry_run">>}, Opts)}.
+    {Args, maps:merge(#{mode => <<"dry_run">>, vhost => <<"/">>}, Opts)}.
 
 usage() ->
-    <<"stream_s3_gc [--mode <dry_run|delete>]">>.
+    <<"stream_s3_gc [<stream>] [--vhost <vhost>] [--mode <dry_run|delete>]">>.
 
 usage_additional() ->
     [
+        [
+            <<"<stream>">>,
+            <<"Restrict GC to a single stream. When omitted, every stream is swept.">>
+        ],
+        [<<"--vhost <vhost>">>, <<"The virtual host of the stream.">>],
         [
             <<"--mode <mode>">>,
             <<"dry_run (default) to report only, delete to remove dangling objects.">>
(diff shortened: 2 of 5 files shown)


the-mikedavis (changes requested):
One minor thing while we're here: let's make GC possible to run on a specified stream in the CLI command as well (and update the operations doc to show it used that way).

the-mikedavis (approved):

PR #234, opened 2026-06-18
Guard the post-reset GC sweep against a deposed writer

the-mikedavis:
The orphaned-fragment reclamation added after a local-log-ahead reset deletes remote data objects below a floor read from the local just-reset manifest, which sits above the committed floor, and classifies data objects by offset with no epoch gate. A deposed writer on the minority side of a partition keeps running, arms the reset via local retention, and would sweep-delete a successor's live committed fragments in that gap. The lookup read it relied on, `db:get`, is a low-latency local read that succeeds with stale state on a minority rather than failing, so it did not protect against this.

Make the sweep fail closed for a deposed writer:

- add `db:get_consistent/1`, a strongly consistent quorum-requiring read bounded by a timeout, keeping Khepri's read options inside the `db` module; the stream-scoped GC lookup uses it and skips on any error, so a minority writer that cannot reach a quorum does not sweep
- the reset threads the writer's own epoch into the sweep, and the sweep proceeds only when the committed epoch equals it, so a deposed writer that can still reach a quorum also skips. The operator CLI path pins no epoch and relies on the consistent read alone
- the consistent read runs in the spawned GC task, not the reader, so a partition cannot stall the reader

Connects #230

### DIFF
--- a/src/rabbitmq_stream_s3_db.erl
+++ b/src/rabbitmq_stream_s3_db.erl
@@ -28,6 +28,10 @@ writer cannot make progress anymore.
 
 -define(PATH(StreamId), ?RABBITMQ_KHEPRI_ROOT_PATH([rabbitmq_stream_s3, StreamId])).
 -define(STREAM_QUEUE_DELETION_TRIGGER_ID, rabbitmq_stream_s3_db_sq_deletion).
+%% Bounds a consistent read so it cannot block indefinitely waiting for a quorum
+%% that will not form (for example on a minority partition); timing out there
+%% surfaces as an error, which a fail-closed caller treats as "skip".
+-define(CONSISTENT_READ_TIMEOUT_MS, 30_000).
 
 -doc """
 Version number of a manifest object.
@@ -46,7 +50,7 @@ Zero indicates that the manifest has not been created yet.
 
 -export([setup/0]).
 
--export([get/1, list/0, count/0, put/5, queue_path/1]).
+-export([get/1, get_consistent/1, list/0, count/0, put/5, queue_path/1]).
 
 -define(C_SPROC_TRIGGERS, 1).
 -define(C_GETS, 2).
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
Reviewed thoroughly. This correctly closes the split-brain data-loss hole the post-reset GC introduced, and the two-layer defense is sound. I verified the load-bearing pieces locally:

- `favor => consistency` and `timeout` are the correct khepri option names and values, and `rabbit_khepri:adv_get/2` forwards them straight through to `khepri_adv:get/3`. A typo here would have silently defeated the guard, so it was worth confirming.
- The four new `epoch_permits_sweep` tests pass (compiled the branch with the proper include paths and ran them): no-epoch CLI path, matching epoch, deposed writer (committed > writer), and uncommitted writer (committed < writer).
- Keeping the consistent read inside the spawned GC task rather than the reader is the right call so a partition cannot stall the reader, and confining the khepri read options to the `db` module is good encapsulation.

One question, non-blocking, just to confirm intent rather than a change request.

On the reset path the sweep is gated on `CommittedEpoch =:= WriterEpoch`, and the reset itself only does a local `put_manifest`, it does not commit to Khepri. So the equality holds only once this writer has already committed at least once at its epoch. A freshly elected writer at epoch N that hits a local-log-ahead reset on its first read, before ever committing at epoch N, will see the predecessor's committed epoch N-1 and skip the sweep. That is the `epoch_permits_sweep_uncommitted_writer_test` case.

That skip is safe, it errs toward not deleting, so there is no data-loss concern. I just want to confirm it is intended that automatic orphan reclamation does not fire on that specific path (fresh writer, reset before first commit) and waits for a later reset after a commit, with the operator CLI available in the meantime. If so, that is a reasonable conservative trade-off and I have no objection.

Approving.

the-mikedavis:
Yep! That's the idea - we lean towards being pessimistic and not deleting objects (for now, assuming that GC will eventually take care of this later) rather than possibly delete something live. The local put_manifest should eventually end up in Khepri anyways after it's been debounced.

PR #235, opened 2026-06-18
Add a helper to ensure stream IDs are converted to binary correctly

### LINKED ISSUE #188: Validate plugin is Unicode-clean for non-ASCII stream names
## Summary

The plugin converts `osiris:name()` (which is `string() | binary()`) to a binary stream ID using `iolist_to_binary/1` in multiple call sites. This is correct for Latin-1 codepoints (0-255) and for the common case where `Name` is already a binary, but would produce mojibake for stream names containing Unicode codepoints > 255 when `Name` is a list.

## Scope

`rabbitmq_stream_s3_hooks.erl` has 7 instances of `iolist_to_binary(Name)`. The same pattern may exist elsewhere in the plugin.

## What to validate

- Confirm whether RabbitMQ stream names can contain non-Latin-1 Unicode characters in practice (stream protocol, AMQP 1.0, management API, definitions import)
- If they can, replace `iolist_to_binary(Name)` with `unicode:characters_to_binary(Name)` at all call sites
- If stream names are guaranteed to be binaries by the time they reach osiris, document that assumption and consider adding a guard or assertion

## Risk

Low. Stream names in production are almost always ASCII binaries passed from the stream protocol or AMQP layer. This is defensive hardening, not a known failure.

the-mikedavis:
*Issue #, if available:* fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/188.

There's an existing doc investigation describing how stream IDs are safe but nothing about Unicode & graphemes. Ulitmately the server ends up limiting down stream IDs to

```
[A-Za-z0-9_-=]
```

So the Unicode changes in this PR are unreachable. But we really ought to have defense-in-depth against any future changes that expand that character class.

### DIFF
--- a/src/rabbitmq_stream_s3.erl
+++ b/src/rabbitmq_stream_s3.erl
@@ -72,6 +72,7 @@ efficiently using the `rabbitmq_stream_s3_array` module.
 -export([
     uid/0,
     format_uid/1,
+    ensure_stream_id/1,
     offset_filename/2,
     ref_key/2,
     manifest_key/2,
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
I had a small commit!

Issue #233, opened 2026-06-18
No publisher backpressure when sustained ingest outruns S3 upload throughput


lukebakken:
## Summary

When a stream's sustained ingest rate exceeds the sustained S3 upload throughput, the plugin has no backpressure tied to upload *lag*. The un-uploaded backlog grows on the local volume until one of two things happens, and neither is graceful:

1. With a large `max-length-bytes` (much larger than the local volume), the backlog fills the disk and trips the node disk resource alarm, which hard-blocks stream publishers at the connection level.
2. With a small `max-length-bytes` (smaller than the peak backlog), local retention trims un-uploaded segments before they reach S3, triggering the upload-stall recovery reset (#225) and producing a consumer-visible data gap.

The configuration users will reach for ("a 100 GiB stream on a 15 GiB volume, with everything beyond a small local working set living in S3") is *supported and works* as long as average upload keeps pace with average ingest. The failure mode is purely a throughput-mismatch problem, and it is easy to hit under a burst or a slow-S3 period.

## How retention actually works today

There are two independent retention passes. They are applied in fold order by osiris `evaluate_retention0` (`deps/osiris/src/osiris_log.erl`), with the plugin's local retention function prepended ahead of the user's specs (`rabbitmq_stream_s3_hooks:on_retention_updated/2`).

**Pass 1 - local retention function (plugin).** `eval_local_retention/2` in `rabbitmq_stream_s3_hooks` keeps only the segment containing the last-uploaded offset and everything newer, deleting fully-uploaded older segments. In steady state the local footprint is therefore bounded by *upload lag*, not by `max-length-bytes`. This is the intended behavior and is why a 100 GiB policy can live on a 15 GiB volume.

**Pass 2 - user `max_bytes` spec.** This runs in two distinct locations:

- Locally, osiris `eval_max_bytes` runs on the survivors of pass 1.
- Remotely, against the S3 tier, via `rabbitmq_stream_s3_manifest:entries_to_remove/3` -> `remove_for_max_bytes/4`. It walks leading fragments subtracting each fragment's size from the manifest `total_size` until the total is at or below `max_bytes`, advances `first_offset`, and the reaper deletes the trimmed objects from S3 after the manifest persist completes (`rabbitmq_stream_s3_replica_reader:maybe_evaluate_remote_retention/4` -> `on_remote_retention/5` -> `flush_deferred_deletions/1`).

[...]

PR #236, opened 2026-06-18
Raise content-verification max-length-bytes default to 8 GB

lukebakken:
## Summary

`make content-verification` ran with `--max-length-bytes 500000000` (500 MB), a single unthrottled producer, and 256-byte messages. Under that configuration the producer outruns S3 upload, so the un-uploaded backlog grows. Once the backlog crosses 500 MB, local retention (osiris `eval_max_bytes`, applied to the survivors of the plugin's local retention function) trims un-uploaded segments before they reach S3. That discards confirmed-but-not-yet-uploaded data and triggers the upload-stall recovery reset (#225), so the replay phase reports gaps for confirmed sequences that were never delivered.

This is not a disk problem: the volume stayed around 13.5 GiB free throughout the run. The data loss came entirely from the 500 MB bound trimming un-uploaded data, which is exactly the small max_bytes regime described in #233.

## Change

Raise the `CONTENT_VERIFY_MAX_LENGTH_BYTES` default from 500 MB to 8 GB, comfortably above the observed peak backlog (~1.5 GiB at the 500 MB bound) and well below the test volume, so the target verifies content integrity rather than exercising the upload-stall recovery path. A comment explains why the bound must exceed peak upload lag and links #233.

## Caveat

This relies on S3 upload keeping pace with ingest on average (which it did in the failing run; the disk stayed flat). A test that deliberately stresses a small bound under sustained overload will still lose confirmed data until publisher backpressure based on upload lag exists. That is tracked in #233.

## Related

- #233 No publisher backpressure when sustained ingest outruns S3 upload throughput
- #225 upload-stall recovery

### DIFF
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -49,7 +49,11 @@ cleanup: package
 
 CONTENT_VERIFY_DURATION ?= 300
 CONTENT_VERIFY_PRODUCERS ?= 1
-CONTENT_VERIFY_MAX_LENGTH_BYTES ?= 500000000
+# This must exceed the peak un-uploaded backlog: when the publisher outruns the
+# S3 upload, local retention trims un-uploaded segments once the backlog crosses
+# this bound, which discards confirmed-but-not-yet-uploaded data and makes the
+# replay phase report spurious gaps. See issue #233.
+CONTENT_VERIFY_MAX_LENGTH_BYTES ?= 8000000000
 CONTENT_VERIFY_REPLAY_CONSUMERS ?= 4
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Makes sense :+1:

For an end-to-end test like this I think matching the retention close to the max segment size really stresses the seams of the plugin around independent retention. I think theoretically we probably don't want to change the approach of the plugin, but it makes a case like this a tough test to pass. Bumping up to a few GB seems more realistic also :+1:

PR #239, opened 2026-06-18
Fail loudly on an empty index instead of crashing the reader

the-mikedavis:
`find_position/3` passed the index region of a fragment straight to `find_index_position/2`. When that region read back with fewer than one full record, for example a truncated or partially written fragment object, `find_index_position/2` resolved against an empty array and crashed with `badarg` (offset path) or `badmatch` (timestamp path), taking down the reader and triggering a supervised restart.

Guard the index lookup on `byte_size(IndexData) >= ?INDEX_RECORD_B`. A short index now logs a warning and returns `{error, {empty_index, _}}`, which both `resolve_offset_spec/2` and `init_offset_reader/2` already propagate. A regression test anchors that the unguarded primitive crashes on an empty index in both paths.

### DIFF
--- a/src/rabbitmq_stream_s3_log_reader.erl
+++ b/src/rabbitmq_stream_s3_log_reader.erl
@@ -810,6 +810,7 @@ find_position(Spec, #manifest{} = Manifest, StreamId) ->
     IdxStartPos = ?SEGMENT_HEADER_B + Size,
     case index_data(StreamId, FragmentOffset, Uid, IdxStartPos) of
         {ok, IndexData} when byte_size(IndexData) >= ?INDEX_RECORD_B ->
+            warn_on_partial_index(StreamId, FragmentOffset, IndexData),
             {ChunkId, _, FragPos} = find_index_position(IndexData, Spec),
             %% Position within the fragment object (after the 8-byte header).
             Position = ?SEGMENT_HEADER_B + FragPos,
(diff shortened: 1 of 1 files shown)


lukebakken (approved):

lukebakken (commented):
Traced the change end to end: both crash modes are real and exactly as described (offset path `at(0, 20, <<>>)` -> badarg; timestamp path `try_at` -> `undefined` -> `last(20, <<>>)` -> `undefined` -> badmatch on the `?INDEX_RECORD` match), the guard threshold is right (1-19 byte indexes fall through to the new clause), and `{error, {empty_index, _}}` propagates cleanly through `resolve_offset_spec/2` and `init_offset_reader/2`, both of which already have `{error, _}` clauses. Turning a supervised reader crash into a logged warning plus a clean error is a clear improvement.

Two non-blocking observations, neither worth holding the PR for:

- The test anchors the premise (the primitive crashes on `<<>>`) rather than the fix itself (the new `find_position/3` clause returning `{error, {empty_index, _}}`). That is a reasonable trade-off since `find_position/3` needs a manifest plus an S3 fetch and is not unit-testable here.

- The guard catches "fewer than one full record" but not a trailing partial record after at least one full record: a 25-byte index passes the guard, `len` is 1, and the 5 trailing bytes are silently ignored. No crash, and this is pre-existing behaviour, so arguably out of scope, but it is a milder form of the same truncation the PR addresses.

lukebakken (commented):
Looks good but check out what the 🧞 had to say.

the-mikedavis:
For no. 2 I'll add a warning log for mishapen indexes. I don't think we need an assertion but a log would be interesting to see

PR #240, opened 2026-06-18
Slice a 200 answer to a Range GET instead of trusting it

the-mikedavis:
A Range request must be answered with 206 Partial Content. An intermediary or a non-conformant store can ignore the Range header and answer 200 with the full object. `get_range/3` accepted any 200 or 206 body as the requested range, so a 200 fed the caller the whole object starting at byte 0 rather than the bytes it asked for. For an index or chunk read that is an Exactly-once hazard: the reader parses index records from the wrong offset and resolves to the wrong chunk.

Split the 206 (conformant) and 200 (range ignored) cases. A 206 body is returned as-is. A 200 body is sliced down to the requested range with `slice_range/2`, which mirrors what a conformant 206 would have returned, clamping the range end to the object size as S3 does and returning `{error, {range_not_satisfiable, _, _}}` for a range that starts at or beyond the object. The recovery is logged at debug so a store that never honours Range is discoverable.

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -289,10 +289,19 @@ range.
 get_range(Key, Range, Opts) when is_binary(Key) andalso is_map(Opts) ->
     Headers = #{<<"range">> => range_specifier(Range)},
     case request(<<"GET">>, key_to_path(Key), Headers, <<>>, Opts) of
-        %% HTTP Range requests must return 206 if only a partial range is served,
-        %% according to the RFC.
-        {ok, #{status := Status, body := Data}} when Status =:= 200 orelse Status =:= 206 ->
+        {ok, #{status := 206, body := Data}} ->
             {ok, Data};
+        {ok, #{status := 200, body := Data}} ->
+            %% A Range request must be answered with 206 Partial Content. An
+            %% intermediary or a non-conformant store can ignore the Range
+            %% header and answer 200 with the full object. Returning that whole
+            %% body as the requested range would feed the caller bytes from the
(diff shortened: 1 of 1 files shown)


lukebakken (commented):
Traced the change against the code. The fix is correct: `slice_range/2` has one clause per `range_spec()` form (`{Start, undefined}`, `{Start, End}`, negative suffix) matching `range_specifier/1` exactly, with end-clamping and a catch-all that returns `{error, {range_not_satisfiable, _, _}}`. The new error shape needs no caller changes since `find_position/3` already propagates `{error, _}`. The eunit cases cover each form plus the clamp and unsatisfiable boundaries. Good catch on the Exactly-once hazard.

One substantive scope gap worth raising (not necessarily a blocker for this PR):

The same 200-trusting bug exists on the async path, which this PR does not touch. `handle_async/3` treats 200 and 206 identically (both `{continue, State}`, accumulating the full body) at `rabbitmq_stream_s3_api_aws.erl` ~678-684. That path is `get_range_async/3`, and the only caller is `rabbitmq_stream_s3_remote_reader.erl:390`, i.e. the hot fragment data-read path. So after this PR the index read (sync `get_range`) is protected but the fragment data reads (async) still resolve from byte 0 if a store answers 200, which is arguably the higher-traffic instance of exactly the hazard described here.

A 200 answer mid-stream on the async path is also harder to slice cleanly since the body arrives in `gun_data` frames rather than as one binary, so this is genuinely more involved than the sync fix and reasonable to split into a follow-up. Flagging so it is tracked rather than lost.

lukebakken (approved):
Approving. The fix is correct; the async-path gap I noted is a follow-up, not a blocker.

PR #241, opened 2026-06-18
Reject a stale sync that would roll the replica epoch backward

the-mikedavis:
A sync is a full manifest reset tagged with the writer's epoch and sequence. The two `sync` handlers overwrote the cached `{seq, epoch, writer_node}` unconditionally. Casts from different writer nodes can be reordered, so a delayed sync from a deposed lower-epoch writer could arrive after a newer writer's sync and overwrite it, rolling the cache's epoch and sequence backward (an Epoch monotonicity violation) and re-pinning the stream to the old writer node. The cache then served a stale manifest, and subsequent higher-epoch edits mismatched and forced a re-sync, so it self-healed but routed reads off a wrong manifest in the meantime.

Route both handlers through `maybe_apply_sync/6`, which drops a sync that is not at least as new as what is recorded, comparing epoch first and then sequence so a higher epoch always wins regardless of where its sequence restarted. Dropped syncs bump a new `syncs_rejected` counter and log at info. The decision is a pure `is_stale_sync/3` with eunit coverage.

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -243,6 +243,7 @@ Owned by `rabbitmq_stream_s3_manifest_replica`.
 | Metric                              | Type    | Description                                                       |
 |-------------------------------------|---------|-------------------------------------------------------------------|
 | `rabbitmq_stream_s3_resyncs_requested`                | counter | Re-syncs a replica requested after a manifest broadcast gap or epoch mismatch |
+| `rabbitmq_stream_s3_syncs_rejected`                   | counter | Syncs a replica dropped because they were older than the cached epoch or sequence |
 
 ### S3 API (per-node)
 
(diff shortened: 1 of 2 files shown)


lukebakken (commented):
Traced the change end to end and it is correct, including the field-order detail that is easy to get wrong here.

The recorded tuple is `{Seq, Epoch, WriterNode}` (sequence first), and both `maybe_apply_sync/6` and `is_stale_sync/3` destructure it as `{Seq0, Epoch0, _}`, so the bindings line up. `is_stale_sync/3` then compares `{Epoch, Seq} < {Epoch0, Seq0}`, reordering to put epoch first so a higher epoch always wins regardless of where its sequence restarted. Walked the trickiest eunit case to confirm: `is_stale_sync(5, 10, {1, 6, Node})` is recorded epoch 6 vs incoming epoch 5, so `{5,10} < {6,1}` is true (stale), correctly dropping a deposed lower-epoch writer's delayed sync even though its sequence (10) is higher than the new writer's (1).

Both handlers (call and cast) now route through the same helper, the `undefined` (nothing recorded yet) case is handled, idempotent duplicates are not treated as stale, and the dropped-sync path increments `syncs_rejected` and logs at info with the domain matching the existing `request_resync`. The docs table is updated to match. The pure decision function with eunit coverage is the right shape for this.

No issues. Looks good to me.

lukebakken (approved):
Approving. Verified the epoch/sequence ordering and the stale-sync rejection logic; clean.

PR #250, opened 2026-06-19
Protect the referenced leading group from offset-based GC

the-mikedavis:
GC deleted any data object whose key offset was below the manifest's first_offset, and treated group objects identically to fragments. That is unsound for groups: a group spans a multi-fragment range, and remote retention can partially expire a group, advancing first_offset *into* the group's range while the group object remains referenced (its surviving fragments are still served through it). Its key offset is then below first_offset, so GC deleted a live group index object, orphaning the surviving fragments it pointed to and making those committed offsets unreadable. Reachable on an operator `gc --mode delete` over a live stream whose retention cutoff fell inside a group. (The manifest-reset path is unaffected: it sweeps an empty manifest, so nothing is referenced and every old group below the floor is a true orphan.)

The correct signal is reachability from the current manifest, not offset. Only the first root entry can be a group that is both below first_offset and still referenced (every later entry begins at or above the leading group's coverage end). So GC now records the leading group's object key and protects it:

- leading entry is a level-1 group: protect that one object (its children are fragments, reclaimed correctly by the offset heuristic);
- leading entry is a fragment, or the manifest is empty: nothing referenced, delete below-floor groups as before;
- leading entry is a kilo-/mega-group: its nested leading child group can also straddle the floor as a separate object that cannot be identified without descending the tree, so conservatively skip all group deletion for that stream (rare: a kilo-group needs ~1024 groups). Fragments are still reclaimed.

This is safe under rebalance: a rebalance creates the new group at the front (offset = first_offset), never strictly below it, so GC never considers it; and GC classifies against its own manifest snapshot's first_offset, which is <= the live floor by monotonicity, so a group that becomes the new leading straddler later had offset >= that snapshot floor and is not touched.

parse_key now distinguishes {group, _, _} from fragment {data, _, _}; classify gates group keys on the referenced-leading-group key and the skip-groups flag, threaded through the GC lookup. Inline eunit covers the classify branches and leading_group_info.

### DIFF
--- a/src/rabbitmq_stream_s3_gc.erl
+++ b/src/rabbitmq_stream_s3_gc.erl
@@ -107,16 +107,67 @@ make_handler(delete) ->
 build_lookup(Streams) ->
     maps:fold(
         fun(StreamId, #{epoch := Epoch}, Acc) ->
-            case rabbitmq_stream_s3_manifest_replica:get_range(StreamId) of
-                {FirstOffset, _NextOffset} ->
-                    Acc#{StreamId => #{epoch => Epoch, first_offset => FirstOffset}};
-                empty ->
+            case rabbitmq_stream_s3_manifest_replica:get_manifest(StreamId) of
+                #manifest{entries = <<>>} ->
+                    %% Empty manifest: nothing in the remote tier to compare
+                    %% against (matches the prior get_range/1 `empty` skip).
+                    Acc;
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I think we can be smarter about finding the boundary at which we can delete each kind of group. This patch just covers the possible data loss gap.

Issue #222, opened 2026-06-17
Replica reader registry leaks stale entries on untrappable reader death


lukebakken:
## Summary

The replica reader registry (`rabbitmq_stream_s3_registry`) is a plain ETS table with no process monitoring, and its only deleter is the reader's `terminate/2`. If a reader dies without `terminate/2` running (an untrappable kill), the stale `{StreamId, Pid}` entry survives, and because registration uses `ets:insert_new`, a supervised restart of that reader fails to re-register and exits during init. The stream then cannot recover until something clears the entry.

This is pre-existing and not caused by the two-layer supervision change (#218); that work just made the per-stream restart path more prominent, which is how it surfaced.

## Mechanism

`rabbitmq_stream_s3_registry`:
- `register_name/2` uses `ets:insert_new/2`, returning `no` if the key already exists.
- `unregister_name/1` (`ets:delete/2`) is the only path that removes an entry.
- There is no `monitor` on the registered pids, so the registry never learns that a holder died.

`rabbitmq_stream_s3_replica_reader`:
- registers via the `{via, registry, {StreamId, node()}}` name during init.
- calls `unregister_name/1` only in `terminate/2`.
- traps exits, so `terminate/2` runs on a normal stop or a trappable supervisor shutdown (the configured shutdown is a 5000ms timeout, not `brutal_kill`).

So on the normal paths the entry is cleaned up correctly. The gap is the untrappable death: `exit(Pid, kill)`, an OOM kill, or any `brutal_kill`. `terminate/2` does not run, the ETS row is left pointing at a dead pid, and the per-stream supervisor's restart attempt calls `register_name`, gets `no`, and the new reader exits during init (the same duplicate-name path that `start_child` already tolerates). The stream stays parked with a stale registry entry until a `discover/0` or some other path clears it.

## Impact

Low likelihood, but a real availability gap: a single hard-killed reader can wedge its own stream out of tiering until the plugin is re-enabled or discovery runs. `whereis_name` would also return a dead pid in the window, though callers generally send to it via the registry and would get a delivery to a dead process rather than a crash.

## Possible fixes

- Have the registry `monitor` each registered pid and `ets:delete` on the `DOWN`, so the entry is self-healing regardless of how the holder died. This is the standard pattern for a via-registry and removes the dependency on `terminate/2` running.
[...]

PR #264, opened 2026-06-20
Periodically reconcile orphaned writers and replicas

the-mikedavis:
The lifecycle of replica and replica reader processes is not strongly coupled to the osiris_writer and osiris_replica processes. Instead of trying to hande race conditions directly (which I'm not sure we actually _can_ do), we can periodically check that the associated process in this plugin is running. If it is not yet, we can start and attach it to the writer/replica process.


the-mikedavis:
We could probably reuse this process for https://github.com/amazon-mq/rabbitmq-stream-s3/issues/222

PR #271, opened 2026-06-21
replica reader: Extract a pure state machine for async task I/O

the-mikedavis:
We've been identifying and squashing a lot of bugs, and most of them live in the replica reader shell half of that functional-core/imperative-shell split. This change is a large refactor that moves the async task I/O, which the shell did, into another pure state machine. This makes property based testing possible for the complicated async interleavings.

This state machine is much more formal than the functionality we had directly in the gen_server previously. There are also typing changes to move towards sum-types which make some classes of bugs structurally impossible to represent.

### DIFF
--- a/src/rabbitmq_stream_s3_replica_reader.erl
+++ b/src/rabbitmq_stream_s3_replica_reader.erl
@@ -632,95 +632,16 @@ handle_info({group_upload_result, Gen, Result}, #state{tasks = Tasks0} = State0)
         {group_result, Gen, Result}, Tasks0
     ),
     apply_group_decisions(Decisions, State0#state{tasks = Tasks});
-handle_info(
-    {retention_result, Token, unchanged},
-    #state{retention_token = Token, core = Core0, retention_mon = Mon, retention_timer = TRef} =
-        State0
-) ->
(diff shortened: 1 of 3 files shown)


the-mikedavis:
I did some manual mutation testing and the property test catches all instances of a class of bugs which we've been chasing:

- Timeout-race staleness: an async result that arrives after its own timeout already freed the slot is wrongly delivered to the core
- Cross-recovery mis-attribution: a result from a task spawned before a recovery is wrongly delivered to the new task now occupying the slot
- Gauge drift / lost persist bytes: a failed persist drops its snapshotted bytes instead of returning them to the pending pool, so the bytes-in-persist gauge under-counts
- Phantom upload after reset: a transfer retry resubmits an upload for a Ref that is no longer outstanding (e.g. cleared by a recovery)
- Stale deadline against a re-armed transfer: a liveness deadline fires against a transfer whose deadline was since re-armed with a fresh token, failing the live transfer


Naturally this change doesn't guarantee that the replica reader is bug-free, but it shrinks the hard-to-test parts to the wiring of the task model (e.g. specifying a wrong generation), which is now much much smaller and easy to reason about.

PR #274, opened 2026-06-22
Add formal models in P

the-mikedavis:
This is a handful of formal models based on the components of the plugin and, especially, recently fixed bugs. These are written in [P](https://github.com/p-org/P) rather than TLA+. P is a natural choice for Erlang since it models state machines communicating via message passing.

### DIFF
--- a/.github/workflows/build-rabbitmq-server.yaml
+++ b/.github/workflows/build-rabbitmq-server.yaml
@@ -19,7 +19,7 @@ jobs:
             otp-version: 27
             elixir-version: 1.18
     steps:
-      - uses: erlef/setup-beam@v1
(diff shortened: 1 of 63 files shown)


the-mikedavis:
These models didn't find any new bugs, but they represent the recently fixed bugs:

- gc-reset: c68ea8d (re-validate GC deletions against live first offset), b97225f (reclaim orphans after reset)
- writer-fencing: 93aaea8 (guard post-reset GC sweep against a deposed writer)
- gc-leading-group: c033eff (protect the referenced leading group from offset-based GC)
- tier-routing: 8b1e111 (resolve against the remote tier when the local log is empty)
- read-resolution: 82f7364 (surface group-fetch failures), e3f931b (retry, not local tier)
- orphan-leak: b97225f (reclaim orphaned fragments via GC after a manifest reset)

PR #277, opened 2026-06-22
GC: Tie first offset to epoch

the-mikedavis:
GC reads the deletion floor (first_offset) from the local node's manifest cache, but the committed epoch from a quorum read — with nothing tying them together. On a node lagging a floor-lowering reset, an operator stream_s3_gc --mode delete reads a stale-high floor; still_dangling/1 re-reads that same stale cache, so the safety guard defends nothing and a live re-tiered object gets deleted. Fix: record the writer epoch on every cache write and only sweep when the cached epoch matches the committed epoch (fail closed otherwise).


lukebakken (approved):
🤯

PR #280, opened 2026-06-22
GC: Fix durability bug

the-mikedavis:
Fixes a durability bug where the orphan GC sweep can delete a live, referenced leading group during a remote-tier-ahead reset. No UID collision is required: it's an ordinary GC × reset × retention race.

rabbitmq_stream_s3_gc protects the referenced leading group with two independent guards:

- the classify-time carve-out (classify_group/3 + referenced_group_key), captured once at the sweep snapshot (lookup_entry) and passed by value through list_and_classify
- the execute-time re-check (still_dangling/1), which re-validates an offset finding against the live floor just before deletion

For a group finding, still_dangling/1 re-read only the live floor (Offset < first_offset) — it never re-ran leading_group_info/2. So this sequence within a single sweep deletes live data:

1. GC snapshots at floor 1000, records snapshot leading group (900,1)
2. A remote-tier-ahead reset lowers the floor to 850 and installs a fresh leading group (850,2)
3. Forward retention advances the floor to 870 — (850,2) now straddles the floor and is the live referenced leading group
4. classify (vs the stale snapshot) sees (850,2) below floor 1000 and ≠ snapshot referenced_group_key (900,1) → candidate
5. still_dangling re-reads live floor 870; 850 < 870 → deletes the live leading group (INV#2 dangling reference)

A consumer reading the surviving offsets in (850,2) can no longer fetch it.

The fix: still_dangling/1 now re-derives the carve-out from the live manifest (live_leading_group/3, mirroring classify_group/3) and keeps a group that is the live referenced leading group, or that the live manifest covers in conservative skip_groups mode. It reuses the manifest already fetched for the floor read — no extra round-trip.

### DIFF
--- a/.github/workflows/p-model-check.yaml
+++ b/.github/workflows/p-model-check.yaml
@@ -20,6 +20,7 @@ jobs:
     outputs:
       gc_reset: ${{ steps.filter.outputs.gc_reset }}
       gc_leading_group: ${{ steps.filter.outputs.gc_leading_group }}
+      gc_reset_leading_group: ${{ steps.filter.outputs.gc_reset_leading_group }}
       read_resolution: ${{ steps.filter.outputs.read_resolution }}
       trimmed_segment: ${{ steps.filter.outputs.trimmed_segment }}
       orphan_leak: ${{ steps.filter.outputs.orphan_leak }}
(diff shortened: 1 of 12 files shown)


lukebakken (approved):
🤯

Issue #275, opened 2026-06-22
S3 read path churns connections: non-blocking checkout grows the pool on every momentary miss


lukebakken:
## Summary

The asynchronous S3 read path churns connections in the general pool: a workload whose real concurrency never exceeds 3 connections drove 333 connection creations and 244 `pool_busy` events on a single node during one replay. The pool oscillates between roughly 1 and 14 connections while only 1-2 are ever in use. This is wasted work, not a correctness problem (the run it was observed in passed), but it adds latency and load and would compound under heavier read concurrency.

## Mechanism

The async read path (`get_range_async` -> `request_async` -> `start_async_request` in `rabbitmq_stream_s3_api_aws`) checks out a connection with the **non-blocking** `try_checkout`:

```erlang
start_async_request(Pool, Method, Path, Headers, Body, Opts) ->
    case rabbitmq_stream_s3_api_aws_pool:try_checkout(Pool) of
        busy ->
            {error, pool_busy};
        Conn ->
            ...
```

`try_checkout` (`rabbitmq_stream_s3_api_aws_pool`) returns `busy` the instant there is no already-idle connection, and kicks off an asynchronous `grow(1)`:

```erlang
empty when map_size(Monitors) < MaxSize ->
    ... {reply, busy, grow(1, State0)};
```

The grown connection is not ready until its TLS handshake to S3 completes (tens to hundreds of ms). So the caller (`rabbitmq_stream_s3_remote_reader`) gets `pool_busy`, schedules a retry (25ms, 50ms, 100ms ... backoff), and on each retry the connection is still not up, so it hits `busy` again and grows the pool **again**. A single read can stack many grows before any connection becomes available.

Meanwhile the pool's `IDLE_TIMEOUT_MS` is 10s, so connections grown during a burst that are never used are closed shortly after, only to be re-grown on the next burst. The result is a self-sustaining grow/idle-close cycle.

## Evidence

Single node (a replica serving consumer readers), one content-verification replay (about 9.5 minutes, 4 replay consumers reading the full stream from `first`):

- `gun_up` (connection created): **333**
- `pool_busy` events: **244**
- pool `total` size over time: oscillates roughly **1 to 14**
- `checked_out` distribution: `1` x 301, `2` x 181, `0` x 77, `3` x 18 -> **real concurrency is at most 3, almost always 1**
- general pool config: `min_size=2`, `max_size=50`, `IDLE_TIMEOUT_MS=10000`

[...]

PR #283, opened 2026-06-23
Fix S3 read stalls from dead pooled connections

### LINKED ISSUE #279: Remote-tier reads stall for 15s: pool reuses a connection S3 closed while idle (retry=0)
## Summary

During replay (consumers reading the S3 tier from `first`), individual remote-tier reads stall for the full request timeout (15s), then succeed on retry. The broker logs these as `S3 request timed out` and `Slow remote tier read: ~16000ms`.

Root cause: the connection pool hands out a connection that S3 has already closed while it was idle, and the read request is fired onto that dead (or dying) connection. The request never reaches the wire, so it gets no response until the caller's 15s timeout. The retry lands on a fresh connection and succeeds.

The reads recover on retry, so there is no data loss (the content-verification test passes). The cost is latency and throughput: replay throughput decays over a run as a growing share of reads burn a full 15s timeout plus retry (observed ~196k msg/s early, drifting toward ~115k msg/s).

## Root cause

The S3 connection pool (`rabbitmq_stream_s3_api_aws_pool`) opens connections with `retry => 0`:

```erlang
%% rabbitmq_stream_s3_api_aws_pool:open/0
Opts = #{
    transport => tls,
    protocols => [http],
    ...
    retry => 0
}
```

The failure chain:

1. A connection sits idle in the pool. S3 closes idle keep-alive connections after roughly 10s.
2. gun receives the close. With `retry => 0` it does not reconnect: it either stops the connection process (`gun.erl` `not_connected(_, {retries, 0, normal}, State) -> {stop, normal, State}`) or sits in the `not_connected` state with no socket.
3. The pool's idle timeout (`?IDLE_TIMEOUT_MS = 10_000`) is meant to close idle connections first, but it races S3's close and the pool can still hand out a connection whose process has just died or is terminating. The pool's `gun_down`/`'DOWN'` eviction has not necessarily run before the `try_checkout` that hands the connection out.
4. The reader fires the request with `gun:request/5` on that connection. If gun is still in `not_connected`, the request is postponed (`gun.erl`: "We postpone all HTTP/Websocket operations until we are connected") and, with `retry => 0`, never sent because the connection never reconnects. If the gun process has already stopped, the request goes to a dead process. Either way no bytes reach S3.
5. The reader's request timeout (`request_timeout_ms`, default 15s) fires. The connection is closed and the read is retried on a fresh connection, which succeeds.

## Evidence

Observed on every content-verification replay (no-retention configuration, 4 consumers, ~10 min replay). All evidence is from a single node (a replica serving consumer readers); stalls cluster on whichever replicas the consumers attach to.

**The stalled request never reaches S3.** CloudTrail S3 data events for one traced stall: the reader fired a GET for a fragment at ~T, timed out at T+15s. CloudTrail shows other consumers fetching the identical object successfully within the same second, but the stalled request itself has no record in S3 at all, until a burst of retries appears right after the timeout. Every recorded GetObject in the run is `errorCode=OK` (no throttling, no 5xx).

**The request phase is always `no_response`.** With request-phase instrumentation, every timeout reports `phase=no_response (no headers received)`, `headers_after=n/a`, `bytes=0`, `elapsed=15001ms`. No stall ever received a single response byte.

**The connection is dead or dying at request-fire time.** Capturing the gun connection's state at the moment the request is fired, across 32 stalls in one run, every stall fell into one of two states and none was healthy:

- gun process already exited (`conn_alive=false`, `gun:info` raises `noproc`)
- gun process alive but terminating with reason `normal` (`conn_alive=true`, `gun:info` races the shutdown and gets `exit:normal`), which is the `retry => 0` stop path

Zero stalls showed a connected gun process with a live socket.

**S3 closes idle connections at ~10s.** Pool connection-death logging shows connections going down with `reason=closed killed_streams=0 checked_out=false`, at ages clustered around the 10s idle timeout (observed min ~6.7s, median ~9.4s, max ~13.9s). No connection was observed dying with an in-flight request, consistent with the stall being a request fired onto an already-dead idle connection rather than a connection dying mid-request.

## Impact

- Each stall adds ~16s (15s timeout plus ~1s retry backoff) to a read that otherwise completes in milliseconds.
- Replay throughput decays over a run as more reads hit the timeout path.
- No correctness impact: retries succeed and the content-verification test passes.

## Fix directions (not yet decided)

The defect is that a connection closed by S3 can be checked out and used before the pool evicts it. Candidate fixes:

- Evict a connection from the pool's available set the moment its `gun_down`/`'DOWN'` arrives, and ensure `try_checkout` cannot return a connection whose process is not alive (check `is_process_alive` or gun state on checkout).
- Reconnect rather than postpone: set a small `retry` (and `retry_timeout`) so gun re-establishes a closed connection instead of stranding the request, combined with a much shorter "no response started" timeout so a stranded request fails fast rather than waiting the full 15s.
- Close idle connections before S3 does: set the pool idle timeout below S3's idle-close window (and/or proactively recycle connections), so the pool never holds a connection into the window where S3 closes it.

These are starting points; the fix is a design choice about connection liveness and timeout handling on the read path.

## How this was diagnosed

Diagnostics on branch `diag/s3-read-latency-instrumentation`: request-phase markers on read timeouts (`phase=`, `headers_after=`, `bytes=`), connection-death logging in the pool (`reason=`, `killed_streams=`, `checked_out=`, `age=`), and the gun connection state at request-fire time (`gun_state=`/`conn_alive=`/`sock=`). S3 Server Access Logging and a CloudTrail S3 data-event trail were enabled on the bucket to confirm S3's view; see `test/integration/README.md` for enabling and disabling them.

## Related

- #275 S3 read path connection churn (distinct: that is pool grow/close thrash; this is requests lost on individual closed connections)

lukebakken:
## Summary

Fix S3 read stalls caused by the connection pool handing out connections S3 has already closed while idle. Pre-fix, 31-34 reads per replay stalled for the full 15s request timeout before succeeding on retry; post-fix, zero stalls across two validation runs (10-min and 30-min) under sustained read pressure with dozens of live eviction events confirming the race is handled.

Closes #279

## Changes

**Pool (`rabbitmq_stream_s3_api_aws_pool`)**

- Add `usable/1`: verify a connection's gun process is alive and in the `connected` state before handing it out on checkout
- Add `remove_available/2`: evict a connection from the available set and cancel its idle timer (leaving `monitors` intact for the `DOWN` handler)
- `take_available/2`: skip and evict unusable connections recursively
- `checkout/1` (pending drainer): skip unusable connections without consuming the pending entry
- `gun_down` handler: synchronously evict the connection from `available` the moment the message arrives, closing the window before the `DOWN` fires; log abnormal closes (non-normal reason or killed streams) with connection age and checked-out status
- `grow/2`, `close_connection/2`, `'DOWN'` handler: maintain the new `created` timestamp map for connection-age diagnostics
- Document the `retry => 0` design decision, its #279 consequence, and the guards that protect against it

**Request path (`rabbitmq_stream_s3_api_aws`)**

- Record request-phase markers (`req_started_at`, `headers_at`, `first_data_at`, `bytes_received`) so a timeout reports which phase stalled (no-response vs. headers-only vs. mid-body) and elapsed time
- `mark_first_data/2` runs per body frame; documented as the one hot-path diagnostic cost, kept for high-value triage on any future stall

**Documentation (`test/integration/README.md`)**

- S3-side request logging playbook: how to enable/disable S3 Server Access Logging and CloudTrail S3 data events for diagnosing read latency

## Validation

Two content-verification runs on the `fix/s3-connection-liveness` branch against the test cluster (3 nodes, 4 consumers, full stream replay from `first`):

| Run | Duration | Evictions | `S3 request timed out` | Integrity |
|-----|----------|-----------|------------------------|-----------|
| 1 (short) | ~10 min | 9 | **0** | passed |
| 2 (long) | ~30 min | 29 | **0** | passed |

Pre-fix baseline: ~31-34 timeouts per 10-min run, with replay throughput decaying from ~196k to ~115k msg/s. Post-fix: zero timeouts, throughput flat at ~430k msg/s throughout.

## Related

- #275 (connection churn) shares the same lifecycle area; the two should be addressed together in future work

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws.erl
+++ b/src/rabbitmq_stream_s3_api_aws.erl
@@ -865,6 +865,13 @@ now_ms() ->
     erlang:monotonic_time(millisecond).
 
 %% Record the time and running byte count of the first body frame.
+%%
+%% DIAGNOSTIC (#279): this runs on every body frame of every read, so it is the
+%% one diagnostic with a per-frame hot-path cost (a couple of map operations).
+%% It enriches the timeout summary in `describe_stall/1`, which post-fix should
+%% fire ~never. Kept for now because that enrichment is high-value triage if a
+%% stall ever recurs; revisit dropping it from the hot path if the per-frame
+%% cost ever shows up under profiling.
 mark_first_data(#{first_data_at := undefined, bytes_received := Rx} = State, N) ->
     State#{first_data_at := now_ms(), bytes_received := Rx + N};
 mark_first_data(#{bytes_received := Rx} = State, N) ->
@@ -888,14 +895,10 @@ describe_stall(#{req_started_at := Start} = State) when is_integer(Start) ->
         end,
     HeadersMs = elapsed_or_undef(Start, HeadersAt),
     FirstDataMs = elapsed_or_undef(Start, FirstDataAt),
-    %% The socket's local address and port (captured at request-fire time) join
-    %% this stalled request to its TCP stream in a packet capture, so the
-    %% on-the-wire behaviour can be inspected directly.
-    SockInfo = maps:get(sock, State, "sock=n/a"),
     lists:flatten(
         io_lib:format(
-            "phase=~ts elapsed=~bms headers_after=~ts first_data_after=~ts bytes=~b ~ts",
-            [Phase, Now - Start, HeadersMs, FirstDataMs, Rx, SockInfo]
+            "phase=~ts elapsed=~bms headers_after=~ts first_data_after=~ts bytes=~b",
+            [Phase, Now - Start, HeadersMs, FirstDataMs, Rx]
         )
     );
 describe_stall(_State) ->
@@ -903,36 +906,6 @@ describe_stall(_State) ->
 
 elapsed_or_undef(_Start, undefined) -> "n/a";
 elapsed_or_undef(Start, At) -> integer_to_list(At - Start) ++ "ms".
-
-%% Local socket IP and port for a connection, formatted to locate the TCP
-%% stream in a packet capture (the local port is unique per connection on the
-%% host). Queried only on the slow timeout path; tolerates a dead or unknown
-%% connection.
-sock_info(undefined) ->
-    "sock=n/a conn=undefined";
-sock_info(Conn) ->
-    Alive = is_process_alive(Conn),
-    try gun:info(Conn) of
-        #{state_name := StateName} = Info ->
-            %% state_name=connected with a sock_port means a live socket. Any
-            %% other state, or a missing sock_port, means gun has no usable
-            %% socket for this connection right now (e.g. the pooled connection
-            %% was closed by S3 while idle and is being fired onto regardless).
-            Sock =
-                case Info of
-                    #{sock_ip := SockIp, sock_port := SockPort} ->
-                        lists:flatten(io_lib:format("~ts:~b", [inet:ntoa(SockIp), SockPort]));
-                    _ ->
-                        "none"
-                end,
-            lists:flatten(
-                io_lib:format("sock=~ts gun_state=~p conn_alive=~p", [Sock, StateName, Alive])
-            );
-        _ ->
-            lists:flatten(io_lib:format("sock=none conn_alive=~p", [Alive]))
-    catch
-        C:E -> lists:flatten(io_lib:format("sock=err(~p:~p) conn_alive=~p", [C, E, Alive]))
-    end.
 
 -spec cancel_async(async_req(), async_state()) -> ok.
 cancel_async(StreamRef, #{conn := Conn, stream_ref := StreamRef} = State) ->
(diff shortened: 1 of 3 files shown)


src/rabbitmq_stream_s3_api_aws.erl
```diff
@@ -839,7 +842,13 @@ handle_async(
     StreamRef,
     #{conn := Conn, stream_ref := StreamRef} = State
 ) ->
-    ?LOG_WARNING("S3 request timed out on ~tw/~tw", [Conn, StreamRef]),
+    %% DIAGNOSTIC (#275 follow-up, S3 read stall investigation): report which
```
the-mikedavis: Overall I wonder if we can clean up these DIAGNOSTIC comments? Looks like they're adapted from your increased logging for debugging in this issue

the-mikedavis (commented):
This is great! I just have a minor question about comment cleanliness. I think the log messages seem alright - they seem useful and not so verbose that we need to worry

lukebakken:
Good call - cleaned up in d81fb38. Stripped the `DIAGNOSTIC (#NNN)` markers and rewrote each as a plain comment explaining what the code does without referencing the investigation origin.

src/rabbitmq_stream_s3_api_aws.erl
lukebakken: Done! Cleaned up.

PR #287, opened 2026-06-24
Show oldest timestamp across both tiers in counter

### LINKED ISSUE #278: First-timestamp counter shows local tier only
https://github.com/rabbitmq/rabbitmq-server/pull/15412 shows the first timestamp in the UI. I saw in some manual testing that the "oldest message" on a stream even with no retention set keeps moving forward. Let's check that we have our local-retention function not updating that value. It should be the min between local and remote retention.

the-mikedavis:
*Issue #, if available:* fixes #278

*Description of changes:* We fix the first offset so that the total number of offsets in the stream includes both tiers. We need to do the same for the first timestamp too.


lukebakken:
Beat me to it!

PR #289, opened 2026-06-24
Add troubleshooting doc for host-level constraints

lukebakken:
## What

Adds `docs/troubleshooting.md`, a guide for symptoms that look like plugin bugs but originate in the host (network limits, instance sizing). The first entry documents upload-throughput cliffs caused by EC2 network burst-credit exhaustion on the writer node.

## Why

A high-throughput test held ~200 MiB/s S3 upload for ~10 minutes, then cliffed to a flat ~30 MiB/s floor with publishing backpressured and leader memory climbing. It looked like an upload-path regression, but the plugin was healthy throughout (governor `unlimited`, idle pool capacity, unchanged transfer concurrency, zero 503s). The real cause was the leader's burstable NIC (`m7g.large`) draining its egress credit bucket: `bw_out_allowance_exceeded` was climbing while `pps_` and `conntrack_` allowances stayed at zero, and total NIC egress dropped to exactly the instance's sustained baseline. All S3 uploads happen on the leader, whose egress is shared with replication, so the NIC ceiling is hit well before any single consumer of that bandwidth looks saturated.

The doc captures the trigger, the cascade into backpressure and remote-tier reads, the plugin metrics that rule the plugin out, the host signals that confirm the ceiling, and the resolution options (size for sustained egress, pace with `max_transfer_bytes_per_sec`, or cap producers when benchmarking).

## Changes

- New `docs/troubleshooting.md`, scoped to host-level constraints (distinct from `failure-modes.md`, which covers the plugin's own self-healing scenarios)
- Link it from the operator track in the docs index (`README.md`) and from the "Slow replica reader (upload lag)" failure mode
- Remove em-dashes and en-dashes from `invariants.md` and `failure-modes.md` for ASCII-punctuation consistency


the-mikedavis (approved):
Looks great! TIL about the allowance_exceeded thing

Issue #220, opened 2026-06-17
Accuracy pass over all Markdown docs against current code


lukebakken:
## Summary

Do a thorough accuracy pass over every Markdown doc in the repository, checking each claim against the current code. Several subsystems changed recently (two-layer replica-reader supervision, recursive multi-tier retention, fatal-upload retry, the pool_busy backoff and cold-pool read changes), and the prose docs were written before some of those landed, so they have almost certainly drifted.

A concrete drift example already found: `docs/architecture.md` claimed a parked replica reader is re-attached by "discovery or membership reconciliation", but membership reconciliation does not re-create readers; re-attachment happens only via the writer `on_init` hook or `discover/0` at plugin start. (Being fixed in the supervision follow-up branch.) The rest of the docs need the same scrutiny.

## Scope

Review every `.md` for technical accuracy against the code, not just prose. For each doc, verify that:
- Module, function, and macro names referenced still exist and behave as described
- Supervision tree, process structure, and lifecycle descriptions match the current tree (note the new `rabbitmq_stream_s3_stream_sup` middle layer)
- Retention behavior matches the recursive multi-tier implementation (`docs/manifest.md`, `docs/operations.md`, any retention prose)
- Upload-path docs reflect the fatal-upload retry and governor oversized-transfer behavior
- Read-path docs reflect the remote-reader timeouts, pool_busy backoff, and become_local handling
- Config keys, defaults, CLI commands, and metrics names match the code
- Diagrams and byte-layout tables match the macros in `include/rabbitmq_stream_s3.hrl`

## Files

```
AGENTS.md
README.md
CONTRIBUTING.md
CODE_OF_CONDUCT.md
docs/README.md
docs/DEVELOPMENT.md
docs/GLOSSARY.md
docs/architecture.md
docs/concepts.md
docs/conventions.md
docs/enabling.md
docs/failure-modes.md
docs/invariants.md
docs/log-framing.md
docs/manifest.md
docs/operations.md
docs/read-path.md
docs/scale.md
docs/upload-path.md
docs/user-guide.md
docs/investigations/*.md
test/integration/README.md
test/integration/TODO.md
tla/README.md
.github/ISSUE_TEMPLATE/*.md
```

Note: `docs/investigations/*` are point-in-time write-ups; verify their conclusions are still accurate or mark them as historical rather than rewriting them as current state.

## Out of scope

Prose style and wording polish, unless it is actually wrong. This is an accuracy pass.

PR #290, opened 2026-06-24
Audit configuration for consistency (#223)

### LINKED ISSUE #223: Audit consistency of config
Similar to https://github.com/amazon-mq/rabbitmq-stream-s3/issues/220 or https://github.com/amazon-mq/rabbitmq-stream-s3/issues/214, we should take another pass over configuration. Configuration should be accessed through the config module (https://github.com/amazon-mq/rabbitmq-stream-s3/pull/128), tested, and documented. Cuttlefish schema keys are not crucial for every option, but we should weigh adding to the schema and associated SUITE.

lukebakken:
Closes #223.

## What

Audit pass over configuration so every option is accessed through `rabbitmq_stream_s3_config`, tested, and documented, weighing cuttlefish schema keys per option. Companion to the #220 (docs accuracy) and #214 (log levels) audits.

## Route options through the config module

Three options were read directly via `application:get_env` at their call sites rather than through the config module: `fragment_target_size`, `persist_threshold`, and `persist_interval_ms`. They now go through `rabbitmq_stream_s3_config`, keeping the `maps:get(Key, Opts, <accessor>)` test-injection pattern at the call sites. After this change there are zero `application:get_env(rabbitmq_stream_s3, ...)` calls outside the config module.

`operations.md` already documented `stream_s3.persist_threshold` and `stream_s3.persist_interval_ms`, but the schema had no mappings for them, so those documented keys did nothing. This adds the mappings, so the documented config now takes effect.

## Revive the manifest-tree branching factor

The branching factor `m` was `manifest_rebalance_factor` in the old manifest machine, then re-implemented as a hardcoded `rebalance_threshold` (1024) in the replica reader core with no env fallback, accessor, or schema key. The dead `manifest_rebalance_factor` accessor was the only remaining operator knob for it. This revives it under the name the code and `manifest.md` already use: a `rebalance_threshold` accessor, a `stream_s3.rebalance_threshold` schema key, a snippet, a defaults-test entry, and `operations.md` docs. The reader resolves group sizes from each group header, so varying the factor does not affect reading existing manifests.

## Remove dead configuration surface

- `manifest_debounce_modifications`, `manifest_debounce_milliseconds`, `manifest_rebalance_factor` accessors: belonged to the old manifest machine, no callers
- `stream_s3.account_id` and `stream_s3.read_ahead` schema keys (and the `read_ahead`-only information-unit validator): read nowhere; `read_ahead` was superseded by the remote reader's AIMD adaptive prefetch sizing

## Testing

- `gmake test-build`: clean
- config module eunit: 34/34 pass
- `make ct-config_schema`: passes (`run_snippets`, all 20 snippets translate to the expected app env through the real config generator)
- dialyzer on the changed modules: clean


priv/schema/rabbitmq_stream_s3.schema
```diff
@@ -89,29 +77,6 @@ fun(Conf) ->
 end}.
 
 
-{mapping, "stream_s3.read_ahead", "rabbitmq_stream_s3.read_ahead", [
```
the-mikedavis, line 92: I think this was initially a shim for the osiris config for read-ahead. But it looks like dead code, especially since https://github.com/rabbitmq/rabbitmq-server/pull/14888 was merged in the server

the-mikedavis (approved):

src/rabbitmq_stream_s3_config.erl
```diff
@@ -112,17 +115,30 @@ general_pool_min_size() ->
...
+    application:get_env(?APP, persist_interval_ms, 2000).
+
+%% Manifest-tree branching factor: the number of same-kind leading entries in
+%% the root that triggers factoring them out into a group of the next-higher
+%% kind. A smaller value reduces memory footprint but increases the number of
+%% remote-tier requests during factoring and search; a larger value does the
+%% reverse.
+-spec rebalance_threshold() -> pos_integer().
+rebalance_threshold() ->
+    application:get_env(?APP, rebalance_threshold, 1024).
```
the-mikedavis, line 141: This is fairly high (mainly so that we can support really ridiculously sized streams. It would be fun to do some testing at some point where we tune this down, so we sus out bugs that only happen when the manifest tree hits a higher depth.

PR #296, opened 2026-06-26
Improve CLI commands

the-mikedavis:
Two improvements to the CLI commands:

* Allow them to be run from any node. For most commands this means redirecting to the node with the replica reader process. For local retention evaluation this means that we can run local retention against the requested node, even if it is not the writer.
* Improve the output of the `stream_s3_status` command. It should be human-readable. The new command adds comma-separators for large numbers and formats byte values in human readable format.


lukebakken (approved):
🚀

PR #299, opened 2026-06-26
manifest replica: Clean up when a replica (reader) exits

the-mikedavis:
The metadata stored in the manifest replica process did not have a clearing strategy. We can have this process monitor processes from the plugin and clear out metadata when the process exits. This avoids an incremental build of metadata over a long time.

### DIFF
--- a/src/rabbitmq_stream_s3_hooks.erl
+++ b/src/rabbitmq_stream_s3_hooks.erl
@@ -80,7 +80,7 @@ on_init(writer, Pid, #{name := Name, dir := Dir, shared := Shared, counter := Co
             )
     end,
     append_retention(StreamId, Config);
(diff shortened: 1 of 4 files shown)


lukebakken (approved):
LGTM. Real fix for the unbounded manifest_replica metadata growth.

Verified locally: build clean, dialyzer clean, all three new CT cases pass. The monitor bookkeeping is the tricky part and it's correct: re-registration demonitors + flushes the old ref before installing the new one, and the DOWN handler ignores superseded refs, so a member restart can't evict the live context. The `stopping` gate on the writer-node `forget/1` is also right: it fires only on genuine stream deletion, so a transient reader restart keeps the cached row (no window where a consumer sees the remote tier as absent).

Optional, non-blocking: a reconciler-re-registration test would round out monitor coverage, but the path is shared with the cases already tested.

Issue #302, opened 2026-06-29
Integration tests cannot catch a retention-induced fragment skip (gap detection and retention never overlap)


lukebakken:
## Summary

The integration suite has a blind spot that let a silent consumer-data-loss bug (#173) go unobserved across many soak runs: no test exercises remote-tier retention *and* asserts read contiguity at the same time. The one test that detects gaps deliberately disables retention, and the tests that run heavy retention only assert a message count, not contiguity.

## Why the gap exists

- **`content-verification`** is the only test that verifies ordering, no-duplicates, and no-gaps across an independent from-`first` replay. But by design it sizes `max-length-bytes` large enough that retention never fires during the run, specifically so a legitimate retention-overtake gap (#237) does not produce a false-positive failure. A retention deletion is exactly the trigger for the #173 class of bug, so this test is never put in a state that can produce the gap it is built to detect.
- **`high-throughput`** runs aggressive retention (so the triggering conditions occur), but its replay pass gate is `consumed >= 95% of expected` - a count threshold, not a contiguity check. A silently skipped fragment is well within the 5% slack and is invisible to a count. That slack exists because retention overtake legitimately skips data, which is the same slack that hides a real bug-induced skip.
- **`multi-offset-consumer`** has a lagging from-`first` consumer that reads the remote tier under retention, but it asserts progress, not contiguity.

Net: gap detection and active retention never overlap in a single test, so a retention-induced fragment skip falls into the seam between "the test that can detect gaps" and "the tests that run the triggering conditions."

## Proposed work

Add a content-verification variant (or extend the existing one) that runs **with retention active** and distinguishes a legitimate retention trim from a real skip:

- Drive enough load that remote-tier retention advances the readable floor during replay.
- For each consumer, track the readable floor (`first_offset`) at replay start and end. Classify any gap:
  - A gap entirely **below the floor** (offsets trimmed by retention) is legitimate - the floor moved past them.
  - A gap **at or above the floor** (a fragment still within the retained range was skipped) is a bug - fail the test.
- Keep the existing retention-free content-verification as the strict no-gap-at-all check; this new variant is the retention-aware contiguity check.

This turns the count-based 95% tolerance into a floor-aware contiguity assertion, so a retention-induced fragment skip (the #173 class) is caught rather than absorbed by slack.

## Related

- #173 (the silent fragment-skip bug this gap failed to catch)
- #237 (retention overtake: the legitimate-gap case the assertion must not flag)
[...]

PR #303, opened 2026-06-29
Refresh past the current fragment's offset when it 404s (#173)

### LINKED ISSUE #173: next_fragment_offset/1 can return a live fragment offset, causing silent skip
## Summary

`rabbitmq_stream_s3_remote_reader_core:next_fragment_offset/1` can return the offset of a *live* fragment instead of the 404'd one, causing the log reader to silently skip that live fragment on a specific race.

## Background

The function was introduced in PR #169 (commit 98ac271, "log reader: Refresh fragment iterator in 404s") and a clarification comment was added in 420eb56 stating that it is only called from the `not_found_check_range` branch of `try_serve` and that `next = not_found` guarantees the iterator is positioned at the 404'd entry.

That clarification is incomplete. There are two paths into `{not_found_check_range, State}`:

**Path A** -- `try_fragment_transition` matching `next = not_found`. The consumer has read past the current fragment and the prefetched-next 404'd. The iterator points at the 404'd entry. `next_fragment_offset/1` returns the 404'd offset. Correct.

**Path B** -- `try_read` matching `current_not_found = true`. The *current* fragment 404'd while there was no pending read; the partial buffer is still in place; a later read arrives wanting bytes past the buffer. `try_read` returns `{not_found_check_range, State}`. But the iterator was advanced past the current fragment in `init/5` (or `goto_next_fragment/1`), so it points at the entry AFTER the 404'd current fragment. `next_fragment_offset/1` returns that entry's offset, which may be a perfectly live fragment.

## Consequence

In Path B, `try_serve` emits `{refresh_iterator, NotFoundOffset}` where `NotFoundOffset` is the live next fragment's offset. The shell's `refresh_iterator/2` initializes the iterator at `NotFoundOffset`, then advances past anything `=< NotFoundOffset`. The core's `step({iterator_refreshed, Iterator}, ...)` reads `next/1` which returns the entry AFTER the live fragment, and replies `{next_fragment, Offset}` for it. The log reader repositions there, silently skipping the live fragment.

End-state: data loss visible to the consumer (records from the skipped fragment are not delivered, and `next_offset` advances past them).

## Reachability

The race requires:
1. The remote reader is initialized and starts a request immediately (or a request is in flight).
2. The current fragment returns 404 before any `read` call arrives. The shell forwards the error to the core, which sets `current_not_found = true` and (because `pending = undefined`) emits no effect.
3. A `read` arrives later requesting bytes past the partial buffer.

This is plausible when retention deletes the head fragment between the moment a remote reader is opened and the first read call from the log reader. It is also reachable when an in-flight request 404s before the consumer issues its first read.

## Test gap

The existing 404 tests (`not_found_triggers_refresh_iterator`, `fragment_404_advances_to_next_fragment`, `fragment_404_emits_refresh_iterator`) all issue `{read, ...}` *before* the 404 arrives, so they exercise Path A and `step({request_error, _, _, not_found}, ...)` emits `{refresh_iterator, Fragment}` directly with the 404'd offset, never going through `try_read` -> `not_found_check_range`. Path B is uncovered.

## Suggested fix

Either:

1. Pass the 404'd offset alongside the `not_found_check_range` tag so the caller has the right value regardless of which path produced it (e.g. `{not_found_check_range, NotFoundOffset, State}`).
2. In Path B, use `current_fragment_offset(State)` instead of `next_fragment_offset(State)` since the current fragment is the one that 404'd.

Option 2 is the smaller diff. Option 1 is cleaner because the data origin is explicit.

Add a regression test exercising the order: init -> 404 on current -> read past partial buffer.

## Origin

Found while reviewing https://github.com/amazon-mq/rabbitmq-stream-s3/pull/169.

lukebakken:
Closes #173

## The bug (silent consumer data loss)

The `not_found_check_range` handler in `rabbitmq_stream_s3_remote_reader_core` refreshed the iterator past `next_fragment_offset` for both 404 paths, but the two paths leave the iterator pointing at different entries:

- When the prefetched *next* fragment 404s (`next = not_found`, via `try_fragment_transition`), the iterator points at that 404'd entry, so refreshing past it is correct.
- When the *current* fragment 404s with no read pending (`current_not_found = true`), the iterator was already advanced past the current entry at init, so it points at the *next* entry, which is still live. Refreshing past that offset made the shell's `refresh_iterator/2` (which advances past anything `=< NotFoundOffset`) skip the live fragment, silently dropping a contiguous block of consumer data with no error.

Concrete example with fragments at offsets [100, 200, 300] reading 100: 100 404s, the refresh was told `200`, and the reader resumed at `300`. Fragment 200 was live and silently skipped.

## The fix

Split the decision into `not_found_refresh/1`:

- `current_not_found = true` (current fragment 404'd): refresh past the current fragment's own offset, the one that actually 404'd, keeping the live next fragment.
- otherwise (prefetched next 404'd): unchanged, refresh past `next_fragment_offset` with the `end_of_manifest` / `group_fetch_failed` handling intact.

Refreshing past the current (smaller) offset is always safe: `refresh_iterator/2` re-derives position from the live manifest and lands on the first surviving fragment greater than the target, so a smaller target can never skip a live fragment. Passing the larger `next` offset was the only choice that could.

## Tests

Two regression tests, both confirmed to fail on the pre-fix code and pass with the fix:

- `current_fragment_404_no_pending_read_keeps_next_live_fragment`: the primary path, a multi-fragment manifest with the 404-before-read ordering that the existing single-fragment 404 tests never exercised.
- `current_and_next_404_read_past_current_keeps_surviving_fragment`: the subtle interleaving where the current fragment and its prefetched next both 404 before a read, then the consumer reads past the current fragment. Confirms the refresh still targets the current offset and the surviving fragment is not skipped.

`remote_reader_core_SUITE` and `read_resolution_SUITE` pass; dialyzer clean.

## Note on coverage

This is verified at the functional-core level (deterministic state-machine paths). It is not yet exercised end-to-end against live S3 retention; the integration suite cannot currently catch a retention-induced fragment skip because gap detection and active retention never overlap in a single test. That gap is tracked separately in #302.

### DIFF
--- a/src/rabbitmq_stream_s3_remote_reader_core.erl
+++ b/src/rabbitmq_stream_s3_remote_reader_core.erl
@@ -369,19 +369,7 @@ try_serve(#state{pending = #pending{offset = Offset, bytes = Bytes}} = State) ->
             {State3, Effects} = maybe_start_requests(State2),
             {State3, [{observe, miss, State3#state.read_size} | Effects]};
         {not_found_check_range, State1} ->
-            %% Fragment was 404. Refresh iterator past it. If the iterator
-            %% is exhausted (last fragment in manifest was deleted), fall
-            %% back to refreshing past the current fragment's offset.
-            case next_fragment_offset(State1) of
-                {ok, NotFoundOffset} ->
-                    {State1, [{refresh_iterator, NotFoundOffset}]};
-                end_of_manifest ->
-                    {State1, [{refresh_iterator, current_fragment_offset(State1)}]};
-                group_fetch_failed ->
-                    %% Probing the next entry hit a transient group fetch
-                    %% error. Retry rather than becoming local.
-                    retry_group_fetch(State1)
-            end;
+            not_found_refresh(State1);
         {refresh_iterator, State1} ->
             %% Iterator exhausted. Refresh past current fragment.
             {State1, [{refresh_iterator, current_fragment_offset(State1)}]};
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):
Surgical 🔪

PR #307, opened 2026-06-29
Evaluate local retention when a manifest is resolved or synced (#75)

### LINKED ISSUE #75: Trigger local retention after resolving a manifest
When a manifest is resolved, a stream might be able to use the new information about the remote tier to clean up redundant local data. This would be useful for a replica which is offline for a while (because of an upgrade for example) which has a stream which stops publishing when the replica is brought back online. Currently it would replicate all of the data from the writer but not clean up any redundant data.

lukebakken:
Closes #75

## Problem

Local retention reclaims local segments whose data is already durable in the remote tier. It was only ever triggered by a manifest persist, which a new publish drives. So a stream that learns about the remote tier and then stays idle never reclaimed its redundant local data.

This is the scenario in #75: a replica that was offline (an upgrade, for example), rejoins and catches up from the writer, and whose stream then stops publishing. It replicates all the data but never cleans up the local copies already in S3.

## Fix

Two paths learn about the remote tier without a publish following, and neither evaluated retention:

- **Replica node** (the case #75 describes): `manifest_replica:maybe_apply_sync` applies a full sync but did not evaluate retention; only edits that advance `next_offset` did. A returning replica receives a full *sync*, not incremental edits, so retention never ran. Now a one-shot local retention pass runs after a non-stale sync, guarded on a registered replica context and a non-empty manifest.

- **Writer node**: `replica_reader:on_manifest_resolved` seeds counters on resolution but did not evaluate retention. Add the same one-shot pass so a reader that resolves a non-empty manifest and then stays idle reclaims too. This is defensive: on the writer node the persist-driven pass normally runs first, so this rarely has work to do, but it closes the same gap on that side.

Both reuse the existing local-retention evaluation, which only deletes segments fully below the manifest's `next_offset` (which is where the reader opens), so it cannot trim data a reader is about to read.

## Test

`sync_triggers_local_retention_on_idle_replica` seeds a real multi-segment local log, registers a replica context pointing at it, and syncs a manifest covering all the data, with no edits and no new writes. The older segments are reclaimed by the sync alone. The test fails without the fix (the await times out because the segments are never reclaimed) and passes with it.

The writer-side path calls the same `maybe_evaluate_retention` already exercised by `retention_reclaims_uploaded_segments` and the CLI `evaluate_local_retention`. A dedicated writer-side test was not added: on the writer node the persist-driven retention runs during the initial upload, so constructing uploaded-but-untrimmed local segments at resolve time requires fragile state-forging for a defensive path; the replica-side test covers the substantive scenario end to end.

Full `replica_reader_SUITE` and `manifest_replica_SUITE` pass; dialyzer clean.

## Note

The issue's framing pointed at the writer side, but the substantive gap turned out to be the replica sync path. This was a touch larger than the `E-small` label suggested; closer to `E-medium`.

### DIFF
--- a/src/rabbitmq_stream_s3_manifest_replica.erl
+++ b/src/rabbitmq_stream_s3_manifest_replica.erl
@@ -491,8 +491,36 @@ maybe_apply_sync(StreamId, Seq, Epoch, Manifest, WriterNode, Seqs, Ctxs) ->
         false ->
             write_manifest(StreamId, Manifest, Epoch),
             seed_first_offset_counter(StreamId, Manifest, Ctxs),
+            %% Evaluate local retention once against the freshly synced manifest.
+            %% On a replica, retention is otherwise only driven by edits that
+            %% advance next_offset; a replica that was offline (an upgrade, say),
+            %% receives a full sync on rejoin, and whose stream then stops
+            %% publishing would never reclaim the local segments already durable
+            %% in the remote tier. This one-shot pass reclaims them. It is a no-op
+            %% for an empty manifest or a stream with no registered context here
+            %% (issue #75).
+            maybe_evaluate_retention_on_sync(StreamId, Manifest, Ctxs),
             Seqs#{StreamId => {Seq, Epoch, WriterNode}}
     end.
+
+%% Run local retention against a just-synced manifest, guarding on the same
+%% conditions run_local_retention requires: a registered replica context on this
+%% node and a non-empty manifest (next_offset > 0, so the first_timestamp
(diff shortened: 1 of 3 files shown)


the-mikedavis:
There's another way it can trigger: if a segment is rolled then osiris itself will kick off retention. But either way you need to publish to the stream in order to get retention to kick in, so idle streams might be sitting on data that can be reclaimed.

the-mikedavis (approved):

the-mikedavis (approved):

Issue #309, opened 2026-06-29
Trigger garbage collection periodically and automatically


lukebakken:
## Summary

GC (`rabbitmq_stream_s3_gc:run/1`) is currently only invoked on demand via the `rabbitmq-streams` CLI (and the per-stream `run_stream/2` path used by the replica reader after a manifest reset). There is no periodic, automatic cross-stream sweep. PR #204 noted this when it landed the GC machinery ("In the long run we can have a periodic automatic trigger").

## Why it matters now

The deletion-tombstone fix in #196 makes GC the backstop that reclaims a deleted stream's straggler objects (an upload that completes after the one-shot deletion sweep). The reaper clears the common case immediately, but a straggler left by the race is only reclaimed on the next GC run. With GC being CLI-only, that means a straggler object, and its tombstone, persist until an operator manually runs GC.

This is still strictly better than the pre-#196 behavior (a permanent leak with no recourse), but it leaves cleanup dependent on manual action. A periodic automatic GC would make both straggler reclamation and tombstone cleanup self-healing without operator involvement.

## Proposed work

- Add a periodic trigger that runs a delete-mode sweep on an interval (configurable, off or conservative by default given GC does one strongly-consistent metadata read per stream).
- Run it on a single node at a time. PR #204 already flagged this: "We could also hold a global lock so that GC is only performed by one process at one time." A cluster-wide lock (or leader-only execution) avoids redundant concurrent sweeps.
- Keep the existing CLI entry point for on-demand runs.

## Related

- #196 (deletion tombstone; relies on GC as the straggler backstop)
- #204 (added the GC machinery; flagged the periodic trigger and global lock as future work)

Issue #149, opened 2026-05-19
Garbage-collect dangling remote tier objects


the-mikedavis:
## Why

Objects in S3 can become orphaned (not referenced by any manifest) in several scenarios:

- A deposed writer uploads a fragment but loses the Khepri race. The fragment is in S3 but never referenced.
- A manifest root is uploaded but the Khepri write fails or the process crashes before completion.
- A node's data directory is wiped and restarted. Khepri is recreated from scratch, so the keep-while condition that triggers S3 cleanup never fires. The S3 objects remain indefinitely.
- Stream deletion cleanup is interrupted (reaper task crashes, S3 outage).

These orphans waste storage. Without GC, they accumulate indefinitely.

## Plan

A background process periodically LISTs objects under the `rabbitmq/stream/` prefix and identifies orphans by comparing object keys against the current state in Khepri.

Detection relies on the fact that all object keys encode enough information to determine whether they belong to the past:

- Fragments and groups have the first offset in their key. The manifest's `first_offset` advances monotonically with retention. Any object whose offset is below the manifest's `first_offset` is an orphan (retention deleted it but the object wasn't cleaned up).
- Manifest roots have the epoch in their key (`root.<epoch>.<uid>.manifest`, since https://github.com/amazon-mq/rabbitmq-stream-s3/pull/148/commits/90a06d731f89c49629f17e8db31a9d8a4a5a040f). The current epoch is stored in Khepri. Any root with an epoch below the current epoch is from a previous writer and safe to delete.
- Stream prefixes with no Khepri entry are entirely orphaned (stream was deleted but cleanup didn't complete). All objects under the prefix can be deleted.

This design requires no coordination with replica readers. The GC process only reads from Khepri (which is eventually consistent but always moves forward) and compares against S3 listings. False negatives (missing an orphan) are acceptable since GC runs periodically. False positives (deleting a live object) are impossible because offsets and epochs only move forward.

the-mikedavis:
The garbage collection process is meant to make sure objects don't sit in the remote tier, dangling, forever. It should be rare that we actually delete objects because of garbage collection, though. It should only occur when we plan a deletion and pass the object (or stream) to the reaper but it cannot complete the deletion, probably because RabbitMQ shuts down.

Since we expect it to be rare, we should track metrics for it but also log at info level whenever a key is deleted for garbage collection. This will help us audit how well garbage collection is working.

the-mikedavis:
[...]

PR #310, opened 2026-06-29
Reclaim a deleted stream's straggler objects via a tombstone (#196)

### LINKED ISSUE #196: Stream deletion leaves one dangling S3 fragment from in-flight upload
## Summary

When a stream is deleted, the Khepri trigger deletes S3 objects but consistently leaves one fragment behind. The dangling object persists indefinitely.

## Reproduction

Run `make stream-deletion` from `test/integration/`. The test publishes until S3 is populated (14 objects), deletes the stream, then polls until all objects are removed. It consistently fails with 1 object remaining after 60 seconds.

Reproduced twice:
- Run 1: `rabbitmq/stream/__s3-stream-deletion_1780923100499341566/data/00000000000000986214.98f7bc7d.fragment`
- Run 2: `rabbitmq/stream/__s3-stream-deletion_1780923245568639694/data/00000000000000983270.36117c13.fragment`

Different stream instances, different fragments, same behavior - always exactly 1 object remains.

## Likely cause

The upload pipeline has an in-flight fragment that completes its S3 PUT after the Khepri deletion trigger has already swept the prefix. The trigger enumerates objects and deletes them, but a fragment uploaded during or after the sweep is missed.

The test stops publishing only ~5 seconds before deletion. The replica reader's upload pipeline is async - fragments are assembled, queued at the governor, and uploaded. A fragment that was queued before deletion but uploaded after the sweep would be orphaned.

## Expected behavior

All S3 objects under the stream's prefix should be removed after stream deletion, regardless of in-flight uploads at the time of deletion.

## Test output

```
12:54:10.682 [main] INFO  - S3 populated: 14 objects. Deleting stream: s3-stream-deletion
12:54:10.693 [main] INFO  - Waiting for S3 cleanup (max 60s)...
12:54:12.714 [main] INFO  -   s3-objects=1
... (stays at 1 for 60 seconds)
12:55:11.246 [main] ERROR - FAIL: 1 S3 objects remain after 60s
```

lukebakken:
Closes #196

## Problem

Stream deletion is a one-shot LIST-then-delete sweep. A fragment whose upload completes after the sweep has listed the prefix is missed, leaving a dangling S3 object. Orphan GC could not reclaim it either: after deletion the stream is gone from Khepri, so its leftover key classifies as an *unknown* stream, which GC deliberately skips (it cannot tell a deleted stream's prefix from a live stream this node merely cannot see). The object leaked indefinitely.

The earlier assumption that GC (#149/#204) would pick these up does not hold: #204 implemented GC for live streams only (offset-below-first-offset, stale-epoch), both of which require the stream to still exist in Khepri.

## Fix: a positive "stream deleted" signal

- **db**: write a tombstone at a separate Khepri root (`rabbitmq_stream_s3_deleted/<StreamId>`) with no keep-while, so it survives the keep-while removal of the stream entry that fires the deletion trigger. Payload carries the deletion time. Adds `write_tombstone/1`, `list_tombstones/0`, `delete_tombstone/1`. The tombstone is written by the reaper, **not** the trigger sproc: triggered sprocs run synchronously inside the single `khepri_event_handler` gen_server, where a synchronous `khepri:put` would block that handler on its own Ra command.

- **reaper**: write the tombstone before sweeping (from its own task process), then after the sweep flush the async delete casts (a new `gen_batch_server` call barrier) and re-LIST; if the prefix is empty, drop the tombstone. A straggler left by the race keeps the tombstone for GC.

- **gc**: classify an object whose stream is absent from the live lookup as an orphan when the stream is tombstoned (`reason => stream_deleted`), and clear the tombstone after a delete-mode sweep leaves the prefix empty. A tombstone is consulted **only** when the stream is absent from the lookup, so a live stream's objects are never deleted on the strength of a stale tombstone. `still_dangling` re-checks the tombstone immediately before deleting.

## StreamId aliasing

StreamIds embed a nanosecond timestamp, so a recreated stream gets a distinct id and cannot alias an existing tombstone.

## Tests

- gc eunit: tombstoned-stream reclaim, per-object-kind coverage, isolation between streams, and that a live stream ignores a tombstone (the last caught a real bug where a live stream's above-floor data would have been deleted).
- broker_SUITE: end-to-end GC reclaim of a straggler under a tombstone, asserting the tombstone is cleared; the existing clean-deletion test still passes.

Build clean, dialyzer clean.

## Verification notes

- The `gen_batch_server` flush barrier was verified against its source: it reverses the accumulated batch to arrival order, so the reaper's delete casts are processed before its flush reply.
- The trigger-sproc-cannot-write constraint was confirmed by reading the Khepri trigger dispatch (`khepri_machine` mod_call effect into `khepri_event_handler`).
- Cross-node correctness rests on Khepri replication of the tombstone (a deletion trigger fires on the Khepri leader; GC may run elsewhere), the same replicated-read mechanism the existing stream listing relies on.

## Follow-up

Tombstone cleanup happens on the reaper fast path or the next GC run. GC has no periodic trigger yet (it is CLI-only), so until one exists a straggler and its tombstone persist until a manual GC. Tracked in #309. This is still strictly better than the previous permanent leak with no recourse.

### DIFF
--- a/src/rabbitmq_stream_s3_reaper.erl
+++ b/src/rabbitmq_stream_s3_reaper.erl
@@ -73,6 +73,11 @@ init([]) ->
 handle_batch(Ops, State) ->
     Keys = collect(Ops),
     _ = delete_batched(Keys),
+    %% Reply to any flush barrier in this batch only after the deletes above
+    %% have run. gen_batch_server processes a batch in order and a caller's
+    %% {delete, _} casts are enqueued before its flush call, so a returning
+    %% flush guarantees those deletes are complete (see list_and_delete/1).
+    _ = [gen:reply(From, ok) || {call, From, flush} <- Ops],
     {ok, State}.
 
 terminate(_Reason, _State) ->
@@ -125,11 +130,43 @@ delete_batch(Batch) ->
 
 list_and_delete(StreamId) ->
     Prefix = rabbitmq_stream_s3:stream_prefix(StreamId),
+    %% Record the tombstone before sweeping. This runs in the reaper's spawned
+    %% task (a normal process), not the Khepri trigger sproc, so it can write to
+    %% Khepri. The tombstone is the durable "stream deleted" signal that lets GC
+    %% reclaim a fragment whose upload completes after this one-shot sweep lists
+    %% the prefix (issue #196). Best effort: if it fails the sweep still runs,
+    %% and the worst case is the pre-existing leak this addresses.
+    _ = rabbitmq_stream_s3_db:write_tombstone(StreamId),
     Result = list_and_delete_loop(Prefix, start),
     case Result of
-        ok -> inc(?C_STREAMS_DELETED, 1);
-        _ -> ok
+        ok ->
+            inc(?C_STREAMS_DELETED, 1),
+            maybe_clear_tombstone(StreamId, Prefix);
+        _ ->
+            %% The sweep hit a LIST error and is incomplete; leave the tombstone
+            %% so GC reclaims whatever remains and clears it later.
+            ok
     end.
+
+%% After a clean sweep, confirm the prefix is empty and drop the tombstone so it
+%% does not accumulate. The deletes are async casts to this server, so flush
+%% first to ensure they have run, then re-LIST. If anything remains (an upload
+%% that completed after the sweep listed its page, the #196 race), keep the
+%% tombstone: GC will reclaim the straggler and clear the tombstone then.
+maybe_clear_tombstone(StreamId, Prefix) ->
+    _ = flush(),
+    case rabbitmq_stream_s3_api:list(Prefix, start) of
+        {ok, [], done} ->
+            _ = rabbitmq_stream_s3_db:delete_tombstone(StreamId),
+            ok;
+        _ ->
+            ok
+    end.
+
+%% Synchronous barrier: returns only after all delete casts enqueued before it
+%% have been processed by handle_batch/2.
+flush() ->
+    gen_batch_server:call(?MODULE, flush).
 
 list_and_delete_loop(_Prefix, done) ->
     ok;
--- a/src/rabbitmq_stream_s3_db.erl
+++ b/src/rabbitmq_stream_s3_db.erl
@@ -27,6 +27,17 @@ writer cannot make progress anymore.
 -include_lib("rabbit_common/include/resource.hrl").
 
 -define(PATH(StreamId), ?RABBITMQ_KHEPRI_ROOT_PATH([rabbitmq_stream_s3, StreamId])).
+%% Tombstone path for a deleted stream. Deliberately under a separate root from
+%% ?PATH/1 and with no keep-while condition, so it survives the keep-while
+%% removal of the stream entry (that removal is what fires the deletion trigger).
+%% A tombstone records that a stream's remote-tier prefix holds no live data and
+%% may be swept in full, the positive "stream deleted" signal that lets GC
+%% reclaim objects left by an upload that completed after the deletion sweep
+%% (issue #196). StreamIds embed a nanosecond timestamp, so a recreated stream
+%% gets a distinct id and cannot alias an existing tombstone.
+-define(TOMBSTONE_PATH(StreamId),
+    ?RABBITMQ_KHEPRI_ROOT_PATH([rabbitmq_stream_s3_deleted, StreamId])
+).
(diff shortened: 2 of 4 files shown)


src/rabbitmq_stream_s3_reaper.erl
```diff
@@ -125,12 +130,44 @@ delete_batch(Batch) ->
...
 list_and_delete(StreamId) ->
     Prefix = rabbitmq_stream_s3:stream_prefix(StreamId),
+    %% Record the tombstone before sweeping. This runs in the reaper's spawned
+    %% task (a normal process), not the Khepri trigger sproc, so it can write to
+    %% Khepri. The tombstone is the durable "stream deleted" signal that lets GC
+    %% reclaim a fragment whose upload completes after this one-shot sweep lists
+    %% the prefix (issue #196). Best effort: if it fails the sweep still runs,
+    %% and the worst case is the pre-existing leak this addresses.
+    _ = rabbitmq_stream_s3_db:write_tombstone(StreamId),
     Result = list_and_delete_loop(Prefix, start),
```
the-mikedavis, line 140: I think that we can close this window actually - we can ensure that we write an anchor to Khepri before we attempt to upload the first fragment. I'm working on a P model for this case

the-mikedavis (changes requested):
I've got some opinions on this one - give me a bit to work out the models so I can give a full review

src/rabbitmq_stream_s3_reaper.erl
lukebakken: (Posted by Luke's AI coding agent, with his guidance.)

Agreed, that's a better approach. Writing an anchor to Khepri before the first fragment upload closes the race at the source (no straggler is ever created), rather than this PR's tombstone approach, which is reactive: it lets the straggler exist and relies on GC to reclaim it later. Your version also keeps GC in the rare, last-resort role it is meant for (per your comments on #149), instead of making it the routine cleanup path for every stream deletion.

Happy to rework this PR around the anchor once your P model settles, or to close it in favour of your change if you would rather drive it, since the modeling is the hard part. Either way I will hold off on further changes here until your model is ready.

One finding from this PR that should carry over to the anchor implementation: the deletion trigger sproc cannot write to Khepri itself. Triggered sprocs run synchronously inside the single `khepri_event_handler` gen_server (Khepri dispatches them there via a `mod_call` effect), so a synchronous `khepri:put` from the sproc blocks that handler on its own Ra command. The anchor write will need to happen from the writer/upload path rather than from any deletion-trigger sproc.

lukebakken: @the-mikedavis feel free to modify this PR branch directly, of course!

the-mikedavis: This is what I have in mind for the model: https://github.com/amazon-mq/rabbitmq-stream-s3/pull/311

The idea is that we can write an 'anchor' before we write the first fragment. The absence of the anchor plus objects in the stream in S3 means that there is garbage to be collected. We need to be careful to use a consistent query to check for the anchor, though.

I will get to this change eventually if you don't beat me to it @lukebakken !

the-mikedavis: This design covers a few gaps I think are worth calling out:

* The issue (#196) is covered. Stream deletion gets rid of the anchor. If a fragment is uploaded just after stream deletion, it is anchor-less and so it is considered garbage.
* The review here (alternative tombstone model) is covered. If the tombstone write fails then the stream can't be detected as garbage. As the comment says, the tombstone write is best-effort. In contrast, the anchor must be written in order for a fragment to be written. So it provides a strong ordering guarantee that the GC can rely on.
* It improves over the idea in https://github.com/rabbitmq/khepri/issues/358. I was thinking we could structure the sproc and keep-while conditions so that the manifest pointer entry in the db would be dangling after a stream deletion. Then GC could see that it's dangling (no queue resource) as a way to clean up garbage. But this creates a gap where, if a fragment is uploaded and no manifest is ever uploaded and written to the db, objects from the stream can't be considered garbage.

lukebakken:
Closing PR in great anticipation of @the-mikedavis's implementation 😸

lukebakken:
Removing this from the 1.0.0-beta.1 milestone: it was closed without merging, so no release contains anything from it. The tombstone approach here was not the one that shipped - the anchor-based reap of deleted-stream prefixes landed in #316 as bf2fcfc. #196 remains open.

PR #314, opened 2026-06-30
Probe remote tier bucket accessibility and surface it (#3)

### LINKED ISSUE #3: Creating a new stream with wrong bucket config in rabbit.conf
**Describe the bug**

Two issues:

If you configure your `s3.bucket` with a bucket that your account does not have permission to use, you will see the following crash:

```
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>   crasher:
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     initial call: osiris_writer:init/1
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     pid: <0.765.0>
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     registered_name: []
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     exception exit: {timeout,
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                         {gen_server,call,
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                             [rabbitmq_stream_s3_log_manifest,
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                              {init_writer,
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                                  {resource,<<"/">>,queue,<<"s1">>},
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                                  "/.../__s1_1763145000075864515"}]}}
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>       in function  gen_batch_server:handle_continue/2 (src/gen_batch_server.erl, line 423)
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>       in call from gen_batch_server:init_it/6 (src/gen_batch_server.erl, line 163)
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     ancestors: [osiris_server_sup,osiris_sup,<0.136.0>]
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     message_queue_len: 1
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     messages: [{'$gen_cast',
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                       {register_offset_listener,<0.399.0>,
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                           {rabbitmq_stream_s3_log_manifest,
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                               format_osiris_event,[]},
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>                           -1}}]
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     links: [<0.140.0>]
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     dictionary: []
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     trap_exit: true
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     status: running
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     heap_size: 1598
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     stack_size: 36
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>     reductions: 5586
2025-11-14 18:41:38.436302+00:00 [error] <0.765.0>   neighbours:

```

Second issue:
If you instead configure a bucket name that does not even exist, it 'seems' to work fine.

**To Reproduce**
Configure `s3.bucket` to a bucket that does not exist or one that you don't have the correct permissions for.

**Expected behavior**

There need to be some kind of error stating that we cannot access the bucket, in both cases.

**Screenshots**
N/A

**Additional context**

lukebakken:
Fixes #3.

## Background

The original crash in #3 (a `gen_server:call` timeout taking down `osiris_writer:init/1`) is already gone: the writer hook starts the replica reader asynchronously and only logs a warning on failure, so a bad bucket cannot crash log init. The stream still functions on local disk.

What remained is the original ask: an early, clear error when the configured bucket is wrong or unreachable. Today a misconfiguration is nearly silent. With a nonexistent or wrong-permission bucket, uploads fail at runtime and retry indefinitely on a backoff; the only operator-visible signal is the `response_403` counter plus per-attempt logs. A nonexistent bucket "seems to work fine".

## What this does

A new node-level `rabbitmq_stream_s3_bucket_monitor` singleton issues a `HeadBucket` probe at startup (deferred until credentials are available) and on a slow interval. It maps `404` to "does not exist" and `403` to "access denied", and surfaces the result three ways:

- an `ERROR` log on the transition into inaccessibility (and `INFO` on recovery), logged on transition only so a persistent misconfig does not log every tick
- a node-level `rabbitmq_stream_s3_bucket_accessible` gauge (1/0) for alerting
- the `stream_s3_status` CLI command, in a new "Remote tier bucket" section

The probe runs in a monitored child process so a slow `HeadBucket` never blocks the monitor (or `status/0`). A non-definitive result (credentials not yet loaded, a transient error) neither flips the gauge nor changes the reported status; the gauge starts optimistically at 1 so boot does not report a false inaccessibility, and the monitor retries quickly until it reaches a definitive answer, then backs off to the configured interval.

## Why no `rabbit_alarm`

I considered raising an alarm and decided against it. The only alarm primitive is `{resource_limit, Source, node()}`, and `rabbit_reader` (and the AMQP/stream readers) flow-block publishers for *any* source, unfiltered. Raising an alarm for an inaccessible bucket would block every publisher on the node, across all protocols and queue types, turning degraded tiering (streams still work on local disk) into a publishing outage. The signal here is loud but side-effect free.

## Surface area

- new `check_bucket/1` API behaviour callback, implemented by both the AWS (`HeadBucket`) and file-system backends
- new config `stream_s3.bucket_check.enabled` (default `true`) and `stream_s3.bucket_check.interval` (default 5 minutes), with cuttlefish mappings
- docs: configuration, the new metric, an alerting recommendation, and a troubleshooting entry ("Nothing is tiered to the remote tier")

## Testing

- eunit: full suite (172), monitor state machine (transitions, ERROR-on-transition-only, transient-no-flip, boot deferral, gauge), config defaults
- CT: `api_fs` (incl. `check_bucket`), `rabbitmq_stream_s3_api_aws`, full `broker_SUITE` 7/7 including a new `bucket_monitor_reports_accessible` case asserting the live monitor converges and the gauge is exposed on `/metrics`
- `dialyze` and `xref` clean


the-mikedavis:
I think this looks good but I want to push the milestone back a bit so we have time to test it well. I'll move the milestone around.

lukebakken (commented):
Re-reviewed after the rebase onto main (now includes #323/#324) and the added 403/404 notes.

Design holds up: node-level singleton, no rabbit_alarm (correctly avoids flow-blocking every publisher for a tiering-only condition), fail-open with a definitive-vs-transient split, optimistic gauge at boot, and an isolated spawn_monitor probe so a slow HeadBucket never blocks status/0.

Verified against source: request/5 takes a binary method (HEAD on / is valid), log_unexpected_status/2 and api_fs data_dir/0 exist, the CLI degrades to "unknown" on a badrpc from an older/disabled node, and every docs reference resolves (stream_s3.bucket, stream_s3.region, response_403, metric name, config keys/defaults).

The 403/404 nuance is now documented: a nonexistent bucket can report access_denied when credentials lack s3:ListBucket, so both reasons mean "not usable" with the same remedy; the monitor treats them identically and only the reason differs.

Local verification, all green: eunit 187/187 (monitor state machine + config defaults), ct-api_fs check_bucket, ct-broker 7/7 including bucket_monitor_reports_accessible (live convergence + gauge on /metrics), dialyze and xref clean.

src/rabbitmq_stream_s3_api.erl
```diff
@@ -92,6 +94,17 @@ examples.
...
+-doc """
+Probe whether the configured bucket exists and is accessible with the current
+credentials.
+
+`no_such_bucket` and `access_denied` are definitive: the bucket is misconfigured
+or unreadable and a monitor should surface that loudly. Any other `{error, _}`
+(credentials not yet available, a transient network error, a throttling
+response) is non-definitive and a caller must not treat it as inaccessible.
+""".
+-callback check_bucket(request_opts()) ->
```
the-mikedavis, line 106: I think the `api_fault` backend needs to add `check_bucket/1`

src/Elixir.RabbitMQ.CLI.Ctl.Commands.StreamS3StatusCommand.erl
```diff
@@ -48,13 +48,38 @@ usage_additional() ->
...
+            NodeName,
+            rabbitmq_stream_s3_bucket_monitor,
+            status,
+            [],
+            Timeout
+        )
+    of
+        #{status := _} = Status -> Status;
+        _ -> undefined
+    end.
```
the-mikedavis: I wonder if we should move this into `rabbitmq_stream_s3_replica_reader:status/2`? Then we can avoid a double round-trip. That function is really only meant to be used for the CLI so I think taking on extra responsibility is ok here

src/rabbitmq_stream_s3_bucket_monitor.erl
```diff
@@ -0,0 +1,325 @@
...
+
+-spec init_counters() -> ok.
+init_counters() ->
+    Cnt = seshat:new(rabbitmq_stream_s3, ?MODULE, ?COUNTERS, #{module => ?MODULE}),
+    %% Start optimistic: assume the bucket is accessible until a probe proves
+    %% otherwise, so boot (before the first probe completes) does not report a
+    %% false inaccessibility.
+    counters:put(Cnt, ?C_BUCKET_ACCESSIBLE, 1),
+    persistent_term:put(?COUNTER_KEY, Cnt),
+    ok.
```
the-mikedavis, line 239: I think the gen_server could init the counters if the feature is enabled. That would avoid showing the counter in the prometheus output if the feature is not enabled. I think currently we would show `1` even if the bucket is not actually accessible (if the feature is not enabled).

src/rabbitmq_stream_s3_api.erl
lukebakken: Good catch, that was an incomplete behaviour implementation. Added `check_bucket/1` to the fault backend, delegating to the FS backend like the other non-injected callbacks, in 9e3c124.

src/Elixir.RabbitMQ.CLI.Ctl.Commands.StreamS3StatusCommand.erl
lukebakken: Done in 0d5c650. `status/2` now folds in the node-level bucket accessibility itself (a local call on the node already serving the request), so the CLI is back to a single round-trip. A missing monitor yields `undefined`, rendered as "unknown" as before.

src/rabbitmq_stream_s3_bucket_monitor.erl
lukebakken: Agreed, showing `1` for a probe that never runs was misleading. Moved the counter creation into the monitor's `init/1`, gated on the same `bucket_check_enabled()` branch, so a disabled node exposes no `bucket_accessible` gauge at all. Added an eunit test for the disabled path. Done in 80e4bbd.

Issue #212, opened 2026-06-17
Improve retry strategy for failures in upload path


the-mikedavis:
https://github.com/amazon-mq/rabbitmq-stream-s3/pull/210#issuecomment-4731860585 pointed out a bunch of room for improvement in how we retry transient and also unexpected errors in the upload pipeline. We should audit the current naive retry strategy and improve it for the sake of production-readiness.

lukebakken:
Scope update (partially addressed). The safety and classification asks have shipped; the backoff and observability asks have not.

Done:
- Error classification widened: `is_retriable/1` now matches `internal_error`, `slow_down`, `connection_error`, `timeout`, `transfer_deadline`, and `#{status := S}` for 408/429/5xx.
- A confirmed-but-non-transient fragment is no longer abandoned (data-loss safety): `transfer_failed/3` returns `resubmit_transfer_delayed` and the pipeline stalls at that offset rather than skipping it.

Still outstanding:
- No exponential backoff or cap on either upload-retry path. The transient path resubmits immediately (`{resubmit_transfer, ...}`), and the non-transient path uses a flat `upload_retry_delay_ms` (default 1000ms). The existing `task_retry_delay_constant`/`task_retry_delay_exponent`/`task_retry_delay_max_ms` backoff machinery is not wired into either path.
- No stalled-upload observability metric. A wedged pipeline surfaces only as per-retry WARNING logs plus the generic `transfers_failed` counter; there is no distinct "stalled at offset" gauge or fatal-retry counter.

Narrowing scope to those two items.

the-mikedavis:
Fixed by https://github.com/amazon-mq/rabbitmq-stream-s3/pull/315

PR #315, opened 2026-06-30
Add exponential backoff and stalled-upload observability to the retry path (#212)

lukebakken:
Addresses the two items left outstanding in #212 after the earlier error-classification and data-loss-safety work shipped:

- no exponential backoff or cap on either upload-retry path
- no distinct signal when the pipeline is wedged on a fragment

## Backoff

Both retry paths now back off exponentially with equal jitter, on separate profiles:

- **Transient errors** (throttling, 5xx, timeouts, connection errors) keep an immediate first retry, preserving responsiveness for a one-off blip, then back off from `task_retry_delay_constant` (10ms) by `task_retry_delay_exponent` (2) up to `task_retry_delay_max_ms` (5s). This wires in config keys that existed but were previously read nowhere.
- **Non-transient errors** (a confirmed checksum mismatch, an unexpected 4xx), where the pipeline intentionally stalls at the failed offset until the fragment is durable (the #206 safety), back off from `upload_retry_delay_ms` (1000ms) up to a new `upload_retry_delay_max_ms` (30s).

Equal jitter (delay uniformly distributed in `[delay/2, delay]`) spreads retries that would otherwise fire in lockstep across many streams after a shared incident.

A per-fragment attempt counter in the functional core drives the backoff; it survives resubmits (the Ref is stable) and is cleared once the fragment is durable or on reinit. The pure `backoff_delay/4` and `equal_jitter/1` helpers live in `rabbitmq_stream_s3_util` so the core stays deterministic and the formula is unit tested; jitter is applied in the shell.

## Observability

- `rabbitmq_stream_s3_transfer_retries` counter: all retries scheduled (transient and non-transient)
- `rabbitmq_stream_s3_nontransient_transfer_retries` counter: retries for a confirmed-fatal error (the pipeline stalls at that offset)
- `rabbitmq_stream_s3_upload_stalled_offset` gauge: the head-of-line offset the pipeline is stalled at, or 0 when healthy. It tracks the lowest stalled offset, so an out-of-order completion of a later fragment does not prematurely clear it, and it clears on recovery.

The WARNING log on a non-transient failure now includes the attempt number.

## Why no `rabbit_alarm`

Consistent with the bucket-accessibility work in #314: the only alarm primitive (`resource_limit`) flow-blocks all publishers on the node, which is the wrong response to a tiering-side stall. The loud-but-safe signal here is the metrics plus the WARNING log.

## Docs

`operations.md` gains the new metrics and a stalled-pipeline alerting rule; `failure-modes.md` gains an "Upload retry strategy" section describing the two profiles, jitter, and detection.

## Config

`upload_retry_delay_max_ms` (default 30000) is added. Like the other retry-tuning keys (`upload_retry_delay_ms`, `task_retry_delay_*`), it is an app-env-only advanced knob with no cuttlefish mapping, matching the existing convention.

## Testing

- eunit: full suite (170), the pure backoff/jitter helpers (growth, cap, base-above-cap, jitter bounds), the core attempt-counter lifecycle (escalates across failures, resets on durability and across fragments), and the shell stalled-offset gauge (head-of-line tracking, no premature clear on out-of-order completion)
- CT: full `broker_SUITE`, `replica_reader`, `replica_reader_core`, `replica_reader_tasks`
- `dialyze` and `xref` clean

### DIFF
--- a/docs/failure-modes.md
+++ b/docs/failure-modes.md
@@ -27,6 +27,20 @@ For the concepts behind these scenarios, see [concepts.md](./concepts.md). For t
 
 ---
 
+## Upload retry strategy
+
+A failed fragment upload is always retried; the fragment is never abandoned, because dropping it would advance the manifest over a range that is not durable in S3 and leave a silent hole (issue #206). How the retry is paced depends on the error class.
+
+**Transient errors** (throttling, 5xx, timeouts, connection errors) are expected to clear quickly. The first retry is immediate, preserving responsiveness for a one-off blip. Successive consecutive failures of the same fragment back off exponentially from `stream_s3.task_retry_delay_constant` (10ms) by a factor of `stream_s3.task_retry_delay_exponent` (2), capped at `stream_s3.task_retry_delay_max_ms` (5s).
+
+**Non-transient errors** (a confirmed checksum mismatch, an unexpected 4xx) are unlikely to clear on a tight retry, so the pipeline stalls at the failed offset and retries with a backoff starting at `upload_retry_delay_ms` (1000ms), growing to `upload_retry_delay_max_ms` (30s). Local-tier cleanup also stalls at that offset, so the only durable copy is retained until the upload succeeds.
+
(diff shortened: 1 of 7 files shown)


the-mikedavis (approved):
Looks great 💪

PR #316, opened 2026-07-01
Write anchors to the DB in order to identify dangling objects

### LINKED ISSUE #196: Stream deletion leaves one dangling S3 fragment from in-flight upload

the-mikedavis:
*Issue #, if available:* fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/196 from a behavior / GC aspect. But a full fix of that issue would also need https://github.com/amazon-mq/rabbitmq-stream-s3/issues/309

*Description of changes:* This changes the Khepri layout a bit to have the manifest pointer under one branch for the stream and an anchor node under the other. The anchor gives us strong ordering that we can use to nearly guarantee that we can correctly identify garbage. This is the implementation of the design in https://github.com/amazon-mq/rabbitmq-stream-s3/pull/311.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -155,7 +155,7 @@ Each stream on the writer node has exactly one replica reader. Its lifecycle:
 2. **Resolve manifest.** Fetches the current manifest state to determine where to resume uploading. If S3 is unreachable, retries with backoff. The writer is unaffected.
 3. **Open data reader.** Calls `osiris_writer:init_data_reader/3` starting at the manifest's `next_offset`.
 4. **Drain loop.** On each `{osiris_offset, ...}` notification, reads committed chunk headers via `osiris_log:read_header/1`. Accumulates metadata in the fragment assembly.
-5. **Cut.** When the assembly reaches the size target (default 64 MiB), the fragment is sealed. The core assigns a reference and returns a `{submit_transfer, ...}` effect.
+5. **Cut.** When the assembly reaches the size target (default 64 MiB), the fragment is sealed. The core assigns a reference and returns a `{submit_transfer, ...}` effect. On a brand-new stream the very first cut instead returns a `{write_anchor, ...}` effect and holds the fragment's submit: the shell writes the per-stream anchor to Khepri, and the core releases the held submits only once the anchor is durable, so no S3 object can exist under a prefix whose anchor is absent (this is what makes deleted-stream GC safe; see [operations.md](./operations.md#safety-guarantee-the-anchor)). A reader resuming an established stream already has the anchor and skips this step.
 6. **Submit.** The shell submits the transfer to the governor. The drain loop continues immediately without waiting for the upload to complete.
 7. **Transfer complete.** The governor reports completion. The core applies the completion in cut order (buffering out-of-order arrivals). When a contiguous prefix of completions is ready, they are applied to the in-memory manifest.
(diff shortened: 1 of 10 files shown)


lukebakken (approved):
Approving after a full review of all six changed files plus local verification.

The anchor-before-first-fragment design is implemented cleanly and the load-bearing invariant holds: the core gates `submit_transfer` behind the anchor write, and since `since_persist` only advances on drain (never at cut time), holding fragment uploads holds the whole pipeline, so no manifest or group object can reach S3 before the anchor commits. Gating fragments alone is sufficient.

Highlights:
- Fail-closed is consistent throughout: both `anchor_absent/1` and `fresh_enough_to_delete/2` treat a non-quorum read as do-not-reap.
- Guard 3 (`fresh_enough_to_delete/2`) closes the reset-after-snapshot / stale-high-cache window for offset-based deletions, and correctly exempts `stale_epoch` and `no_anchor`.
- `initial_anchor/1` starts `done` for a non-empty resolved manifest, keeping the reader-restart path off an extra Khepri round-trip.
- Strong test coverage, including that each stream's anchor is read exactly once per page and the pure `cache_at_committed_epoch/2` decision across all four states.

Verification (locally, on the branch): eunit 42/42, ct-replica_reader_core, ct-db (including the new anchor cases), dialyze, and xref all green.

Docs were updated on this branch (commit documenting the anchor-based GC of deleted-stream objects) to reflect the new no_anchor reclamation path.

Note on the Khepri layout change (leaf -> container with manifest/anchor children): get/list/count now read the manifest child, so a pre-upgrade pointer would read as not_found after an in-place upgrade. Confirmed a non-issue here since the plugin is unreleased and has no deployed Khepri state; flagging only for the record.

PR #322, opened 2026-07-06
api aws pool: Make the pool testable without a real S3 endpoint

the-mikedavis:
open/0 called gun:open/3 directly with no injection point, so the pool's state machine (checkouts/checkouts_rev kept in lockstep, the #177 caller-death handling, connection replacement) could only be regression-tested against a real S3 endpoint. Adds open_fun/usable_fun/ close_fun config keys, each defaulting to today's real behaviour, so tests can drive the pool with plain Erlang processes standing in for gun connections. usable_fun and close_fun are needed alongside open_fun because usable/1 calls gun:info/1 (a sys:get_state/1 call a fake process can't answer) and gun:close/1 silently no-ops on a pid outside its own supervision tree.

Adds a proper_statem suite driving the real gen_server through randomized concurrent checkout/checkin/kill sequences, checking the checkouts/checkouts_rev mirror invariant, the #177 regression (a checked-out connection whose caller died must never be reused), a crashed connection being replaced, and pending-queue handling under caller death.

See: https://github.com/amazon-mq/rabbitmq-stream-s3/issues/178

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ include ../../erlang.mk
 # CT suites share a node name (erlang.mk limitation), so they cannot run in parallel.
 .NOTPARALLEL:
 
-CT_QUICK_SUITES = api_fs db replica_reader_core remote_reader_core replica_reader log_reader fragment_iterator fragment_assembly prop manifest_replica manifest_replica_statem
+CT_QUICK_SUITES = api_fs db replica_reader_core remote_reader_core replica_reader log_reader fragment_iterator fragment_assembly prop manifest_replica manifest_replica_statem api_aws_pool_statem
 
 ERLFMT ?= erlfmt
 ERLFMT_FILES = src/*.erl test/*.erl
--- a/src/rabbitmq_stream_s3_api_aws_pool.erl
+++ b/src/rabbitmq_stream_s3_api_aws_pool.erl
@@ -65,7 +65,13 @@ the reader pool avoids reader starvation.
     %% Native monotonic timestamp at which each connection was opened; used
     %% to report age when a connection goes down unexpectedly.
     created = #{} :: #{conn() => integer()},
-    counter :: counters:counters_ref()
+    counter :: counters:counters_ref(),
+    %% Injection points so this pool can be driven in tests without a real S3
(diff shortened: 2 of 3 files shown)


lukebakken (approved):
Reviewed end to end, verified against source and locally. A clean dependency-injection refactor (open_fun/usable_fun/close_fun config keys, each defaulting to the real open/0, usable/1, gun:close/1) plus a thorough proper_statem suite that makes the pool's #177 state-machine invariants testable without a live S3 endpoint.

Runtime:
- Injection coverage is complete: all six runtime call sites route through the funs (OpenFun in grow/2; UsableFun in take_available/checkout; CloseFun in the handle_gun_up stale-close, terminate, and close_connection). The only remaining direct gun:* calls are inside open/0 and usable/1 themselves, i.e. the defaults. No production path bypasses injection.
- Defaults preserve behavior: maps:get(open_fun, Config, fun open/0) etc., so an existing 3-key config is byte-for-byte unchanged.
- test_inspect/1 is -ifdef(TEST)-gated, absent from release builds.

Tests: models the real gen_server through randomized concurrent checkout/checkin/kill sequences with fake processes for gun connections (callers spawned out-of-process, correct since checkout/2 is a single blocking call). The four properties map exactly to the #177/#178 surface: (1) checkouts/checkouts_rev bijective mirror + every available/checked-out conn monitored, checked after every command (field types conn()=>MRef vs MRef=>conn() confirm the mirror asserts a genuine inverse); (2) #177 caller-death drops the conn and grows a replacement; (3) conn crash untracked and replaced; (4) pending-queue serve-on-free and no leak on pending-caller death. fake_open's self() ! {gun_up,...} faithfully mimics gun's async up-signal since open_fun runs synchronously in the pool process. Idle-timeout (#5) explicitly scoped out with rationale.

Local verification, all green: ct-api_aws_pool_statem (exit 0), dialyze (passed successfully), xref (exit 0). Note: api_aws_pool unresolved calls went 1->7, fully explained by the 6 new fun-applications (xref cannot resolve calls through variable-bound funs); no genuinely undefined functions. No blocking issues.

PR #323, opened 2026-07-06
manifest replica: Gate apply_edits on a live reader context

the-mikedavis:
This is basically the same as https://github.com/amazon-mq/rabbitmq-stream-s3/pull/318 but covers applying edits rather than syncing. The related P module is updated to add edits, showing that the same gate is effective for both.

This also adds a `proper_statem` for the manifest replica gen_server, holding the same properties as the P models.

### DIFF
--- a/.github/workflows/p-model-check.yaml
+++ b/.github/workflows/p-model-check.yaml
@@ -396,7 +396,7 @@ jobs:
           cd p/manifest-replica-lifecycle
           for tc in tcCleanupGuarded tcStaleSyncGuarded tcReregisterGuarded \
                     tcConvergenceGuarded tcForgetReleasesWriterRow tcSyncAfterExitFixed \
-                    tcStartupRaceContextFirst tcReconcileRaceResync tcExplore; do
+                    tcEditAfterExitFixed tcStartupRaceContextFirst tcReconcileRaceResync tcExplore; do
             echo "::group::$tc (expect hold)"
             p check -tc "$tc" -i 5000
             echo "::endgroup::"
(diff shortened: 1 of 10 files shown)


lukebakken (approved):
Reviewed end to end, verified against source and locally. Closes the manifest-replica leaked-row gap on the apply_edits write path (the sync path was already gated). Both handle_call/handle_cast clauses now check maps:is_key(StreamId, Ctxs) first and drop through a shared drop_no_context/6 (counted as edits_dropped_no_context, recorded in pending_resync for resync-on-register), symmetric with maybe_apply_sync/8.

Correctness:
- The 'unreachable []-branch' claim holds: every seqs-advancing site (4) is paired with a preceding write_manifest, and the only ETS delete is release_stream, which clears context+seqs+ETS atomically. The removed old []-branches were exactly the ones that advanced seqs without an ETS write, so replacing them with a resync is a strict improvement; the new branch is a fail-safe.
- No caller-contract regression: the new {error, no_context}/{error, apply_failed} returns are on the synchronous apply_edits/4 (handle_call), test-only. Production broadcasts via the async apply_edits/5 cast (handle_cast), which discards the reply (replica_reader.erl:1082).
- Counter wiring sound: ?C_EDITS_DROPPED_NO_CONTEXT = 4 matches the 4th ?COUNTERS entry; the seshat array is sized from ?COUNTERS. is_stale_sync/3 is exported for the statem model.

Tests:
- P-model: eMREdit mirrors the code's gate ordering (context check before seq contiguity); tcEditAfterExitStrands (negative control, NOLEAK fires guard-off) and tcEditAfterExitFixed wired into the CI expect-hold loop and gate list with matching messages. The one modeling simplification (contextless-first edit accepts a baseline where real code returns {error, gap}) is documented and scoped to the TLA+ companion, benign for the leak property.
- proper_statem (repo's first): drives the real gen_server through randomized register/kill/sync/apply_edits/forget, checking no-leak, in-order-applies, gap-never-corrupts, async-release-convergence; the real {error, gap} behavior is pinned here.
- Example tests mirror the sync equivalents.

Local verification, all green: ct-manifest_replica_statem, ct-manifest_replica (21 cases incl. the 2 new), dialyze (only pre-existing dep unknowns), xref (unresolved calls all pre-existing in unrelated modules). No blocking issues.

PR #324, opened 2026-07-06
Add a proper_statem suite for replica_reader's upload lifecycle

the-mikedavis:
replica_reader_SUITE pins each recovery seam with one hand-written example at a time (killed_reader_re_registers_and_resumes, upload_path_recovers_from_trimmed_segment, stale_*_is_ignored, ...). None of them combine these events within a single run. This drives a real replica_reader (real osiris_writer, real local Khepri, the existing _api_fs/_api_fault local-filesystem S3 stand-in) through randomized publish/force-cut/kill/retention/trim sequences and checks, after every command: range monotonicity (the writer/epoch fencing CAS is externally observable as the committed range never regressing), no permanent stall after a kill (the per-stream supervisor restarts it and it resumes tiering), and local retention never durably outrunning the persisted floor (issue #206/#225's local_log_ahead recovery).
    
It deliberately does not re-model the async-task correlation logic (persist/group/retention/transfer staleness, generations, deliver-vs-drop): that is already exhaustively property-tested against the pure replica_reader_tasks core in replica_reader_tasks_SUITE, and replica_reader_core is pointwise-property-tested in prop_SUITE.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ include ../../erlang.mk
 # CT suites share a node name (erlang.mk limitation), so they cannot run in parallel.
 .NOTPARALLEL:
 
-CT_QUICK_SUITES = api_fs db replica_reader_core remote_reader_core replica_reader log_reader fragment_iterator fragment_assembly prop manifest_replica manifest_replica_statem api_aws_pool_statem
+CT_QUICK_SUITES = api_fs db replica_reader_core remote_reader_core replica_reader replica_reader_statem log_reader fragment_iterator fragment_assembly prop manifest_replica manifest_replica_statem api_aws_pool_statem
 
 ERLFMT ?= erlfmt
 ERLFMT_FILES = src/*.erl test/*.erl
--- a/test/rabbitmq_stream_s3_api_fault.erl
+++ b/test/rabbitmq_stream_s3_api_fault.erl
@@ -187,20 +187,34 @@ maybe_delay(Op, Key) ->
     end.
 
 maybe_block(Op, Key) ->
-    case lookup({block, Op}) of
-        {ok, {KeyPat, TestPid, Ref}} ->
+    %% ets:take/2 atomically reads and removes the entry in one operation.
+    %% A separate lookup-then-delete (as this used to be) is a check-then-act
+    %% race: when a fragment cut produces several concurrent uploads for the
+    %% same stream (an unlimited-rate governor spawns one task per submission
+    %% immediately), more than one of their stream_put calls can read the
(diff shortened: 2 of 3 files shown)


lukebakken (approved):
Reviewed end to end, verified against source and locally (including a flakiness rerun). Two test-only commits; no production src/ change.

Commit 1 (TOCTOU fix in rabbitmq_stream_s3_api_fault:maybe_block/2): the bug is real. Under an unlimited-rate governor, a fragment cut spawns several concurrent upload tasks for one stream; the old lookup-then-delete let more than one read the same still-armed one-shot block and park, but the caller awaits only a single {fault_blocked,...}, so the rest hang forever as unmonitored governor tasks. Fix verified: ets:take/2 on a set table ([named_table, public, set]) is atomic (exactly one concurrent caller gets the object), closing the double-park; tuple shapes are consistent across the block/2 insert, the take match, and the non-match re-insert; the non-match re-insert correctly restores the entry for a later matching call. The suite also defends the complementary race in do_trigger_trimmed_segment's not_hit branch (drain_late_block/1).

Commit 2 (the statem suite): scope is well-judged - it explicitly does not re-model async-task correlation (covered by replica_reader_tasks_SUITE and prop_SUITE) and targets the combinatorial gap the hand-written examples miss. The three properties are checked at the integration level only this harness can observe: (1) range monotonicity (fencing CAS observable as a never-regressing committed range); (2) no permanent stall after kill - do_kill returns the old pid and kill_ok awaits a *different* registered pid, verifying genuine supervisor restart, with kills capped at 3/run under the intensity-5/10s budget so every kill exercises real resume; (3) local floor never durably outruns the persisted floor (#206/#225), a bounded poll correctly framed as transient-legal, validated by the author to catch a #225 regression when the local_log_ahead guard is disabled.

Verified: all fault-backend (block_once/2, await_blocked/2, release/2, reset/0) and helper (flush_writer/1, await_offset/3, segment_file_offset/1) APIs exist with matching arities; offset semantics right (await_offset(written) maps to committed.next_offset >= written, correct since records occupy 0..written-1, with the under-threshold-tail non-guarantee handled by the force_fragment_cut nudge per poll); do_trigger_trimmed_segment's fixed +100 in next_state is correct for symbolic generation; per-iteration cleanup is thorough and the one intentionally-leaked Khepri container node is documented with sound reasoning.

Local verification: ct-replica_reader_statem exit 0, 'OK: Passed 40 test(s)'; ran 3 more times (flakiness check, since the fixed race was ~1-in-15) all green, 160 iterations total with no flakes; dialyze and xref both green. No blocking issues.

PR #325, opened 2026-07-06
Jepsen: Add member-churn nemesis for the manifest-replica leak

the-mikedavis:
Add a member-churn fault that each tick delete_replicas a random replica from a random stream, so an osiris member departs its node permanently mid-sync. This targets the :replica checker's leaked-replica-row assertion, which leader-move cannot reproduce: leader-move restarts every member in place on the same node set, so each departed member re-registers a context that reclaims any orphan, whereas a deleted member does not return.

delete_replica also restarts the stream's writer (the coordinator sets target = stopped on every running member, not just the removed one), so it drifts the client's send offsets like leader-move. Downgrade the kafka workload checker to advisory under member-churn too, letting the durability checker carry the safety verdict.

Bounded to keep every stream at leader plus at least one replica, so durability holds for the final read, and delete-only by design so an orphan stays visible at the checker's end-of-run snapshot.

Validated green on a fresh five-node cluster under member-churn,s3-latency: the :replica checker reports no divergence, stale floors, or leaked rows, durability is clean, and the sync-context guard was exercised 395 times.

> [!NOTE]
> This is not suitable to add to CI because repeated random delete_replica commands sent to the stream coordinator can get a stream wedged. After some debugging I'm pretty sure it's the liveness issues in https://github.com/rabbitmq/rabbitmq-server/pull/16881


lukebakken (approved):
Reviewed end to end. Test-harness only (+116/-16): adds a `member-churn` Jepsen nemesis that each tick `delete_replica`s one random replica from a random `jepsen-*` stream, bounded to keep every stream at leader + >=1 replica. This is the one fault that can positively drive the `:replica` checker's `:leaked-replica-row` assertion, since a deleted member leaves its node for good (no re-registration to reclaim the sync-raced cache row) where `leader-move` restarts members in place.

Verified against the server tree:
- `rabbit_stream_queue:delete_replica/3` signature (VHost, Name, Node) matches
- `amqqueue:get_type_state/1`/`get_name/1` exist; `replica_nodes` is a valid type-state key
- `element(4, ...)` on the resource record = name ({resource, virtual_host, kind, name})
- Fail-safe: a queue deleted between list/1 and lookup/1 yields not_found -> get_type_state fallback #{} -> replica_nodes [] -> stream filtered out, no crash
- Bound is correct: replica_nodes is followers only, candidate requires >=2, so a delete always leaves leader + >=1 follower (durability holds for the final read)
- BACKLOG/README citations verify verbatim: restart_stream/2, update_stream0/3, initial_cluster_size(undefined) -> length(rabbit_nodes:list_members()), and version() -> 7 (so the #16881 fix at machine version 8 is genuinely absent and the 'keep out of CI' hazard is real)

Wiring is consistent (full-nemesis, fault->bg-op, known-faults, validate-faults!); core.clj's downgrade-when now covers both leader-move and member-churn with documented reasoning. Confirmed member-churn is not in the jepsen.yaml CI matrix. Docs are candid about the limits (fires only against a fix-removed negative control; unbounded pending_resync follow-up; writer-liveness wedge attributed to #16822/#16881). No changes needed.

Issue #326, opened 2026-07-13
Flaky replication_survives_reader_restart: core rebuild drops await_offset waiters, plus a too-tight barrier timeout


lukebakken:
## Summary

`replica_reader_SUITE:replication_survives_reader_restart` fails intermittently under full-suite load. Investigation showed this is not a single flake but two independent problems, and it surfaced a third, still-open flake in a neighbouring case:

1. A real bug in the replica reader: a core rebuild discards `await_offset` waiters without replying, so a blocked caller hangs to its own timeout even though the reader is healthy and advancing. This is the dominant cause of the restart-case failure.
2. A genuinely timing-sensitive barrier: the `await_offset/1,2` default deadline was 1000ms, which is occasionally too tight under full-suite CPU contention.

Both are addressed in 9ff7ed3. A third, unrelated flake (see "Still open" below) was observed during the investigation and is not fixed.

## Root cause of the restart-case hang (the real bug)

Every path that rebuilds the reader's core, `resolve_and_start/1` (manifest reinitialize after a commit conflict, and the retry-resolve path) and `restart_at_local_floor/3` (local-log-ahead / remote-ahead reset), replaces the whole state via `rabbitmq_stream_s3_replica_reader_core:init/2`, which starts with an empty `waiters` list. A caller already blocked in `await_offset` at that moment was silently dropped with no reply and blocked until its own `gen_server:call` timeout.

This was confirmed with reader-state captures at the point of timeout across several reproductions:

- The resolved reader pid was identical and alive at call time and at timeout (`same=true`), so this is not a registry or restart race.
- At timeout the reader showed `manifest_next_offset=30`, `persisted_next_offset=30` (the barrier target was 15), and `waiters=0`, with the advancing persist having completed roughly one second into the wait.

The only way `waiters` returns to empty without a reply being sent is a core rebuild via `init/2`. The restart test exercises exactly the reinitialize path, so it hit this reliably enough to flake.

`await_offset` has no production caller today (it is a test barrier used by the CT helper and `broker_SUITE`), so this is not a production data-loss path. It is still a latent reader-robustness bug: any future `await_offset` user would be exposed, and dropping a waiter on rebuild contradicts the documented recovery intent (see `docs/failure-modes.md`, which lists "await-offset waiters never return" as the failure recovery exists to prevent).

## Fix

9ff7ed3:

[...]

Issue #79, opened 2026-03-23
Improve deletion of remote tier data of large streams


the-mikedavis:
When deleting stream which has a lot of data in the remote tier, the process spawned in `rabbitmq_stream_s3_server:execute_task/2` can run for much longer than the other tasks. It deletes the entire stream by requesting 1000 object keys from S3 under the prefix for the stream's ID and then issuing a DeleteObjects request to delete those keys. Large streams may have millions of objects so this process can take thousands of round trips and many minutes to complete.

The ad-hoc processes the server uses for tasks like uploads work well since they are short-lived. Since deletion can be long-running we should investigate spawning a gen_server to perform deletion instead. The task process' mailbox gradually gets longer and longer since it is only consumed with `gun:await/4` selective receives, and that probably slows the process down.

lukebakken:
Closing as addressed. Remote-tier deletion of large streams is now handled the way this issue asked: a long-lived, supervised `gen_batch_server` reaper (`rabbitmq_stream_s3_reaper`, supervised in `rabbitmq_stream_s3_sup`) batches `DeleteObjects` up to 1000 keys per request (`?MAX_BATCH`), fed by a short-lived task that pages through LIST and casts discovered keys back to the reaper (`delete_stream/1` -> `list_and_delete_loop/2`). The old long-running-task design with the growing selective-receive mailbox (the `rabbitmq_stream_s3_server:execute_task/2` path) no longer exists. Implemented in commit eb41832, hardened by 38aef16 and 3da765a, with orphan GC (`rabbitmq_stream_s3_gc`) as a backstop. Nothing in the issue text remains outstanding.

the-mikedavis:
Actually there's one more hacky thing to improve: https://github.com/amazon-mq/rabbitmq-stream-s3/blob/3607fcc87ce7305d3f5a7f62823881d1658dc2a3/src/rabbitmq_stream_s3_reaper.erl#L60-L67

We should have a gen_server in charge of listing keys and sending them to the reaper. The listing part of stream deletion has the same mailbox growth issue.

the-mikedavis:
Fixed in #327

PR #327, opened 2026-07-13
Move stream-deletion listing into a backpressured lister gen_server

lukebakken:
## What

Moves whole-stream remote tier deletion off the reaper's spawn-based listing task and into a dedicated, supervised `rabbitmq_stream_s3_lister` gen_server that paces LIST against DELETE under backpressure, then hardens that lister against reaper faults and duplicate deletion jobs.

Fixes the listing follow-up in #79.

## Why

Stream deletion used to `spawn` an unsupervised task that listed a stream's S3 prefix as fast as S3 answered and cast every page to the reaper. Listing is cheap while deletion is the slow, throttling-prone side, so listing could outrun deletion and regrow the reaper's mailbox without bound, the exact problem the reaper rework had fixed, just moved to the listing side.

## How

Commit 54f8a2b introduces the lister:

- A supervised singleton gen_server serializes stream deletions FIFO and lists one page at a time
- Each page is handed to the reaper via a new synchronous `delete_objects_sync/2` that returns only once the page has been deleted, so the next LIST happens only after the previous page drains (the backpressure). Keys in flight are bounded to a single page
- The `streams_deleted` counter moves from the reaper to the lister; the exported metric name is unchanged
- Supervisor, `db.erl`, API init, and two test call sites are repointed; `architecture.md` and `operations.md` updated

Commit 93c57b8 hardens the lister against three robustness gaps found in review, all in its synchronous coupling to the reaper:

- The backpressure call waited `?DELETE_TIMEOUT_MS` (60s), but that call waits for the reaper's whole `handle_batch`, which folds in unrelated fire-and-forget retention casts and splits into sequential 60s sub-batches. Under S3 throttling the batch could exceed the caller's deadline and crash the lister while deletion was still progressing. The call now waits `infinity`; each S3 request stays individually bounded, so `handle_batch` always completes in bounded time
- A reaper crash or restart made the synchronous call exit with `noproc`. The lister held its entire FIFO queue in process state, so any such exit restarted it with an empty queue, silently dropping every pending deletion. The reaper call is now wrapped so a reaper fault abandons only the current stream to orphan GC and leaves the lister and its queue intact
- The Khepri deletion trigger can fire more than once for a stream, and the lister did not de-duplicate, so a stream could be listed twice and double-count `streams_deleted`. A stream already in flight or already queued is now skipped

One review finding was left unaddressed by design: if `delete_stream` is cast while the lister is restarting, the cast is silently dropped. This is the intended failure mode, backstopped by orphan GC (see architecture.md), and hardening it would duplicate the GC mechanism as a fragile special case.

## Testing

- `gmake dialyze`, `gmake xref`: clean (only pre-existing unresolved calls in untouched modules)
- `gmake eunit`: 192 passed, including new coverage for the de-duplication logic
- `gmake ct-replica_reader`, `ct-replica_reader_statem`, `ct-broker`: deletion cases (`stream_deletion_cleans_remote_tier`, `stream_deletion_during_active_upload`) green in every run
- `erlfmt -c`: all files conform

The pre-existing flaky `replication_survives_reader_restart` is tracked separately in #326 and is unrelated to this change.

### DIFF
--- a/src/rabbitmq_stream_s3_lister.erl
+++ b/src/rabbitmq_stream_s3_lister.erl
@@ -123,8 +123,6 @@ maybe_start(#state{} = State) ->
 %% backpressure: the next list happens only after the previous page drained.
 page(StreamId, Prefix, Continuation, State) ->
     case rabbitmq_stream_s3_api:list(Prefix, Continuation) of
-        {ok, [], done} ->
-            finish(StreamId, State);
         {ok, Keys, done} ->
             case delete_page(StreamId, Keys) of
                 ok -> finish(StreamId, State);
--- a/src/rabbitmq_stream_s3_lister.erl
+++ b/src/rabbitmq_stream_s3_lister.erl
@@ -67,8 +67,8 @@ init([]) ->
 handle_call(_Request, _From, State) ->
     {reply, ok, State}.
 
-handle_cast({delete_stream, StreamId}, #state{queue = Q} = State) ->
-    {noreply, maybe_start(State#state{queue = queue:in(StreamId, Q)})};
+handle_cast({delete_stream, StreamId}, State) ->
+    {noreply, enqueue(StreamId, State)};
 handle_cast(_Msg, State) ->
     {noreply, State}.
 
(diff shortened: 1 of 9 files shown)


src/rabbitmq_stream_s3_lister.erl
```diff
@@ -0,0 +1,224 @@
...
+    State.
+
+%% List one page and hand it to the reaper synchronously (the call returns only
+%% once the page has been deleted), then schedule the next page. This is the
+%% backpressure: the next list happens only after the previous page drained.
+page(StreamId, Prefix, Continuation, State) ->
+    case rabbitmq_stream_s3_api:list(Prefix, Continuation) of
+        {ok, [], done} ->
+            finish(StreamId, State);
+        {ok, Keys, done} ->
```
the-mikedavis, line 126: `delete_page/2` has a short-circuit `[] -> ok` (in the reaper) so this branch can be covered by the branch below

the-mikedavis:
I'm a bit hesitant to add the synchronous part here since I'm not sure it's necessary. I wonder if we really do need backpressure. Let me dig into it a little bit. I suspect that backpressure is correct anyways, since I could see a degenerate case here where you delete TiB+ of data across many streams concurrently, but I wonder how expensive that would actually be for the reaper to handle.

the-mikedavis (approved):
Thinking about it more, the backpressure might be nice here, to avoid exploding the memory of the reaper.

Overall I think my main concern with the listing API is that it used to put a bunch of junk in the caller's mailbox. I don't think that's actually true anymore now that we have the connections in pools. So the backpressure here (plus the new metric) are the improvements rather than avoiding mailbox junk

PR #328, opened 2026-07-13
Add automatic periodic tiered storage GC

### LINKED ISSUE #309: Trigger garbage collection periodically and automatically

lukebakken:
## What

Adds `rabbitmq_stream_s3_gc_scheduler`, a periodic `gen_server` that triggers a delete-mode cross-stream GC sweep on a timer. A sweep must run on only one node at a time, so it is elected each round by a non-blocking cluster-wide lock (`global:trans/4` with zero retries) with no fixed leader, so the sweep survives the loss of any node.

Closes #309.

## Why

`rabbitmq_stream_s3_gc:run/1` was otherwise only invoked on demand via the `stream_s3_gc` CLI command (and the per-stream reset path). Without a periodic trigger, a straggler object left by the deletion race (an upload that completes after the one-shot deletion sweep) and its tombstone persist until an operator runs GC by hand. This backstops that gap and makes straggler reclamation self-healing.

## Configuration

Automatic GC is off by default. When enabled it runs on a conservative 24 hour interval: a full sweep is a bucket-wide `LIST` plus one strongly-consistent metadata read per stream, and the objects it reclaims are not urgent.

```ini
# boolean, default false
stream_s3.gc.enabled = true

# positive integer milliseconds, default 86400000 (24 hours)
stream_s3.gc.interval = 86400000
```

## Notes on reclamation scope

Reclamation is eventual, not immediate. Deleted-stream orphans are identified from Khepri via a strongly-consistent anchor read, so any sweeping node reclaims them. An object orphaned below a live stream's first offset is recognised only against the sweeping node's local manifest replica cache, so those are reclaimed over successive rounds as each stream's caching nodes win the lock. The moduledoc spells this out.

## Testing

- New multi-node `gc_scheduler_SUITE` exercises the cluster-wide lock's mutual exclusion (the eunit VM runs without distribution, so `global` does not arbitrate there)
- eunit gating tests for enabled/disabled init
- `config_schema_SUITE` snippets for the two new keys
- `eunit`, `ct-gc_scheduler`, `ct-config_schema`, `xref`, `dialyze` all green

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -115,6 +115,22 @@ stream_s3.bucket_check.enabled = true
 stream_s3.bucket_check.interval = 300000
 ```
 
+### Automatic garbage collection
+
+Garbage collection reclaims dangling S3 objects (see [Garbage collection](#garbage-collection)). It can run automatically on an interval in addition to the on-demand `stream_s3_gc` CLI command. Automatic GC is off by default: a full sweep is a bucket-wide `LIST` plus one strongly-consistent metadata read per stream, and the objects it reclaims (a straggler left by the deletion race and its tombstone) are not urgent.
+
+Each node runs its own timer, but a non-blocking cluster-wide lock ensures only one node sweeps per round, so a larger cluster does not multiply the sweep cost. There is no fixed leader; the sweep survives the loss of any node.
+
+```ini
+# Whether a delete-mode GC sweep runs automatically on an interval.
+# Type: boolean. Default: false.
+stream_s3.gc.enabled = true
+
+# Interval between automatic GC sweeps, in milliseconds.
+# Type: positive integer. Default: 86400000 (24 hours).
+stream_s3.gc.interval = 86400000
+```
+
 ### Tuning the upload path
 
 ```ini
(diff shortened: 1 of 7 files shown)


the-mikedavis (approved):
Looks great to me. I added another config key `stream_s3.gc.mode = dry_run | delete` (default `delete`) so that we can try GC automatically and spot-check it before enabling the destructive mode.

Issue #298, opened 2026-06-26
Validate remote-read memory on representative instance sizes; decide whether fan-out can threaten node availability


lukebakken:
## Background

#176 investigated per-node memory peaks during heavy remote-tier reads and reached a clear conclusion: the memory is the remote readers' AIMD prefetch buffers (refc binaries), each capping at ~130-140 MB (a ~64 MiB current buffer plus a ~64 MiB prefetched-next), it is fully reclaimed once consumers disconnect (no leak), and it scales linearly with the number of remote readers reading concurrently on a node, with no per-node ceiling: `peak ~= (concurrent remote readers) x ~140 MB`. A re-run also showed that forcing aggressive GC on the reader (`fullsweep_after = 0`) does not reduce the sustained-read footprint, because the buffers are actively live.

That investigation ran on large test nodes (`m7g.8xlarge`, ~128 GB), where the peak was harmless and no memory alarm fired. This issue is the follow-up: confirm whether the same behavior is benign on the instance sizes that actually matter, and decide whether a mitigation is warranted.

## The risk to validate

The peak scales with a quantity the broker operator does not control: how many consumers happen to read the remote tier concurrently on one node. Nothing bounds that count today. Plausible ways it spikes:

- A consumer fleet (re)starting from `first` or an old offset at once (deploy, restart, bulk replay) produces many cold remote readers on the leader node simultaneously.
- Many streams on one node, each with a lagging consumer.

The concern is an availability one, not a leak: if remote-read buffers push a node into the memory resource alarm, RabbitMQ blocks publishers cluster-wide. That would turn a read-side load pattern into a write-side outage. The exposure is far sharper on smaller instances than the test rig: the same handful of readers that were invisible at a few GB on a 128 GB node could trip the alarm on a small broker.

## Questions to answer

1. On representative / managed instance sizes (not the 128 GB test rig), how many concurrent remote readers does it take to approach the memory alarm threshold, given ~140 MB per reader?
2. Can realistic consumer behavior actually produce that many concurrent remote readers on a single node? Is there anything upstream (connection limits, consumer scheduling, how cold replays are distributed) that already bounds the fan-out?
3. If the alarm can be tripped: does it in fact block publishers cluster-wide as expected, and how disruptive is recovery?

## Mitigation (only if the validation shows a real risk)

Do not pre-commit to a specific mechanism. If fan-out on representative instances can threaten availability, the design space includes (not exclusive):

[...]

Issue #176, opened 2026-05-28
Investigate per-node memory peaks during heavy remote tier reads


lukebakken:
## Summary

In a 2-hour high-throughput soak test on the combined branch of PR #169 and PR #171, peak per-node memory grew 2.7x to 5.3x relative to start-of-test. Post-test memory returned to a normal baseline. We do not yet have evidence to distinguish "expected buffer growth under heavy S3 reads" from "leak in the read path." This issue tracks the investigation.

## Observations

From the test runner's snapshot harness:

| Node | Start memory | Peak memory | Ratio |
|---|---|---|---|
| `rabbit@ip-10-0-1-162` | 263 MB | 720 MB | 2.7x |
| `rabbit@ip-10-0-1-21` | 270 MB | 1348 MB | 5.0x |
| `rabbit@ip-10-0-1-12` | 262 MB | 1385 MB | 5.3x |

After the test finished, current Erlang VM memory on each node is around 80-85 MB total (`erlang_vm_memory_bytes{kind=processes}` ~27 MB plus `kind=system` ~56 MB). So peaks were genuinely transient during heavy reads, not residual.

No memory alarms fired during the run. The test ran 6 streams with 12 producers and 12 consumers; consumers were reading aggressively from the remote tier under S3 throttling (120/363/72 `S3 request timed out` warnings, many 10-16s `Slow remote tier read` warnings).

## Plausible sources

The remote reader uses an AIMD-controlled prefetch window. From `rabbitmq_stream_s3_remote_reader_core:build_cfg/1`:

```
read_size_min = 1_048_576           %% 1 MiB
read_size_max = 67_108_864          %% 64 MiB
initial_read_size = 4_194_304       %% 4 MiB
```

Each remote reader holds a buffer up to `read_size` bytes for the current fragment plus a pre-fetched buffer for the next fragment. With `read_size` ramping toward 64 MiB on consecutive hits, a single reader can hold over 100 MiB at peak. The post-test counters show 8/6/5 remote readers per node having been initialized (some of which transitioned to local), so per-reader peaks of 100+ MiB are sufficient to explain a ~1 GiB peak with no leak.

The `read_size_bytes` histogram that would let us validate this distribution is one of the dead metrics tracked in #175. Without it we cannot confirm that AIMD actually drove the buffer up to 64 MiB.

[...]

PR #330, opened 2026-07-14
Optimize remote reader binary handling

the-mikedavis:
*Issue #, if available:* related to https://github.com/amazon-mq/rabbitmq-stream-s3/issues/176, https://github.com/amazon-mq/rabbitmq-stream-s3/issues/298

*Description of changes:* This change swaps out the remote reader's buffer type. It was a flat binary and newly prefetched data was appended to it (`<<Buffer/binary, NewData/binary>>`. This change adds a queue which holds the prefeteched binaries in order, without modifying/appending any binaries.

Appending a binary is fairly cheap and `binary:part/3` is fairly cheap too, but together with sending the binaries as messages, they pin binaries so that they cannot be reclaimed quickly, and produce extra copies of the binaries. This is very noticeable in the log reader at high throughput. Instead we want to avoid modifying the binaries that we get from the HTTP client at all, so we avoid making extra copies of the data. The investigation doc has all of the details on why the old mechanism was expensive, and data on how quick and memory efficient the new block queue type is.

### DIFF
--- a/src/rabbitmq_stream_s3_read_buffer.erl
+++ b/src/rabbitmq_stream_s3_read_buffer.erl
@@ -161,6 +161,12 @@ read_iodata(Pos, Len, #buffer{start_pos = StartPos, end_pos = EndPos}) ->
 %% into it, so a read near start_pos never pays for the whole block list.
 take(_Skip, 0, _Front, _Rear) ->
     [];
+take(_Skip, _Len, [], []) ->
+    %% No blocks remain but Len > 0 (the clause above consumed Len == 0). This
+    %% is unreachable while read_iodata/3's range check guards every call, but
+    %% crashing here turns a broken invariant into a loud error rather than an
+    %% infinite loop: the clause below would otherwise spin on lists:reverse([]).
+    error(buffer_underrun);
 take(Skip, Len, [], Rear) ->
     take(Skip, Len, lists:reverse(Rear), []);
 take(Skip, Len, [Block | Blocks], Rear) when Skip >= byte_size(Block) ->
--- a/src/rabbitmq_stream_s3_read_buffer.erl
+++ b/src/rabbitmq_stream_s3_read_buffer.erl
@@ -1,0 +1,196 @@
+%% Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+%% SPDX-License-Identifier: Apache-2.0
+
+-module(rabbitmq_stream_s3_read_buffer).
+-moduledoc """
+A contiguous window of stream bytes stored as a queue of immutable binary
+blocks.
+
+The remote read path buffers data delivered by S3 range requests and serves
+the log reader's reads out of that buffer. Storing the window as a single
+binary and growing it with `<<Buffer/binary, Data/binary>>` interacts badly
+with the runtime once reads are served from it:
+
+- Serving a read with `binary:part/3` and replying with the result seals the
+  underlying writable binary (the runtime pins it when the sub-binary is
+  copied into a message), so the next append copies the entire window rather
+  than appending in place. In steady state nearly every delivery re-copies
+  the whole prefetch window (up to `read_size_max`, 64 MiB).
+- A reply sub-binary keeps the entire window generation alive in the
+  consumer's heap: a ~300-byte chunk-header read can pin 64 MiB until the
+  consumer process happens to collect garbage.
+
+This module avoids both by never appending into a binary. Deliveries are
+retained as-is, in order, as sealed blocks; consumed blocks are dropped
+whole; reads assemble exactly the requested bytes from the blocks they
+overlap. Appending is O(1), dropping frees block-by-block, and a reply pins
+at most the blocks it overlaps.
+
+Addressing is absolute: positions are byte offsets within the fragment
+object, the same coordinates the remote reader core uses. The buffer covers
+`[start_pos, end_pos)` and maintains the invariant that blocks are
+contiguous: `end_pos - start_pos` always equals the sum of the block sizes.
+
+Reads of at most 512 bytes are returned as fresh copies rather than
+sub-binaries, so the frequent small chunk-header over-reads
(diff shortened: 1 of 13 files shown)


lukebakken (commented):
No correctness issues found. The buffer rewrite is semantics-preserving, and the two new property tests (`read_buffer_matches_model`, `remote_reader_core_reply_bytes_exact`) pin down byte-exactness under arbitrary append/read/drop interleavings. A few cleanup notes inline.

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -84,13 +89,14 @@ and transitions immediately if available, or signals that more data is needed.
     %% Current fragment
     fragment_ref :: #fragment_ref{},
     key :: rabbitmq_stream_s3:key(),
-    buffer = <<>> :: binary(),
-    start_pos :: byte_offset(),
+    buffer :: rabbitmq_stream_s3_read_buffer:buffer(),
+    %% Start of the most recently served read: the buffer's retention floor.
+    %% Reads are non-decreasing, so blocks entirely below it are dropped when
+    %% a read is served.
     current_pos :: byte_offset(),
```
lukebakken: `current_pos` is now write-only. It is assigned in six places (init, `deadline_expired`, `iterator_refreshed`, `try_read`, and the two `goto_next_fragment` branches) but read nowhere: no guard, no right-hand side, no accessor, and the shell never touches it. Before this change it was the sole input to `add_data_current`'s compaction; the block queue replaced that with `drop_before(Offset, ...)` in `try_read`, driven by the read `Offset` rather than this field.

The comment above also no longer holds: the retention floor is the `Offset` passed to `drop_before`, not this field, so "the buffer's retention floor... blocks entirely below it are dropped when a read is served" is misleading. Suggest deleting the field (and this comment) along with its six assignment sites, the same cleanup already done for `start_pos`/`end_pos`. Not a one-click suggestion since the assignment sites need removing too, or the record references won't compile.

src/rabbitmq_stream_s3_read_buffer.erl
```diff
@@ -0,0 +1,183 @@
...
+""".
+-spec read_iodata(byte_offset(), non_neg_integer(), buffer()) -> [binary()].
+read_iodata(_Pos, 0, _Buffer) ->
+    [];
+read_iodata(Pos, Len, #buffer{start_pos = StartPos, end_pos = EndPos} = Buffer) when
+    is_integer(Pos) andalso is_integer(Len) andalso Len > 0 andalso
+        Pos >= StartPos andalso Pos + Len =< EndPos
+->
+    #buffer{front = Front, rear = Rear} = Buffer,
+    take(Pos - StartPos, Len, Front ++ lists:reverse(Rear));
```
lukebakken: `take(Pos - StartPos, Len, Front ++ lists:reverse(Rear))` materializes `Front ++ lists:reverse(Rear)` on every read, and the normalized list is never stored back, so `rear` is re-reversed each time. In steady state appends land in `rear` and `drop_before/2` only migrates blocks into `front` when `front` empties, so `rear` holds most of the window (up to ~64 blocks at the AIMD ceiling). The read path issues multiple reads per delivery (the 303-byte header over-read plus a chunk-data read), each paying O(block_count) even when the requested bytes sit in the first block or two near `start_pos`.

Walking `Front` first and only reversing `Rear` if bytes remain would let the common small read never touch `rear`. Not a correctness issue and the benchmark shows it isn't a practical regression, but it is per-read O(n) work the flat binary didn't have.

src/rabbitmq_stream_s3_read_buffer.erl
```diff
@@ -0,0 +1,183 @@
...
+(`?CHUNK_HEADER_B + ?MAX_FILTER_SIZE` = 303 bytes) pin nothing at all.
+""".
+
+-include("include/rabbitmq_stream_s3.hrl").
+
+%% Reads at or below this size are copied out rather than returned as
+%% sub-binaries into a block. Copying a few hundred bytes is cheaper than
+%% letting a short read pin a block (typically ~1 MiB, see
+%% ?BUFFER_PENDING_DATA_BYTES in the S3 API module) in the consumer's heap.
+-define(COPY_MAX_B, 512).
```
lukebakken, line 49: Minor: `512` is unexplained and not tied to the read it exists for. The only small read on this path is the chunk-header over-read of `?CHUNK_HEADER_B + ?MAX_FILTER_SIZE` = 303 bytes, so there is ~209 bytes of headroom with nothing anchoring the two. If `?MAX_FILTER_SIZE` ever grew past this, header reads would silently start pinning a full block again with no compile-time warning. A one-line note documenting the intent:

```suggestion
%% Reads at or below this size are copied out rather than returned as
%% sub-binaries into a block. Copying a few hundred bytes is cheaper than
%% letting a short read pin a block (typically ~1 MiB, see
%% ?BUFFER_PENDING_DATA_BYTES in the S3 API module) in the consumer's heap.
%% Sized to cover the chunk-header over-read (?CHUNK_HEADER_B +
%% ?MAX_FILTER_SIZE = 303 bytes), the only small read on this path, with
%% headroom; if that over-read ever grows past this, header reads pin a
%% block again.
-define(COPY_MAX_B, 512).
```

lukebakken (approved):

the-mikedavis:
This PR introduced micro benchmarks but I wanted to gather data from end-to-end consumption as well. I published 1_000_000 messages at 2_000 msg/s with size 10_000 bytes via stream-perf-test to a 3-node cluster on latest main (including this patch) 7e3cf0ab670e115ec357e3acdfd293ad0b6c66cd, and on another broker on the 1.0 beta.1 tag. Then I consumed the stream with no rate limit from first offset. These tests were against `m7g.xlarge` EC2 instances.

The beta build had a mean of 1_720.5 MB where main had 120.3 MB allocated during consumption - around 14.3x less memory. There was also around a 5.7% increase in consumption throughput, but I didn't gather enough data samples to know whether this was statistically significant. It also varied by instance in a way that I don't understand yet.

PR #331, opened 2026-07-16
Fix flaky replica_reader tests: preserve await_offset waiters across core rebuild, fix vacuous barrier in local_ahead_discards_manifest

lukebakken:
Fixes two distinct causes of the flaky `replica_reader` tests tracked in #326.

## Mode 2: core rebuild dropped `await_offset` waiters (real reader defect)

`replication_survives_reader_restart` timed out on `await_offset`. The reader was healthy and advancing, but every path that rebuilds the functional core (manifest reinitialize after a commit conflict, local-log-ahead / remote-ahead reset) replaces the whole state via `core:init/2`, which starts with no waiters. A caller already blocked in `await_offset` across such a rebuild was dropped without a reply and hung until its own `gen_server:call` timeout.

Fix: add `carry_waiters/2` to the core and call it at both rebuild sites in the reader (`resolve_and_start/1` and `restart_at_local_floor/3`), moving pending waiters onto the freshly built core and replying to any the resolved manifest already satisfies. Proven via reader pid/state capture across clean reproductions.

Commit 9ff7ed3

## Mode 3: vacuous `await_offset` barrier in `local_ahead_discards_manifest` (test-only)

This case flaked at the "new fragments exist" assertion (roughly 1 in 30 to 100 runs). Reader state captured at the moment of failure showed a single healthy reader, correctly reset to the local floor, with fragment transfers still in flight and no persist completed yet. Nothing regressed and nothing was deleted: the fresh fragments simply had not uploaded when the test asserted.

The barrier was ill-posed. On the local-log-ahead reset the reader resets to the local log's first readable chunk id and installs an empty manifest whose `next_offset` equals that floor. The test derived its `await_offset` target from the segment-file offset reported by `list_segment_offsets` and awaited that plus ten. Because `await_offset` gates on `last_persisted_manifest.next_offset`, the empty reset manifest satisfied it vacuously, before any fresh fragment uploaded, so the assertions raced the in-flight uploads.

`await_offset` cannot be used here: the test cannot compute the reset floor a priori, so any offset it names can be satisfied by the empty reset manifest. The fix waits instead for the observable outcome via the suite's `?awaitMatch` idiom (a fragment that is not one of the pre-restart fragments appears in the remote tier), and relaxes the final assertion from `hd(NewFragments) >= FirstLocal` (which races local retention advancing the segment floor) to `hd(NewFragments) > lists:max(OldFragments)`.

Commit 3250bc8

## Validation

- Mode 3: 300 consecutive isolated runs of `local_ahead_discards_manifest`, 0 failures (the bug previously reproduced by run 30)
- Full `replica_reader` suite: 38 ok, 0 failed
- Mode 2: `replica_reader_core` suite 82 ok; full `replica_reader` suite clean across 40 runs; dialyze and xref pass

## Peer-node death seen during investigation: diagnosed, not a plugin bug

A peer-node death (`erpc,noconnection` cascade across the `with_replica` cases) was seen a couple of times under a 30s hammering diagnostic loop. It is a test-infrastructure artifact, now confirmed end to end.

The `with_replica` group starts one shared peer via `peer:start/1` (`rabbitmq_stream_s3_cth.erl:98`). OTP's `peer` spawns an `origin_link` watchdog on the peer (stdlib `peer.erl:1420`) that monitors the origin (CT) node's control process over distribution and calls `erlang:halt()` on `DOWN`. So whenever the origin/peer distribution connection drops, the peer self-terminates with no `erl_crash.dump`, and every subsequent `erpc` call to the shared peer fails with `{erpc,noconnection}`. Under the 30s hold loop the origin was under enough host CPU/memory pressure that distribution net-ticks were not serviced in time, net_kernel declared the peer down, and the peer halted itself. The reader state captured at that failure (`manifest_next_offset=30`, `persisted_next_offset=30`, `anchor=done`, `waiters=0`) shows the reader had already finished its work, so the plugin was never at fault.

Confirmed both ways:

- Positive: a standalone probe that starts a peer as the hook does, then `erlang:disconnect_node/1` from the origin (what a net-tick timeout does), reproduces the exact signature: the peer OS process is gone (`kill -0` reports no such process), `erpc:call` returns `{erpc,noconnection}`, and no crash dump is written
- Negative: 60 consecutive normal `with_replica` group runs produced 0 peer deaths and 0 `noconnection`; the death only ever appeared under the now-removed 30s hammer loop

This cannot occur in production: the cross-node registry path (`rabbitmq_stream_s3_registry.erl:102`) already catches `_:_` and degrades to `undefined`. No plugin change is warranted.

### DIFF
--- a/src/rabbitmq_stream_s3_replica_reader.erl
+++ b/src/rabbitmq_stream_s3_replica_reader.erl
@@ -1373,10 +1373,15 @@ resolve_and_start(#state{cfg = #cfg{stream = StreamId}} = State0) ->
     case resolve_manifest(StreamId) of
         {ok, Manifest} ->
             State1 = on_manifest_resolved(Manifest, State0),
-            {Core, _Effects} = rabbitmq_stream_s3_replica_reader_core:init(
+            {Core0, _Effects} = rabbitmq_stream_s3_replica_reader_core:init(
                 Manifest, State1#state.config
             ),
-            start_reading(State1#state{core = Core});
+            %% Carry any await_offset waiters from the discarded core so a caller
+            %% blocked across the rebuild is not dropped without a reply.
+            {Core, WaiterEffects} = rabbitmq_stream_s3_replica_reader_core:carry_waiters(
+                State1#state.core, Core0
+            ),
+            start_reading(execute_effects(WaiterEffects, State1#state{core = Core}));
         {retry, Reason} ->
             %% Logged at WARNING (not ERROR) like the sibling init_data_reader
             %% retry in start_reading: this retries once a second, so a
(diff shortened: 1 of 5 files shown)


sdewhitt (approved):

the-mikedavis (approved):
Nice! I believe waiters are only used in tests so this should have no impact outside of testing 👍 

(I might be misremembering though.)

Issue #335, opened 2026-07-20
Audit for raw OTP exit terms leaking through {error, _}-return contracts (found in #332)


lukebakken:
Audit the codebase for raw OTP exit terms that leak through `{error, _}`-return contracts, as first found in #332.

## Background

#332 found that a saturated upload pool exits the `gen_server:call` inside `rabbitmq_stream_s3_api_aws_pool:checkout/2`, and that raw OTP exit term (`{timeout, {gen_server, call, [_Pool, {checkout, _}, _]}}`) escaped the pool boundary and propagated up through code whose contract is `{ok, _} | {error, _}`. The consequence was a misclassification: the transient condition was never matched by name, so it fell through to the non-transient path.

The specific shape is `exit:{timeout, {gen_server, call, [_, _, _]}}` from a `gen_server:call` that timed out or hit a dead process, but the general anti-pattern is broader: any function specced to return `{ok, _} | {error, _}` (or a plain value) that internally calls `gen_server:call`, `gen_statem:call`, `erpc:call`, `gen_server:stop`, etc. and does not catch the resulting exit lets a raw runtime term escape to callers that only handle tagged returns.

## Known instances

- `rabbitmq_stream_s3_api_aws_pool:checkout/2` and its call sites (the subject of #332; being fixed by centralizing the translation in `checkout/2`)

## Task

Audit the codebase for the same class of leak:

- Find every `gen_server:call` / `gen_statem:call` / `erpc:call` / `gen_server:stop` (and similar exit-raising calls) whose enclosing function is specced or documented to return `{ok, _} | {error, _}` or a plain value
- For each, determine whether a timeout, `noproc`, `noconnection`, or `{shutdown, _}` exit can escape to a caller that only handles tagged returns
- Where a leak exists, translate the exit into a named domain term (e.g. `pool_busy`, `timeout`) at the boundary that owns the knowledge, mirroring the `checkout/2` fix, so classifiers and callers never match on raw OTP internals
- Prefer catching and converting at the function that makes the raising call, not in a distant classifier

## Notes

- Not every uncaught exit is a bug: a `gen_server:call` between two internal processes where the callee crashing should crash the caller is fine. The audit targets the boundary case where a raising call sits behind an `{error, _}` contract.
- The cross-node `erpc:call` in `rabbitmq_stream_s3_registry` already catches `_:_` and degrades to `undefined`; that is the correct pattern and can serve as a reference.

lukebakken:
[...]

PR #336, opened 2026-07-20
Fix pool checkout timeout misclassified as non-transient (issue #332)

### LINKED ISSUE #332: is_retriable/1 misclassifies pool checkout timeout as non-transient
A saturated upload pool makes `rabbitmq_stream_s3_api_aws_pool:checkout/2` exit the caller. The governor wraps it (rabbitmq_stream_s3_governor.erl:209, `catch Class:Reason -> {error,{Class,Reason}}`), so `transfer_failed/3` gets:

```
{exit,{timeout,{gen_server,call,[rabbitmq_stream_s3_upload_pool,{checkout,#Ref<...>},10000]}}}
```

`is_retriable/1` matches the bare atom `timeout`, never this wrapped shape, so it falls through to `false`. Same failure mode the function's own comment describes for the old `{http,500}` clauses: matching a term the runtime never produces. Because the governor wraps everything, all upload-task exits hit the fatal branch.

No data loss (both branches `resubmit_transfer_delayed`), but a transient saturation event gets the non-transient retry profile (`upload_retry_delay_ms` 1000ms up to 30000ms instead of the ~10ms transient path), sets the stall gauge, and logs the per-fragment WARNING at replica_reader.erl:994. Observed 216k of these lines in ~23 min on one writer node under load, where the real cause was pool capacity, not the 403/404 class "non-transient" implies.

Fix: unwrap the governor `{Class,Reason}` envelope before classifying and treat `{exit,{timeout,{gen_server,call,[_Pool,{checkout,_},_]}}}` (and wrapped `connection_error`/`timeout`) as transient. Regression test with the governor-wrapped term.

Related: `do_commit` shares `?UPLOAD_POOL` and crashed 134x under the same saturation (replica_reader.erl:2350).

Found this from a Jepsen test against EC2+S3, running the plugin (an adaptation of the Jepsen tests in the plugin here). I'm not sure which nemesis triggered the outage. One nemesis introduces an IAM role restriction that denies traffic to the bucket.

lukebakken:
Fixes #332

## Problem

A saturated upload pool timed out the `gen_server:call` inside `rabbitmq_stream_s3_api_aws_pool:checkout/2`, and that raw OTP exit term (`{timeout, {gen_server, call, [_Pool, {checkout, _}, _]}}`) escaped the pool boundary. The governor wrapped it in its `{error, {Class, Reason}}` envelope, and the replica reader's `is_retriable/1` never matched the wrapped shape, so it fell through to the fatal branch.

The consequence was a misclassification, not data loss (both branches retry): a transient saturation event took the non-transient retry profile (`upload_retry_delay_ms`, 1s escalating to 30s, instead of the fast transient path), set the stall gauge, and logged a per-fragment WARNING. Under load one node emitted 216k of these lines in ~23 minutes, where the real cause was pool capacity, not the 403/404 class that "non-transient" implies.

## Fix

Own the translation in `checkout/2`, the one function that knows the checkout is a timed `gen_server:call`, rather than teaching a distant classifier to pattern-match OTP exit internals.

- `checkout/2` now returns `{ok, conn()} | {error, pool_busy}`. It catches `exit:{timeout, _}` and returns the named `pool_busy`, and still re-raises any other exit (e.g. a genuinely dead pool) after sending the cancellation cast, preserving the prior behaviour for real breakage
- `with/3` matches the tagged return and propagates `{error, pool_busy}`. This also closes a latent leak in the synchronous `request/5` path, which previously let the raw exit escape (and routes PUTs to the same upload pool)
- Both async call sites in `rabbitmq_stream_s3_api_aws` (`stream_put` upload, `start_async_request` read) drop their local `try`/`catch` and match the tagged return, so the OTP-exit-shape match exists nowhere outside the pool
- `is_retriable/1` classifies the named `pool_busy` atom as transient

`pool_busy` is already a first-class transient elsewhere in the codebase: the read path returned it before this change, `remote_reader_core` gives it a dedicated backoff, and it is in `prop_SUITE`'s transient list. This change makes the upload and synchronous paths consistent with the read path.

## The related `do_commit` note in #332

The issue notes that `do_commit` shares the upload pool and crashed 134x under the same saturation. That is covered by this same fix and needs no separate change. `do_commit` calls `rabbitmq_stream_s3_api:put/2`, which reaches the pool through `with/3`; the crash was the raw checkout exit propagating into the spawned commit process. With `with/3` now returning `{error, pool_busy}`, `classify_persist` delivers `persist_failed(pool_busy)`, and `persist_failed/2` already retries on any reason other than `conflict`/`not_found`. Fixing the leak at the pool boundary resolves both the transfer and persist manifestations at once.

## Tests

- `api_aws_pool_statem_SUITE`: a deterministic `checkout_timeout_returns_pool_busy` (exhaust a size-1 pool, assert the next checkout returns `{error, pool_busy}`)
- `replica_reader_core_SUITE`: `transfer_failed_pool_busy_retriable_resubmits` asserts `pool_busy` is classified transient (`resubmit_transfer`, not `resubmit_transfer_delayed`)

## Validation

- `api_aws_pool_statem_SUITE`: 2 ok, 0 failed
- `replica_reader_core_SUITE`: 83 ok, 0 failed
- `replica_reader_SUITE`: 38 ok, 0 failed
- `replica_reader_tasks_SUITE`: 13 ok, 0 failed
- `gmake dialyze`: passed
- `gmake xref`: passed

## Follow-up

#335 tracks a codebase-wide audit for the same class of leak (a raw exit-raising `*:call` behind an `{ok, _} | {error, _}` contract).

### DIFF
--- a/test/api_aws_pool_statem_SUITE.erl
+++ b/test/api_aws_pool_statem_SUITE.erl
@@ -45,6 +45,7 @@ over the invariants above.
 
 -include_lib("common_test/include/ct.hrl").
 -include_lib("proper/include/proper.hrl").
+-include_lib("eunit/include/eunit.hrl").
 
 -define(POOL, statem_aws_pool).
 -define(MIN_SIZE, 1).
@@ -57,7 +58,7 @@ over the invariants above.
 -define(AWAIT_MS, 1500).
 
 all() ->
-    [no_leak_or_inconsistency].
+    [no_leak_or_inconsistency, checkout_timeout_returns_pool_busy].
 
 init_per_suite(Config) ->
     {ok, _} = application:ensure_all_started(seshat),
@@ -75,6 +76,26 @@ end_per_suite(Config) ->
 
 no_leak_or_inconsistency(_Config) ->
     rabbit_ct_proper_helpers:run_proper(fun prop_no_leak_or_inconsistency/0, [], 150).
+
+%% A saturated pool must surface as {error, pool_busy}, not a raw
+%% gen_server:call exit escaping the pool boundary (issue #332).
+checkout_timeout_returns_pool_busy(_Config) ->
+    _ = seshat:new_group(rabbitmq_stream_s3),
+    Config = (pool_config())#{name => busy_pool, min_size => 1, max_size => 1},
+    {ok, Pid} = rabbitmq_stream_s3_api_aws_pool:start_link(busy_pool, Config),
+    try
+        %% Take the pool's only connection and hold it, so the next checkout
+        %% cannot be served and must time out.
+        {ok, Conn} = rabbitmq_stream_s3_api_aws_pool:checkout(busy_pool, 1000),
+        ?assertEqual(
+            {error, pool_busy},
+            rabbitmq_stream_s3_api_aws_pool:checkout(busy_pool, 100)
+        ),
+        %% The connection held above is still valid and checks back in cleanly.
+        ok = rabbitmq_stream_s3_api_aws_pool:checkin(busy_pool, Conn)
+    after
+        gen_server:stop(Pid)
+    end.
 
 prop_no_leak_or_inconsistency() ->
     ?FORALL(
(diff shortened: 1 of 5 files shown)


test/api_aws_pool_statem_SUITE.erl
```diff
@@ -76,6 +77,26 @@ end_per_suite(Config) ->
...
+        {ok, Conn} = rabbitmq_stream_s3_api_aws_pool:checkout(busy_pool, 1000),
+        ?assertEqual(
+            {error, pool_busy},
+            rabbitmq_stream_s3_api_aws_pool:checkout(busy_pool, 100)
+        ),
+        %% The connection held above is still valid and checks back in cleanly.
+        ok = rabbitmq_stream_s3_api_aws_pool:checkin(busy_pool, Conn)
+    after
+        gen_server:stop(Pid)
+    end.
```
the-mikedavis: This might work better as an eunit test in the pool module - I'd like to keep this suite specific to the proper_statem stuff

lukebakken: Got it

the-mikedavis (approved):

PR #339, opened 2026-07-21
Mark manifests as pending during resolution

### LINKED ISSUE #333: `first` offset log reader skips remote tier shortly after node reboot
A consumer attaching with `first` shortly after node boot can read from the local floor instead of the beginning of the stream in S3, silently skipping the entire remote tier.

The manifest cache is seeded from a publish-driven persist, the reconciler's periodic re-seed tick (about 60 seconds), or a resolved manifest fetched at replica reader startup. On an idle stream right after boot, none of these have run yet, so `get_manifest(StreamId)` returns `undefined`. `rabbitmq_stream_s3_log_reader:resolve_remote_location/2` treats `undefined` as "this stream has no remote tier" and resolves `first` to `{local, first}`, which is correct for a genuinely un-tiered stream and wrong for a tiered stream whose cache just hasn't been populated yet. Nothing distinguishes the two.

The deeper problem is that the cache has two real states doing the job of three: attached-but-not-yet-resolved, and never-attached, both collapse to `undefined`. Each call site invented its own fallback for that value independently (read path fails open, retention fails safe, GC fails closed), with nothing enforcing that those fallbacks agree with each other or with any stated availability contract. `docs/invariants.md`'s exactly-once read invariant is stated over the manifest itself, not over the node-local cache the read path actually consults.

It also wasn't caught. Every suite exercising the read path publishes before reading, which seeds the cache as a side effect of arranging the fixture, so the cold path never got exercised: the suite would need to subscribe before any publish on a freshly restarted node with no replica to re-seed it via sync. The P models of the read path model the cache as always present (`bool nonEmpty`), so a miss is outside the alphabet and unrepresentable, let alone checkable.

Found independently by both manual testing and a Jepsen test against EC2

the-mikedavis:
*Issue #, if available:* fixes https://github.com/amazon-mq/rabbitmq-stream-s3/issues/333

*Description of changes:* I was surprised to find https://github.com/amazon-mq/rabbitmq-stream-s3/issues/333 manually. That the bug survived is a failure of the documentation and P models, and there's room for improvement in the manifest cache API to make decisions around cache state explicit. There are a lot of changes in this branch around that theme, but the actual changes to the manifest cache and log reader are fairly straightforward.

When starting, we mark the manifest for a stream as pending if it hasn't been downloaded yet. This allows us to defer operations that need the manifest (e.g. reading from `first`) until it has been downloaded - they fail closed.

### DIFF
--- a/src/rabbitmq_stream_s3_gc.erl
+++ b/src/rabbitmq_stream_s3_gc.erl
@@ -286,8 +286,8 @@ still_dangling(#{reason := no_anchor}) ->
     %% the absence is permanent, so there is nothing to re-validate.
     true;
 still_dangling(#{reason := below_first_offset, stream_id := StreamId, key := Key}) ->
-    case rabbitmq_stream_s3_manifest_replica:get_manifest(StreamId) of
-        #manifest{first_offset = FirstOffset} = Manifest ->
+    rabbitmq_stream_s3_manifest_replica:with_manifest(StreamId, #{
+        resolved => fun(#manifest{first_offset = FirstOffset} = Manifest) ->
             case parse_key(Key) of
                 {data, _StreamId, Offset} ->
                     Offset < FirstOffset;
@@ -296,14 +296,14 @@ still_dangling(#{reason := below_first_offset, stream_id := StreamId, key := Key
                         not live_leading_group(StreamId, Key, Manifest);
                 _ ->
                     false
-            end;
-        Unresolved when Unresolved =:= undefined; Unresolved =:= pending ->
-            %% No live manifest to compare against (missing row, or a pending
-            %% marker whose manifest has not resolved yet): do not delete (a
-            %% later sweep reclaims a genuine orphan once a floor is known
-            %% again).
-            false
-    end.
+            end
+        end,
+        %% No live manifest to compare against (a pending marker whose manifest
+        %% has not resolved yet, or a missing row): do not delete (a later
+        %% sweep reclaims a genuine orphan once a floor is known again).
+        pending => fun() -> false end,
+        absent => fun() -> false end
+    }).
 
 %% Whether the group object Key is protected by the LIVE manifest's leading-group
 %% carve-out: it is the live referenced leading group, or the live manifest is in
--- a/src/rabbitmq_stream_s3_gc.erl
+++ b/src/rabbitmq_stream_s3_gc.erl
@@ -297,9 +297,11 @@ still_dangling(#{reason := below_first_offset, stream_id := StreamId, key := Key
                 _ ->
                     false
             end;
-        undefined ->
-            %% No live manifest to compare against: do not delete (a later sweep
-            %% reclaims a genuine orphan once a floor is known again).
+        Unresolved when Unresolved =:= undefined; Unresolved =:= pending ->
+            %% No live manifest to compare against (missing row, or a pending
+            %% marker whose manifest has not resolved yet): do not delete (a
+            %% later sweep reclaims a genuine orphan once a floor is known
+            %% again).
             false
     end.
 
--- a/src/rabbitmq_stream_s3_hooks.erl
+++ b/src/rabbitmq_stream_s3_hooks.erl
@@ -150,20 +150,22 @@ on_retention_evaluated(Cnt, #{name := Name}) ->
     %% of -1, which must never reach the counter. get_range/1 returns `empty`
     %% in that case, so the override only runs when the remote tier actually
     %% holds data and its first_timestamp is a real timestamp.
-    case rabbitmq_stream_s3_manifest_replica:get_manifest(StreamId) of
-        #manifest{entries = <<>>} ->
-            ok;
-        #manifest{first_offset = RemoteFirst, first_timestamp = RemoteFirstTs} ->
-            LocalFirst = counters:get(Cnt, ?C_OSIRIS_LOG_FIRST_OFFSET),
-            counters:put(Cnt, ?C_OSIRIS_LOG_FIRST_OFFSET, min(LocalFirst, RemoteFirst)),
-            LocalFirstTs = counters:get(Cnt, ?C_OSIRIS_LOG_FIRST_TIMESTAMP),
-            counters:put(Cnt, ?C_OSIRIS_LOG_FIRST_TIMESTAMP, min(LocalFirstTs, RemoteFirstTs));
-        pending ->
-            %% Not yet resolved: nothing known to override the counters with.
-            ok;
-        undefined ->
-            ok
-    end.
+    rabbitmq_stream_s3_manifest_replica:with_manifest(StreamId, #{
+        resolved => fun
+            (#manifest{entries = <<>>}) ->
+                ok;
+            (#manifest{first_offset = RemoteFirst, first_timestamp = RemoteFirstTs}) ->
+                LocalFirst = counters:get(Cnt, ?C_OSIRIS_LOG_FIRST_OFFSET),
+                counters:put(Cnt, ?C_OSIRIS_LOG_FIRST_OFFSET, min(LocalFirst, RemoteFirst)),
+                LocalFirstTs = counters:get(Cnt, ?C_OSIRIS_LOG_FIRST_TIMESTAMP),
+                counters:put(
+                    Cnt, ?C_OSIRIS_LOG_FIRST_TIMESTAMP, min(LocalFirstTs, RemoteFirstTs)
+                )
+        end,
+        %% Not yet resolved: nothing known to override the counters with.
+        pending => fun() -> ok end,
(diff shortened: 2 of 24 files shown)


lukebakken (commented):
High-effort review of this PR. The pending-manifest design is sound and the `with_manifest/2` total interface is a real improvement: making all three cache states mandatory map keys turns a forgotten state into a match failure at the call site rather than a silent fallback.

Findings below. Six are inline on the relevant lines. Three more are not in this PR's diff because they live in the core stream code (`deps/rabbit`, `deps/rabbitmq_stream`) that consumes the widened resolver contract, so they are recorded here.

## Cross-repo (core stream consumers, not in this diff)

These matter because `init_offset_reader/2` can now return `{error, {manifest_not_resolved, _}}`, a shape that did not exist before this PR. The core call sites were written against the old always-`{ok, _}` (or never-erroring) contract:

1. `deps/rabbit/src/rabbit_stream_queue.erl:488` (AMQP `begin_stream/7`) hard-matches `{ok, Seg0} = osiris:init_reader(...)`. An AMQP consumer subscribing at `first` or below the local floor on a cold-cache tiered stream now hits a `badmatch` crash instead of a retryable error. Severity depends on whether AMQP consumers actually subscribe at `first`/below-floor on tiered streams in practice.
2. `deps/rabbit/src/rabbit_stream_queue.erl:680` (`stream_local_member_change` re-seat) has the same unguarded `{ok, Log1} = osiris:init_reader(...)` badmatch on the pending error.
3. `deps/rabbitmq_stream/src/rabbit_stream_reader.erl:2334` (native protocol): a resolver failure now yields `{?RESPONSE_CODE_NO_OFFSET, 0, ...}`, where `first`/timestamp specs previously never errored on this path, a silent behavior change for native stream consumers.

Whether we harden the core call sites, soften the plugin's fail-closed contract, or accept and document the window is a scoping call worth deciding before merge.

src/rabbitmq_stream_s3_hooks.erl
```diff
@@ -92,6 +92,11 @@ on_init(acceptor, Pid, #{name := Name, leader_pid := LeaderPid, counter := Count
     rabbitmq_stream_s3_manifest_replica:register_replica_context(
         StreamId, Pid, Dir, Shared, Counter
     ),
+    %% Mark the cache row pending before the member finishes init so a consumer
+    %% attaching on this node fails closed (and retries) until the writer's
+    %% sync lands, rather than resolving the remote tier as absent.
+    %% Insert-if-absent: a sync that already landed is never downgraded.
+    ok = rabbitmq_stream_s3_manifest_replica:mark_pending(StreamId),
```
lukebakken: The pending marker is hand-paired with `register_replica_context` at four sites (here, line 111, line 376, and `replica_reader.erl:391`). Establishing it at each call site is fragile: any new attach path that registers a context but forgets this line silently regresses to serving a cold cache as "absent". See the separate comment on line 376 for a site that already does. Consider folding the `insert_new` pending write into the `register_replica_context` gen_server handler so no caller can omit it (the writer path in `replica_reader.erl` would keep its own call, as it does not go through `register_replica_context`).

src/rabbitmq_stream_s3_hooks.erl
```diff
@@ -356,6 +370,10 @@ attach_replica(Pid) ->
     rabbitmq_stream_s3_manifest_replica:register_replica_context(
         StreamId, Pid, Dir, Shared, Counter
     ),
+    %% Same pending marker as on_init(acceptor, ...): readers fail closed until
+    %% the writer's sync lands. Insert-if-absent, so re-discovery of an already
+    %% synced replica changes nothing.
+    ok = rabbitmq_stream_s3_manifest_replica:mark_pending(StreamId),
```
lukebakken: This `mark_pending` is correctly paired with the `register_replica_context` call above it. The reconcile path `register_replica_context/2` at line 283 calls `register_replica_context` but does not call `mark_pending`. A replica re-registered by reconciliation (after the `manifest_replica` singleton restarted and dropped its context) therefore leaves no pending row, so a consumer attaching in that window resolves the remote tier as absent instead of failing closed, the exact silent-skip this PR sets out to prevent.

src/rabbitmq_stream_s3_manifest_replica.erl
```diff
@@ -409,7 +486,7 @@ handle_cast({put_manifest, StreamId, Manifest, Epoch}, State) ->
     {noreply, State};
 handle_cast({apply_edit, StreamId, Edit}, State) ->
     case ets:lookup(?TABLE, StreamId) of
-        [{_, Manifest0, Epoch0}] ->
+        [{_, #manifest{} = Manifest0, Epoch0}] ->
```
lukebakken, line 497: This cast handler is narrowed to `#manifest{}` here. Its sibling `handle_cast({apply_edits, ...})` at line 526 still matches the loose `[{_, Manifest0, _}]`, so a `pending` row would flow into `apply_edits_catching/3` as if it were a manifest. The synchronous `handle_call({apply_edits, ...})` at line 380 was narrowed; the cast at 526 appears to have been missed. Narrowing it to `[{_, #manifest{} = Manifest0, _}]` gives the same fail-into-resync behavior.

src/rabbitmq_stream_s3_log_reader.erl
```diff
@@ -185,6 +194,24 @@ init_offset_reader(OffsetSpec, Config) ->
...
+            %% the replica reader's manifest_resolution_failures) is the
+            %% alertable signal.
+            ?LOG_INFO(
+                "Failing reader setup closed for stream '~ts': offset spec ~w "
+                "may be in the remote tier but the manifest is not yet "
+                "resolved on this node",
+                [StreamId, OffsetSpec],
+                ?DOMAIN
+            ),
+            record_resolve_failed_closed(T0),
```
lukebakken, line 221: `init_offset_reader/2` records `resolve_failed_closed` here. The sibling callback `resolve_offset_spec/2` at line 178 returns `{error, _}` at line 184 without recording the counter. If any consumer path drives resolution through `resolve_offset_spec/2` rather than `init_offset_reader/2`, fail-closed events on that path are invisible to the metric that this PR's own log message points at as the alertable signal.

src/rabbitmq_stream_s3_manifest_replica.erl
```diff
@@ -286,6 +350,11 @@ init([]) ->
 handle_call({put_manifest, StreamId, Manifest, Epoch}, _From, State) ->
     write_manifest(StreamId, Manifest, Epoch),
     {reply, ok, State};
+handle_call({mark_pending, StreamId}, _From, State) ->
+    %% Insert-if-absent: never downgrade a resolved manifest (or an existing
+    %% pending marker) that survived a reader restart.
+    _ = ets:insert_new(?TABLE, {StreamId, pending, undefined}),
```
lukebakken, line 358: The `insert_new` no-downgrade guarantee (a resolved row must never revert to `pending`) is the load-bearing reason `mark_pending` is safe to call on every attach/rediscovery. There is no test asserting it. Suggest a case in `manifest_replica_SUITE`: `put_manifest` then `mark_pending`, then assert the row is still the resolved `#manifest{}`.

src/rabbitmq_stream_s3_gc.erl
```diff
@@ -296,12 +296,14 @@ still_dangling(#{reason := below_first_offset, stream_id := StreamId, key := Key
...
-            %% No live manifest to compare against: do not delete (a later sweep
-            %% reclaims a genuine orphan once a floor is known again).
-            false
-    end.
+            end
+        end,
+        %% No live manifest to compare against (a pending marker whose manifest
+        %% has not resolved yet, or a missing row): do not delete (a later
+        %% sweep reclaims a genuine orphan once a floor is known again).
+        pending => fun() -> false end,
```
lukebakken, line 304: The `still_dangling` refactor adds `pending`/`absent` branches that both return `false` (do not delete). The GC property suite does not appear to exercise a `below_first_offset` candidate against a `pending` cache row, so the "never delete on an unresolved manifest" invariant is untested. A property case that marks the row pending and asserts `still_dangling` returns `false` would cover it.

the-mikedavis:
Agreed on the server changes. I will draft some changes and update the branch, or make an issue to follow-up

the-mikedavis:
I force-pushed the osiris branch to add in a `{error, unavailable}` return for the log reader which can be retried. Now I'll make some changes in the server branch to handle that

the-mikedavis:
Initial swing at the server change: https://github.com/amazon-mq/upstream-to-rabbitmq-server/commit/d41bd5941e5df7402a1d407138c9d03116635058

I also rebased on the v4.3.3 tag.

I'll make a follow-up here to start using the `{error, unavailable}` return.

PR #343, opened 2026-07-28
Cancel unnecssary governor submissions

### LINKED ISSUE #342: In-flight fragment uploads are not cancelled when their stream is deleted
### Describe the bug

When a stream is deleted, its replica reader's in-flight fragment uploads keep running to completion instead of being cancelled, wasting upload-pool capacity on work for a stream that no longer exists.

Deletion reaches the replica reader through the `stop` core effect, emitted from `persist_failed/2` in `rabbitmq_stream_s3_replica_reader_core.erl` when a persist gets `not_found` against a non-zero expected revision (the stream's Khepri metadata node is gone). `execute_effect(stop, State)` in `rabbitmq_stream_s3_replica_reader.erl` just sets `stopping = true`, and the gen_server stops on the next handler pass via `maybe_stop/1`. `terminate/2` kills any in-flight persist task and closes the local log reader, but does nothing about outstanding transfers.

Those transfers were handed to the governor via `rabbitmq_stream_s3_governor:submit/4`, a cast; the governor spawns a bare process per admitted transfer (`spawn_task/1`) and keeps no record of which reader submitted it. The spawned process runs the upload to completion (success, retries exhausted, or a liveness deadline) and then sends `{transfer_result, Ref, Result}` back to `ReplyTo`, which is by then the dead reader pid. The message is simply dropped. So the upload keeps occupying a pool connection until it finishes naturally, well after the stream is gone.

### Reproduction steps

1. Sustain producer load on a stream so it has multiple fragment uploads in flight
2. Delete the stream
3. Observe that `transfers_in_flight` for that stream/node does not drop immediately, and the upload pool's checked-out connection count is unaffected by the deletion; the in-flight uploads for the deleted stream keep running

### Expected behavior

Deleting a stream should immediately reclaim whatever pool capacity its in-flight uploads were occupying, by cancelling them rather than letting them run to completion.

This is safe to do: `transfer_failed/3`'s core invariant (issue #206), that a confirmed fragment must never be abandoned, or the manifest could advance past a range that never reached S3, only holds for a *live* stream with a durable, referenced manifest. Once the stream is deleted, the manifest is gone and there is no downstream reader that could observe a hole, so killing outstanding transfers at that point does not risk correctness. The pool side is already safe for this too: `handle_down/3` in `rabbitmq_stream_s3_api_aws_pool.erl` (issue #177) already handles a caller dying while holding a checked-out connection by dropping the connection and growing a replacement, which is exactly what happens if an upload task is killed mid-transfer.

### Additional context

What's missing: the governor has no way to reach a submitted task by its `Ref`. `submit/4` only carries `Fun`, `Size`, `ReplyTo`, `Ref`; the governor's `#state{}` retains no Ref-to-pid mapping for spawned tasks, and no way to remove a still-pending (not yet admitted) item beyond scanning `pending`.

To support cancellation, the governor would need:
- A map from `Ref` to the spawned task pid, populated in `spawn_task/1` and cleared when the result arrives
- A `cancel/1` API that, for an admitted task, kills the pid and drops the `Ref` from the map; for a still-queued item, removes it from `pending` without ever spawning it

The reader side already has the full set of outstanding `Ref`s via `rabbitmq_stream_s3_replica_reader_tasks:transfers/1`. The natural call site is right where `stopping` gets set, in `execute_effect(stop, State)` (`rabbitmq_stream_s3_replica_reader.erl:1153`), before the gen_server stops: walk the transfers map and cancel each `Ref`.

Scope: this is deletion-triggered cancellation only. It does not cover cancelling uploads on writer failover; a deposed replica reader's in-flight fragment becomes an orphan object in S3 (see `failure-modes.md`), which is harmless, just wasted work, and is a separate question.

Found while investigating a Jepsen reboot-churn run that produced a ~161,000-transfer upload backlog; every stream was deleted at the end of that run and the backlog drained at the same slow rate regardless (`active_requests` sampling made the drain rate itself hard to read; see the checkout-timeout and admission-control issues filed alongside this one for what was actually gating it).

the-mikedavis:
*Issue #, if available:* closes #342

*Description of changes:* Upload submissions are casts to the governor. We can cancel some of these submissions if they are no longer necessary. Two examples: 1/ stream has been deleted, 2/ uploading replica reader process crashed. The two commits here cover these cases both for the regular path (stream deletion) and for degenerate cases like a large governor message backlog. We also introduce a new metric to capture this.

I expect that this change will be noticeable in my Jepsen tests which induce a very large governor backlog.

### DIFF
--- a/docs/conventions.md
+++ b/docs/conventions.md
@@ -32,6 +32,14 @@ For the manifest cache this is enforced structurally, not by convention: branchi
 
 The direction of each explicit miss clause matters: consumers must fail in the direction that cannot lose data. Readers fail closed (error, the consumer retries), retention deletes nothing, GC skips, resolution goes to the authoritative store. A fallback that degrades silently reads as robustness in review and is exactly where correctness leaks out.
 
+## Derived gauges
+
+A gauge whose value is a pure function of process state should be published from that state on every handler return, not incremented and decremented at the edges. Counters live in `persistent_term` and outlive the process that maintains them; the state they describe does not. An inc/dec gauge therefore ratchets permanently on the one path that deliberately runs no cleanup - a crash, where the pending decrements are lost with the old incarnation's mailbox and the fresh one starts from empty state. A derived gauge self-heals, because `init/1` publishes from the empty state and each later mutation republishes the truth. It also cannot drift from the state it reports, since there is no second place to forget an edge.
+
+`rabbitmq_stream_s3_governor:derive_gauges/1` is the reference shape: one function, called on every callback return that can change what it reports. See the Derived gauges section of that module's moduledoc.
+
+Where a gauge spans several processes it cannot be derived from any one of them, and inc/dec is the only option. Then weld both edges to the single state mutation they mirror, so an edge cannot be added without the counter following: api_aws's `active_requests` counts one in-flight request per pool checkout across every pool, and its two edges live in `rabbitmq_stream_s3_api_aws_pool:add_checkout/3` and `del_checkout/3`, the only two places the `checkouts` maps change. Keep the counter's key, size, and indices private to the module that defines `?COUNTERS`, and reach it from elsewhere (including tests in other modules) through an exported accessor keyed by metric name rather than a restated index.
+
 ## Cold-state test dimension
 
 Every stateful fixture a test warms up implicitly is a state the suite must also construct cold. An end-to-end test that publishes and then reads has, as a side effect, seeded the manifest cache, attached the hooks, and resolved the manifest: the act of arranging the fixture destroys the cold-start state, so the warm path is the only path such suites can ever exercise.
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -304,8 +304,13 @@ Owned by `rabbitmq_stream_s3_governor`. Single counter set per node, no per-stre
 |-------------------------------------|---------|-------------------------------------------------------------------|
 | `rabbitmq_stream_s3_governor_submissions_received`     | counter | Total transfer submissions received                               |
 | `rabbitmq_stream_s3_governor_oversized_admissions`     | counter | Transfers admitted on credit because they exceed the burst allowance; a persistently climbing value means the configured rate is low relative to fragment sizes |
+| `rabbitmq_stream_s3_governor_dropped_dead_replyto`     | counter | Submissions dropped because the requester was already dead, either at intake or while queued behind the token bucket; expected to climb during stream deletion or replica-reader restarts, and not an error |
 | `rabbitmq_stream_s3_governor_tasks_in_flight`          | gauge   | Transfer tasks currently executing                                |
 | `rabbitmq_stream_s3_governor_pending_submissions`      | gauge   | Submissions queued waiting for token-bucket capacity              |
+
+Both governor gauges are derived from the state they describe rather than incremented and decremented, so they self-heal after a governor crash instead of ratcheting. See the Derived gauges section of the `rabbitmq_stream_s3_governor` moduledoc, and [conventions.md](./conventions.md#derived-gauges).
+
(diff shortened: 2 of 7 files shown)


lukebakken (approved):
Approving. Reviewed the diff and validated the branch on a live 3-node cluster.

## What this addresses

Beyond the cancellation work in #342, a high-effort review of this branch surfaced 10 findings, all fixed here, one per commit. The substantive ones:

- Gauges are now derived from the state they describe and republished on every handler return, rather than incremented at the edges. seshat counters live in `persistent_term`, so they outlive the gen_server while `#state{}` resets; an inc/dec gauge therefore ratchets on the one path that runs no cleanup. Where a gauge spans processes (`active_requests`) it cannot be derived, so both edges are welded to the single map mutation they mirror, and the pool now owns it outright.
- The governor stops on an exit signal from an unexpected link instead of silently swallowing it. Setting `trap_exit` for the parent path would otherwise have made the process immortal against every other link, which is weaker than the untrapped default.
- A requester is watched from intake rather than from task spawn, so a submission that dies while still queued behind the token bucket is reaped. `drop_pending/2` does this in one pass and releases each dropped item's watch. Note the original review framed this as an O(n^2) drop inside the refill tick; that framing was wrong, since `drain_pending/1` only ever inspects the queue head.
- Running tasks are keyed by monitor reference so no attempt is orphaned, and a still-queued submission sharing a Ref with a running task is cancelled.

## Verification

Local: eunit 236/236 across the plugin, CT suites green (replica_reader 39/39, api_aws 83/83, replica_reader_statem 13/13, plus replica_reader_core, tasks, api_aws_pool_statem), dialyzer and xref clean, erlfmt clean.

Live, on a 3-node m7g.8xlarge cluster running 966b7e3, in both governor configurations:

**Unrestricted (`max_transfer_bytes_per_sec = unlimited`, the default).** 45-minute soak: 898.8M messages at a flat 334.8k msg/s, ~330 MiB/s sustained to S3, 13,230 submissions. Replay verification passed with all 8 consumers independently reading the full stream across both tiers, 9,897,137 messages each, no gaps. At quiescence every node reported `pending=0 tasks=0 requesters=0 checked_out=0`, with 211,328 checkouts against 211,161 check-ins and no leaked connection. Zero crashes, zero retries, zero 5xx.

**Paced (100MB/s rate, 200MB burst).** Burst was set explicitly above the measured ~75.5 MB fragment size so submissions queue normally rather than being admitted on credit; `governor_oversized_admissions` stayed at 0, confirming that. This reached the state the default configuration cannot: with items actually waiting on the token bucket, deleting the stream produced `governor_dropped_dead_replyto = 3`, exercising the queued-drop path live rather than only in eunit. Counters reconcile exactly: 26 submissions = 5 completed + 3 dropped while queued + 3 retention-race failures + 15 cancelled at deletion. `requesters` still unwound to 0 on the mixed path where some items were dropped from the queue and others cancelled mid-flight, which is the refcount invariant this branch depends on.

Pool capacity is reclaimed promptly on deletion as #342 asked: the cancelled uploads' connections returned through the caller-death path from #177, leaving the pool fully idle.

## Note on the integration suite

`make stream-deletion` failed once during validation, leaving a single S3 object. That is #196, not a regression here: this branch touches no deletion-path file, the fragment's PUT was already on the wire ~570ms before `delete_member`, and a dry-run GC sweep identifies the object as `no_anchor`, so GC would reclaim it. The test cannot pass on a default cluster because `gc_enabled` is `false` with a 24h interval, which is shorter than its 60s deadline. Details and suggested remedies are on #196. Under pacing the same test passed with an empty prefix.

the-mikedavis:
I will make a build for another round of Jepsen testing on EC2

the-mikedavis:
I think this worked as expected: my Jepsen tests no longer show a huge spike of in-flight data after streams have been deleted under latency or S3 outage. I didn't see the new metric increment, but I think that's expected as I don't have the governor's token bucket configured in my environment yet - so nothing enters the `pending` queue

the-mikedavis:
After setting up the governor token bucket settings I saw a real spike of

```promql
rate(rabbitmq_stream_s3_governor_dropped_dead_replyto[60s])
```

on a test broker, deleting streams when there was a big backlog after a latency nemesis.

Screenshot 2026-07-30 at 15 46 40

PR #348, opened 2026-07-30
Remove 'absent' state from manifest cache

### LINKED ISSUE #346: osiris replica init crash on unguarded manifest-replica singleton call
**Describe the bug**
`rabbitmq_stream_s3_hooks:on_init/3` (the acceptor path, at lines 96 and 108) calls `rabbitmq_stream_s3_manifest_replica:register_replica_context/5`, which issues an unguarded `gen_server:call` to a bare, node-local registered process name (`rabbitmq_stream_s3_manifest_replica.erl:306-307`). When that per-node singleton is not yet started, as during node boot, or is mid-shutdown, as during a node kill, the call exits with `{noproc, ...}` or `{shutdown, ...}` in the caller. The caller is the `osiris_replica` process running `osiris_log:init/2`, so the exit crashes the replica. Its supervisor restarts it, which re-races the still-unavailable singleton. `discover_child/1` and `reconcile_child/1` in the same module already wrap this identical call in a try/catch that logs and continues; `on_init/3` is the one call site without that tolerance, and `apply_edits/4` and `evaluate_local_retention/1` in `rabbitmq_stream_s3_manifest_replica.erl` already convert this same call's exit into a named `{error, {manifest_replica_down, _}}` result rather than letting it escape.

**To Reproduce**
Run a 3-node cluster with the `rabbitmq_stream_s3` plugin under stream traffic and repeatedly kill and restart the RabbitMQ application on each node in quick succession (a Jepsen-style fault injection run did this 6 to 8 times per node within one hour). Osiris stream replica recovery runs early during node boot and calls the node-local `rabbitmq_stream_s3_manifest_replica` gen_server via the acceptor path of `on_init/3` before or during the window that singleton is unavailable.

**Expected behavior**
Replica initialization should tolerate the manifest-replica singleton being temporarily unavailable rather than crashing the replica process.

**Environment**

EC2 m7g.xlarge 3-node cluster running latest main (80e17a87f717296898092dc2f52f9be14f23601d).

Window: 2026-07-29T23:26:45Z to 2026-07-30T00:26:47Z

**Additional context**
About 3,044 distinct `osiris_replica:init/1` crashes occurred cluster-wide during the 1-hour window. The failure reason split is roughly 13,436 line-mentions of `{noproc, ...}` versus 4,322 of `{shutdown, ...}`. Crash bursts at 23:37 to 23:44 and 00:25 to 00:26 line up exactly with node restart windows.

The same root cause also drops in-flight syncs: `rabbitmq_stream_s3_syncs_dropped_no_context` reached 1,002 cluster-wide over the window, compensated roughly 1:1 by `rabbitmq_stream_s3_resyncs_requested` at 1,084. Compensation is confirmed by counter correlation; full resync completion is not verified from counters alone.

No S3 backend errors, no data or resource alarms, and no cluster partition or minority signatures were observed. `rabbitmq_stream_s3_edits_dropped_no_context` was 0 for the whole window: the write path drops nothing, because the crash happens before any edit is processed. So this is a robustness and availability issue (replica churn, noisy crash logs, retry storms), not an observed data-loss issue in this run. Correctness (message loss or duplication) cannot be fully ruled out from logs and metrics alone; that verdict belongs to the Jepsen harness's own consistency check.

### LINKED ISSUE #347: Manifest cache uses 'absent' as a default state
The manifest cache has three states for a manifest, enforced by a `with_manifest` helper that takes functions to execute depending on the state the manifest is in: `#manifest{}` (present), absent, or `pending`. These states are OK themselves, but the way the lifecycles fit together, a manifest is considered absent if it has not been marked as `pending`. Marking `pending` is done through a gen_server call, and the underlying storage (ETS table) is tied to the lifecycle of that gen_server.

This is incorrect: if a replica process starts before the plugin is initialized and the manifest cache server is not started, failing to register context means that the manifest cache will treat the manifest as `absent`. Reconciliation will eventually correct that, but in that window, readers which connect to that replica will treat `first` as the beginning of the _local_ log.

This needs to be swapped so that absent is a positive state and pending is the default state.

Currently this is protected against by the bug https://github.com/amazon-mq/rabbitmq-stream-s3/issues/346 - if a replica attempts to set as `pending` while the manifest replica server is not running, the replica crashes. This issue must be fixed in order to fix #346.

Found this behavior manually as a result of fixing #346. (So I'll say that it was indirectly found by Jepsen tests 😁)

the-mikedavis:
*Issue #, if available:* fixes #346, fixes #347

*Description of changes:* This is two fixes bundled together. Bundling them together is necessary because #347 is protected by the bug #346. The first fix avoids replica process crashes when the manifest cache is not yet online, or is shutting down. The second fix removes the "absent" state from the manifest cache. `pending` becomes the default state for a manifest. Once _resolved_, the manifest can either have data or be empty, but it cannot be absent.


src/rabbitmq_stream_s3_manifest_replica.erl
```diff
@@ -302,9 +303,20 @@ its predecessor's DOWN evict the live context.
...
+        gen_server:call(
+            ?MODULE, {register_replica_context, StreamId, MemberPid, Dir, Shared, Counter}
+        )
+    catch
+        exit:{noproc, _} ->
+            {error, noproc};
+        exit:{shutdown, _} ->
+            {error, shutdown};
+        exit:{timeout, _} ->
+            {error, timeout}
```
the-mikedavis, line 318: This captures a different set of exits than the other calls in this module. I prefer this subset for this specific call, I don't want a catchall - if there's something unexpected I'd rather it exit the replica

the-mikedavis:
I'm going to merge this down as I did a few rounds of review before submitting (and I don't like the remaining results :P), and this is what the Jepsen tests are now running into more than anything (against the `kill` nemesis).

PR #351, opened 2026-07-31
remote reader: Pipeline prefetch across concurrent range GETs

### LINKED ISSUE #349: Remote reader throughput is bottlenecked by (lack of) fetch concurrency
Remote-tier prefetch issues at most one outstanding S3 range GET per fragment at a time. That caps a single remote reader's bandwidth at roughly one S3 connection's transfer rate, below what a lagging consumer can drain and below the local (page-cache) tier. AIMD grows the fetch *size*, not fetch *concurrency*, so it cannot close this gap: on a fetch-latency-bound workload the window gets halved by misses about as fast as hits can grow it, and settles near its floor regardless of `read_size_max`.

Found while running a test broker through a high throughput produce/consume workflow. Broker running an `m7g.2xlarge` 3-node cluster with three client EC2s (`c6gn.xlarge`) each using this `stream-perf-test` setup:

```bash
stream-perf-test \
  --streams "stream-node-$CLIENT_NO" \
  --uris "rabbitmq-stream+tls://$AUTH@$BROKER_HOST:5551/%2f" \
  --producers 1 \
  --consumers 2 \
  --size 65536 \
  --rate 1192 \
  --confirm-latency \
  # ...
```

With the stream writers, producers and consumers all balanced evenly, consumers eventually fell behind and `rabbitmq_stream_consumer_max_offset_lag` grew linearly. In this setup, the remote reader was slower than local (1,830 msg/s/node remote vs 2,387 msg/s/node local). `rabbitmq_stream_s3_buffer_hit` was 97%.

The remote reader only issues a new range GET for a fragment when there is not already an outstanding request for the fragment, so fetches within a fragment are strictly serial. With one outstanding fetch, delivered bandwidth asymptotes to a single connection's transfer rate regardless of window size (`W / (TTFB + W/BW)` -> `BW` as `W` grows).

<details><summary>Defining that...</summary>

```
W :: prefetch window size
BW :: single connection transfer bandwidth
TTFB :: time to first byte

TTFB + W/BW :: total time for a fetch of size W
W / (TTFB + W/BW) -> BW :: delivered bandwidth asymptotes to connection transfer rate as W grows
```

</details>

On the testing workload: 143 ms per 5.18 MiB fetch (`rabbitmq_stream_s3_request_duration_seconds`), so 38 MB/s per connection, matching the measured `bytes_received` of 39-43 MB/s/node. A consumer draining faster than that catches the fill front every time the single in-flight fetch is in progress, producing a miss cadence set by the supply/demand race at the front rather than by window size. `adjust_read_size/2` halves the window on any miss and only grows it additively (+1 MiB) after a run of hits, so the window sawtooths near `read_size_min` and never accumulates enough hits to reach `read_size_max`. The knob AIMD controls (fetch size) is not the deficit (fetch concurrency). Remote reads therefore cannot reach local-tier throughput for a workload whose consumer demand exceeds one S3 connection's transfer rate.

We need to pipeline prefetches, increasing request concurrency. If prefetch is pipelined to depth N, `requests_in_flight` should rise toward N, per-reader `bytes_received` should scale roughly with N, `buffer_miss` should drop, and the AIMD window should grow past its current floor-hugging behavior toward `read_size_max`.

the-mikedavis:
*Issue #, if available:* fixes #349

*Description of changes:* the AIMD prefetch in the remote reader is not really effective at higher throughputs, see #349. We need to parallelize and pipeline the prefetches so that we are not bottlenecked by the throughput of a single S3 connection. Before this PR, there could only be a single fetch occurring against one fragment at one time, so max two connections (current and next).

Two operator-visible defaults change as a consequence:

- `general_pool_max_size` goes from 50 to 200. Each remote reader can now hold `prefetch_max_depth` (8) connections at once instead of 1-2, and the general pool also serves manifest, group and index GETs, so the old ceiling covered only a handful of lagging consumers per node. Connections are opened on demand and closed when idle, so the higher ceiling costs nothing at rest; exhausting it degrades a reader's prefetch depth rather than failing its reads
- `rabbitmq_stream_s3_read_size_bytes_bucket` is replaced by `rabbitmq_stream_s3_prefetch_window_bytes_bucket`. Panels and alerts naming the old metric go blank rather than error and have to be repointed by hand; the bundled Grafana dashboard already is


the-mikedavis:
I did some testing of this branch on an m7g.2xlarge 3 node cluster. Single consumer reading remote tier only, 64 KiB messages, no producer running during the consume phase. Same dataset (1M messages, about 65.5 GB) published identically before each test.

Delivered rate went from about 1,018 msg/s (66.7 MB/s) on main to about 3,781 – 4,440 msg/s (241 - 284 MB/s) on this PR. The Prometheus gauges confirm the mechanism: `requests_in_flight`  on main stays pinned at 1 for the whole run, since main never issues a second concurrent range GET. On this PR it rides between 4 and 8, reaching the configured depth ceiling repeatedly. `buffer_miss` rate dropped from about 3.1% to about 0.3%.

Node RSS during the consume window was roughly half on this PR versus main (about 127 MiB over baseline vs about 253 MiB), despite running three times faster. This is whole-node RSS, not an isolated measurement of the remote reader's own buffer, but it lines up with the design: the window is capped at 32 MiB plus one request, versus the old 64 MiB window plus a 64 MiB next-fragment prefetch.

the-mikedavis:
I also tried tuning `prefetch_max_depth` and `prefetch_max_window` on the same stream data and didn't see much difference past a depth of 10. At that point it reached nearly 5000 msg/s. Past that I think the consumer is CPU bound for this stream make-up: all 64KiB messages were in their own chunks, so per-chunk costs dominate. With a stream of larger chunks I would expect less CPU work.

lukebakken (approved):
Reviewed the prefetch pipeline end to end. Four things came out of it, all addressed in 246a264 on this branch:

- the prefetch window histogram's boundaries are computed at boot and frozen, but each reader read `prefetch_request_size` and `prefetch_window_max` from the app env when it started, so the two ends could disagree and every observation above the old ceiling landed in +Inf. Both now come from `prefetch_sizing/0`. This was already quietly defeating the two `log_reader_SUITE` tests that shrink the sizing per case to make ranges interleave
- `general_pool_max_size` 50 -> 200 and the `read_size_bytes_bucket` -> `prefetch_window_bytes_bucket` rename are operator visible, so I wrote both into the PR description
- `drop_prefetch/1` bypassed `checked/1`. It routes through it now, so every queue mutator does
- stale `fail_req/4` reference corrected to `fail_range/3`

eunit 253, remote_reader_core 87, log_reader 18, prop 51, all passing. Format check clean.

Not covered: a throughput test of the new prefetch depth against real S3. Concurrency changes of this shape fail in ways a diff read does not surface, so that is worth running before this carries production load.

the-mikedavis:
With 128 KiB messages I saw an unbounded consumer reach a consistent 330.25 MiB/s on this branch on the same m7g.xlarge, with next to no memory impact. CPU spiked to 50%-ish. The gain from 64KiB messages seems significant but not linear. I think there must be a balance point between per-chunk and per-byte costs on the consumption path as well. This build did not enable CRC32 checking on the consumption path (an option exposed by this plugin), but it would weigh into this balance.

the-mikedavis:
Similarly with 256 KiB messages I saw 1260 msg/sec (330,301,440 bytes/sec), so that seems to be the cap. Not sure what the bottleneck is at this throughput.

the-mikedavis:
Slight increase with 512 KiB messages: 746 msg/s (391,118,848 B/sec)

the-mikedavis:
347 msg/s for messages of size 1,000,000 B

PR #353, opened 2026-08-02
pool: Avoid warn logging when connections are down with normal/closed

the-mikedavis:
Gun does not end a stream if the server replies with a header requesting that gun close the connection, but no body. So a connection that handles a request like this emits a `{gun_down, ...}` message with one stream ref in the killed-streams list. This close is still normal, so there's no need to log it. It's just a quirk of Gun.

In a high throughput test I saw this log about 3,700 times over a 24 hour period.

### DIFF
--- a/src/rabbitmq_stream_s3_api_aws_pool.erl
+++ b/src/rabbitmq_stream_s3_api_aws_pool.erl
@@ -313,14 +313,15 @@ handle_gun_up(Conn, #?MODULE{close_fun = CloseFun} = State) ->
 %% arrives; a request fired onto a down connection is silently lost until it
 %% times out. `monitors` is left intact so the `DOWN` handler still runs the
 %% final cleanup and grows a replacement.
-%%
-%% Log abnormal closes with the connection's age and checkout status.
-%% Clean idle closes are not logged: reason=normal is gun's own clean
-%% shutdown; reason=closed is the remote side (S3) closing an idle
-%% keep-alive connection.
-handle_gun_down(Conn, Reason, _KilledStreams = [], State) when
-    Reason =:= normal orelse Reason =:= closed
-->
+handle_gun_down(Conn, normal, _KilledStreams, State) ->
+    %% Avoid a warning log even if `KilledStreams` is not empty. The current
+    %% version of gun in use today (2.4.0) does not clean up the HTTP/1 request
+    %% stream when the request's response has the header `connection: close`
+    %% and does not have a body, so the `gun_down` message reports a reason of
+    %% `normal` and a `KilledStreams` list of length one.
+    {noreply, remove_available(Conn, State)};
+handle_gun_down(Conn, closed, _KilledStreams = [], State) ->
+    %% Avoid a warning log if the remote closed an idle connection.
     {noreply, remove_available(Conn, State)};
 handle_gun_down(Conn, Reason, KilledStreams, #?MODULE{checkouts = Checkouts} = State) ->
     ConnAge = get_conn_age(Conn, State),


lukebakken:
Anything to report upstream for `gun`?

the-mikedavis:
Possibly yeah. I think there's one branch where we need to call an `end_stream` helper. But in that branch the entire connection is closing so it might be moot. I'll make a PR and see what Loïc thinks

src/rabbitmq_stream_s3_api_aws_pool.erl
```diff
@@ -318,7 +318,7 @@ handle_gun_up(Conn, #?MODULE{close_fun = CloseFun} = State) ->
 %% Clean idle closes are not logged: reason=normal is gun's own clean
 %% shutdown; reason=closed is the remote side (S3) closing an idle
 %% keep-alive connection.
-handle_gun_down(Conn, Reason, _KilledStreams = [], State) when
+handle_gun_down(Conn, Reason, _KilledStreams, State) when
```
lukebakken: The widened guard covers `reason=closed` too, and `closed` reaches `disconnect/2` from a remote socket FIN (gun.erl:1629) or a `setopts` failure (gun.erl:1748). With a non-empty killed-streams list that means S3 dropped a request that was still in flight, which is the abnormal case the `[]` pattern existed to separate from idle keepalive expiry.

After this change that case logs nothing, and nothing else logs it either: the requester's `stream_error` handling is `?LOG_DEBUG` (`rabbitmq_stream_s3_api_aws.erl:763`) and the retry is silent (`transfer_failed` -> `resubmit_transfer` bumps `transfer_retries` with no log, unlike the non-transient path at `replica_reader.erl:1049`). So a broker repeatedly losing in-flight uploads to connection resets would be invisible in the logs.

Widening only the `normal` case fixes the gun quirk without that loss:

```erlang
handle_gun_down(Conn, normal, _KilledStreams, State) ->
    {noreply, remove_available(Conn, State)};
handle_gun_down(Conn, closed, _KilledStreams = [], State) ->
    {noreply, remove_available(Conn, State)};
```

src/rabbitmq_stream_s3_api_aws_pool.erl
```diff
@@ -318,7 +318,7 @@ handle_gun_up(Conn, #?MODULE{close_fun = CloseFun} = State) ->
 %% Clean idle closes are not logged: reason=normal is gun's own clean
 %% shutdown; reason=closed is the remote side (S3) closing an idle
 %% keep-alive connection.
```
lukebakken: This comment describes `reason=closed` as "the remote side (S3) closing an *idle* keep-alive connection", but the word "idle" was carried by the `_KilledStreams = []` pattern that this diff removes. As written it now tells a future reader that a connection dropped mid-request is still logged at warn level.

The gun quirk motivating the change is also not recorded anywhere in the source. `AGENTS.md` asks for exactly that: "If you learn something surprising (a framework quirk, a type mismatch, a non-obvious ordering requirement), capture the learning in a comment." The PR body has the useful detail (gun leaves a completed stream in the killed-streams list when the server replies `fin` with `Connection: close`; ~3,700 occurrences over 24h) and it would be worth keeping next to the clause.

lukebakken (commented):
The premise checks out. In `gun_http:handle_response/8`, a `fin` response with `Connection: close` returns a bare `close` command with no `{state, ...}` update, so `end_stream/1` never runs and the completed stream is still in `streams` when `disconnect/2` calls `Protocol:down(ProtoState)`. Hence `reason=normal` with a one-element killed-streams list for a request that actually succeeded. I also confirmed `normal` can only reach the handler with a non-empty list via that one path: the other `close` emissions in `gun_http` either have no streams or call `end_stream/1` first.

Two comments inline. The substantive one is that widening the guard for `reason=closed` as well also silences genuine mid-request drops, which is the case the empty-list pattern was carrying.

Also noting that `handle_gun_down/4` has no direct test coverage: `api_aws_pool_statem_SUITE` kills fake connections outright and never feeds a `gun_down` message, so neither clause is reachable in the suite. Pre-existing, not made worse here.

lukebakken (approved):

the-mikedavis:
I opened up https://github.com/ninenines/gun/pull/374. There's also a `gun_error` message after the `gun_response` so I think this must be a bug rather than intentional.

Issue #356, opened 2026-08-04
Retention deleting segments faster than upload leaves the remote tier unreadable


lukebakken:
> [!NOTE]
> This issue was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who directed the investigation, the filing, and the assignees and milestone, but did not author the prose. The measurements come from Claude's test runs against a live 3-node cluster, and the code references were verified against the deployed `streams-tiered-storage-combined` branch. The analysis and the proposed direction are Claude's, and the framing has already been revised more than once as later test runs sharpened it (most recently on 2026-08-05, when an outage-isolation run showed runs 6 and 7 were the retention mechanism rather than outage recovery), so treat the conclusions as reviewable rather than settled.

## Summary

When local retention deletes whole segments at least as fast as the uploader consumes them, each deletion strands the segment backing the manifest's `next_offset`. The writer discards its remote manifest, restarts from the local log's first offset, and loses the new position the same way. Sustained, the remote tier stays unreadable for as long as the condition holds.

Observed on a 3-node `m7g.8xlarge` cluster (plugin at 6673cc3): 599 recovery cycles in 15 minutes, manifest unreadable in 27 of 27 samples, ending at `first_offset == next_offset` with `total_size = 0` and `entries = <<>>`.

## Mechanism

Retention is a `foldl` over the spec list in osiris `evaluate_retention0/2`, with the plugin's local fun prepended:

1. `rabbitmq_stream_s3_hooks:eval_local_retention/2` keeps the active segment plus every un-uploaded segment, and deletes fully-uploaded older ones.
2. The user's `max_bytes`, via osiris `eval_max_bytes/3` (`deps/osiris/src/osiris_log.erl:2357-2378`), then runs on those survivors. It is upload-unaware, deletes whole segments only, and unconditionally retains the latest:

```erlang
eval_max_bytes(IdxFiles, MaxSize) ->
    [Latest|Older] = lists:reverse(IdxFiles),
    eval_max_bytes(Older,
                   MaxSize - file_size_or_zero(
                               segment_from_index_file(Latest)),
                   [Latest]).
```

[...]

PR #359, opened 2026-08-05
Count the local-log-ahead recovery once and publish the reset range

### LINKED ISSUE #358: Recovery metrics: `local_log_ahead_recoveries` double-counts, manifest gauges keep the discarded range
> [!NOTE]
> This issue was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who directed the investigation, the filing and the assignees, but did not author the prose. Both defects were found by Claude while investigating #356; the gauge divergence was observed on a live 3-node cluster and every code reference was verified against the deployed checkout. Both fixes and the regression test are applied and passing locally. The analysis is Claude's, so treat it as reviewable rather than settled.

Two metric defects in the local-log-ahead recovery path, both found while investigating #356. Neither affects data movement; both corrupt the observability the plugin's own alerting and runbooks depend on.

## Defect 1: `local_log_ahead_recoveries` double-counts the upload-path recovery

A single upload-path recovery increments the counter twice, so `rabbitmq_stream_s3_local_log_ahead_recoveries` reads 2x the real event count on that path.

`handle_local_log_ahead/4` (`src/rabbitmq_stream_s3_replica_reader.erl`) increments the counter and then calls `resolve_and_start(reset_for_recovery(State0))`. That restart runs through `start_reading0/1`, which re-opens the data reader at the *same* stale `next_offset`, takes the same `{error, {offset_out_of_range, {LocalFirst, _}}}` when `LocalFirst > StartOffset`, and increments the counter again. Its own comment says so: "start_reading0/1 re-opens the data reader there, hits its offset_out_of_range branch (LocalFirst > StartOffset), and performs the manifest discard, the jump to LocalFirst, and the recovery-counter increment."

One logged recovery, one manifest discard, counter reads 2. Both increments happen inside a single reader callback, so an external observer never sees the intermediate 1.

This matters because the counter is the documented signal for distinguishing a benign one-shot recovery from the sustained loop in #356. `docs/operations.md:482` alerts on `rate(rabbitmq_stream_s3_local_log_ahead_recoveries[1h]) > 0`, and #356's proposed refinement is a *threshold* separating "a few cycles, recovered" from "sustained, tier is gone". Any threshold tuned against the current counter is tuned against inflated numbers, and #356's own measurements ("2 per run") are suspect on the upload path.

Introduced by f536999, which is an ancestor of the deployed 6673cc3. The reader-init path (`start_reading0/1`) already counted this recovery before that commit; f536999 added a second increment for the same event rather than a missing one.

Note the sibling path is correct: `?C_REMOTE_TIER_AHEAD_RECOVERIES` is incremented only in `start_reading0/1`, with no second increment elsewhere.

### Fix

Drop the increment in `handle_local_log_ahead/4` and let `start_reading0/1` be the single counting point for both recovery directions.

## Defect 2: `restart_at_local_floor/3` leaves the manifest gauges advertising the discarded range

`restart_at_local_floor/3` installs the fresh (empty) manifest, writes it to the local cache via `rabbitmq_stream_s3_manifest_replica:put_manifest/3`, and broadcasts it to replicas, but never calls `update_manifest_gauges/2`. `rabbitmq_stream_s3_manifest_first_offset` and `rabbitmq_stream_s3_manifest_next_offset` therefore keep reporting the range of the manifest that was just thrown away, until some later event happens to refresh them.

Both recovery directions reach this function (local-ahead and remote-tier-ahead), so both leave the gauges stale.

Observed on a 3-node `m7g.8xlarge` cluster after a leader restart: the gauges read `[129554, 194331)` while every node's manifest cache row read `[257045, 257045)`. Prometheus and Grafana advertised 64777 offsets of remote tier that no longer existed, contradicting both the cache row the same function had just written and what `get_range/1` returned.

`docs/operations.md:278-279` documents these gauges as "Offset of the oldest record in the remote tier" and "Next offset to upload to the remote tier". After a reset both are wrong, and wrong in the direction that makes a discarded range look recoverable. `docs/failure-modes.md:80` tells operators to read `rabbitmq_stream_s3_manifest_next_offset` to confirm a remote-tier-ahead recovery happened; the gauge does not move, so that check fails.

Introduced when `restart_at_local_floor/3` was factored out in af75dc3. `update_manifest_gauges/2` dates to 30a096c and is called from the two `on_manifest_resolved/2` clauses, which is why the gauges recover eventually but not at reset time.

### Fix

Call `update_manifest_gauges(FreshManifest, State)` in `restart_at_local_floor/3` after the `put_manifest` call.

## Test coverage

No test anywhere in `test/` referenced `local_log_ahead_recoveries`, which is why the double-count went unnoticed. `upload_path_recovers_from_trimmed_segment` in `test/replica_reader_SUITE.erl` already drives exactly the path that double-counted.

Asserting on a final total does not work: under that test's tight `max_bytes` retention, continued writing can legitimately drive a second genuine recovery, and the total was observed to vary between 1 and 2 across runs with the fix in place. The deterministic signal is the *first nonzero* reading, because a double increment happens entirely within one reader callback and so can never be observed as 1.

With that assertion the case fails 3/3 runs with the increment restored (reading 2 or 3) and passes 5/5 with it removed. Full suite: 39 cases, 0 failures.

## Environment

- Plugin at 6673cc3 (deployed), defects still present on `main` at 9d947b8
- Server `streams-tiered-storage-combined` at 0a1c4ad56a, osiris `tiered-storage-abstractions` at d810043
- OTP 27.3.4.13 (erts-15.2.7.9)
- 3-node cluster, `m7g.8xlarge`, us-west-2

lukebakken:
> [!NOTE]
> This PR description was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who directed the work and approved each commit but did not author this prose. Both defects were found by Claude while investigating #356; the gauge divergence was observed on a live 3-node cluster. The analysis and the test design are Claude's, so treat them as reviewable rather than settled.

*Issue #, if available:* fixes #358

*Description of changes:* two independent metric defects in the local-log-ahead recovery path, one commit each. Neither affects data movement; both corrupt the observability the plugin's own alerting and runbooks depend on.

**1afa13f counts the local-log-ahead recovery once instead of twice.** `handle_local_log_ahead/4` incremented `local_log_ahead_recoveries` and then called `resolve_and_start/1`. That restart runs through `start_reading0/1`, which re-opens the data reader at the same stale `next_offset`, takes the same `offset_out_of_range` error, and increments the counter again. One logged recovery and one manifest discard therefore read as 2. f536999 added this second increment for an event `start_reading0/1` already counted, so the fix drops it and leaves `start_reading0/1` as the single counting point, matching how `remote_tier_ahead_recoveries` already works.

This matters because the counter is the documented signal for telling a benign one-shot recovery from the sustained loop in #356. `docs/operations.md:482` alerts on its rate, and #356's proposed refinement is a threshold separating "a few cycles, recovered" from "sustained, tier is gone"; any threshold tuned against the current counter is tuned against inflated numbers.

**36aacc7 publishes the reset range from `restart_at_local_floor/3`.** That function installed the fresh manifest, wrote it to the local cache and broadcast it to replicas, but never called `update_manifest_gauges/2`, so `manifest_first_offset` and `manifest_next_offset` kept reporting the range of the manifest just discarded. Observed on a 3-node `m7g.8xlarge` cluster after a leader restart: the gauges read `[129554, 194331)` while every node's cache row read `[257045, 257045)`, so Prometheus advertised 64777 offsets of remote tier that no longer existed. Both recovery directions reach this function, so both left the gauges stale, and `docs/failure-modes.md:80` tells operators to read `manifest_next_offset` to confirm a recovery happened.

**Testing.** No test anywhere in `test/` referenced `local_log_ahead_recoveries`, which is why the double-count went unnoticed. `upload_path_recovers_from_trimmed_segment` already drives exactly that path, so the assertion goes there.

Asserting a final total does not work: under that test's tight `max_bytes` retention, continued writing can legitimately drive a second genuine recovery, and the total was observed varying between 1 and 2 across runs with the fix in place. The assertion is on the first nonzero reading instead, which is deterministic because a double increment happens entirely within one reader callback and so can never be observed as 1. With it, the case fails 3/3 runs with the increment restored and passes 5/5 with it removed.

Full `replica_reader_SUITE`: 39 cases, 0 failures. The first commit was verified in isolation so it stands on its own.

The gauge fix has no test. Reproducing it needs a leader restart against a live cluster, which this suite has no harness for; it was found by observation rather than by a failing case.

### DIFF
--- a/test/replica_reader_SUITE.erl
+++ b/test/replica_reader_SUITE.erl
@@ -805,7 +805,7 @@ upload_path_recovers_from_trimmed_segment(Config) ->
     %% a second recovery, so the total is not deterministic. The first nonzero
     %% reading is: a double increment happens entirely within one reader callback,
     %% so no external observer can ever see 1 when both fire.
-    ?assertEqual(1, await_first_nonzero(Config, local_log_ahead_recoveries, 30_000)),
+    ?assertEqual(1, await_first_nonzero(Config, local_log_ahead_recoveries, 5_000)),
 
     %% Keep writing so the reset reader is driven by osiris offset notifications
     %% to drain and re-upload (in #225 writes are continuous). No permanent
@@ -816,7 +816,7 @@ upload_path_recovers_from_trimmed_segment(Config) ->
             Write(5),
             get_range(Config)
         end,
-        30_000
+        5_000
     ).
 
 remote_tier_ahead_discards_manifest(Config) ->
--- a/test/replica_reader_SUITE.erl
+++ b/test/replica_reader_SUITE.erl
@@ -1626,7 +1626,8 @@ await_first_nonzero(Config, Name, Timeout, Start) ->
         _ ->
             Elapsed = erlang:monotonic_time(millisecond) - Start,
             case Elapsed > Timeout of
-                true -> ct:fail({counter_stayed_zero, Name, Timeout});
+                true ->
+                    ct:fail({counter_stayed_zero, Name, Timeout});
                 false ->
                     timer:sleep(5),
                     await_first_nonzero(Config, Name, Timeout, Start)
--- a/test/replica_reader_SUITE.erl
+++ b/test/replica_reader_SUITE.erl
@@ -792,6 +792,21 @@ upload_path_recovers_from_trimmed_segment(Config) ->
     %% stalling the manifest at NextA forever.
     ok = rabbitmq_stream_s3_api_fault:release(TaskPid, Ref),
 
+    %% This recovery counts exactly once. The upload path recovers by restarting
+    %% through resolve_and_start/1 -> start_reading0/1, whose local-ahead clause
+    %% re-opens at the same stale next_offset and does the counting; incrementing
+    %% on the upload path as well would report this single event as two. The count
+    %% is the documented signal for telling a benign one-shot recovery from a
+    %% sustained retention-outruns-upload loop (#356), so inflating it is not
+    %% cosmetic.
+    %%
+    %% Assert on the first nonzero reading rather than a final total: continued
+    %% writing under this test's tight max_bytes retention can legitimately drive
+    %% a second recovery, so the total is not deterministic. The first nonzero
+    %% reading is: a double increment happens entirely within one reader callback,
+    %% so no external observer can ever see 1 when both fire.
+    ?assertEqual(1, await_first_nonzero(Config, local_log_ahead_recoveries, 30_000)),
+
     %% Keep writing so the reset reader is driven by osiris offset notifications
     %% to drain and re-upload (in #225 writes are continuous). No permanent
     %% stall: the manifest leaves NextA and advances past it.
@@ -1586,3 +1601,34 @@ reconcile_recovers_replica_after_cache_restart(Config) ->
         end,
         5000
     ).
+
+%% Read one per-stream replica reader counter by name. This suite starts writers
+%% with `reference => StreamId` rather than a queue resource, so the reader
+%% registers under init_metrics/2's labelless fallback id.
+stream_counter(Config, Name) ->
+    StreamId = ?config(stream_id, Config),
+    Id = {rabbitmq_stream_s3_replica_reader, StreamId},
+    case seshat:counters(rabbitmq_stream_s3, Id, [Name]) of
+        #{Name := Value} -> Value;
+        undefined -> undefined
+    end.
+
+%% Poll a per-stream counter and return the first nonzero value observed. Used to
+%% assert the increment size of a single event, which a final total cannot pin
+%% down when the event can legitimately recur.
+await_first_nonzero(Config, Name, Timeout) ->
+    await_first_nonzero(Config, Name, Timeout, erlang:monotonic_time(millisecond)).
+
+await_first_nonzero(Config, Name, Timeout, Start) ->
+    case stream_counter(Config, Name) of
+        V when is_integer(V) andalso V > 0 ->
+            V;
+        _ ->
+            Elapsed = erlang:monotonic_time(millisecond) - Start,
+            case Elapsed > Timeout of
+                true -> ct:fail({counter_stayed_zero, Name, Timeout});
+                false ->
+                    timer:sleep(5),
+                    await_first_nonzero(Config, Name, Timeout, Start)
+            end
+    end.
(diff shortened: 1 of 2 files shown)


lukebakken:
Verified live on a 3-node cluster (source build, `fix/358-recovery-metrics` @ `0caf689`) with a new regression test, `gauge-reset-test.sh`.

The gauge staleness needs a live reproduction because the plugin self-heals: under normal operation local retention is pinned at the durability watermark, so the local log is only transiently ahead and any reset is immediately re-advertised by a subsequent successful upload. The test forces the condition deterministically by publishing under a #356 local-log-ahead workload (10mb window, default segments) and then **denying S3 on the writer's IAM role**, so the only manifest writes are resets and the gauge cannot be refreshed by anything else.

- **Buggy build (`6673cc3`):** after the reset the cache row read `[943193, 943193)` while the per-stream gauge still advertised the discarded range `[471587, 471587)` -> test fails.
- **This branch:** 3 resets fired under the outage (recovery counter `0 -> 3`), and afterward cache row == gauge `[1414920, 1414920)` -> test passes.

Confirmed the fix from the *loaded* module's abstract code on all three nodes, not just source on disk: `restart_at_local_floor/3` now calls `update_manifest_gauges`.

test/replica_reader_SUITE.erl
```diff
@@ -792,6 +792,21 @@ upload_path_recovers_from_trimmed_segment(Config) ->
...
+    %% is the documented signal for telling a benign one-shot recovery from a
+    %% sustained retention-outruns-upload loop (#356), so inflating it is not
+    %% cosmetic.
+    %%
+    %% Assert on the first nonzero reading rather than a final total: continued
+    %% writing under this test's tight max_bytes retention can legitimately drive
+    %% a second recovery, so the total is not deterministic. The first nonzero
+    %% reading is: a double increment happens entirely within one reader callback,
+    %% so no external observer can ever see 1 when both fire.
+    ?assertEqual(1, await_first_nonzero(Config, local_log_ahead_recoveries, 30_000)),
```
the-mikedavis: 30 seconds seems too long for this suite. Tests should always take less than a second since there aren't many moving parts. I think we usually use 1s or 5s in this suite

lukebakken: Good call. Both 30s polls in this case are now 5s: the counter wait you flagged and the range-advance awaitMatch right below it. The case runs in ~185ms, so 5s is ample headroom.

PR #360, opened 2026-08-13
Correct and tidy the plugin documentation set

lukebakken:
> [!NOTE]
> This PR description was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before it was posted. The documentation changes were verified against the source on this branch.

Documentation-only pass over the plugin's Markdown, in three layers.

Mechanical cleanup:
- Replace em/en dashes with ASCII hyphens throughout
- Unwrap hard-wrapped prose so each paragraph is a single line

Factual corrections (verified against source):
- Manifest full-root capacity is 64 EiB, not ~70 EiB
- Byte-size figures and unit labels in scale.md and manifest.md
- The chunk header is a fixed 48 bytes; the bloom filter follows it and is absent when a chunk carries no filter values (log-framing.md, conventions.md, GLOSSARY.md)
- Stream ID leading-underscore count, status section count, and the root manifest key diagram
- Drop the stale "(planned)" note on orphan GC and the retired read_size_max prefetch ceiling
- Fix a broken docs link, the P model checker flag, and make -> gmake
- Correct the Tier overlap invariant's local range

Source-cross-check of architecture.md, upload-path.md, and read-path.md against the modules they describe:
- Add the two supervision-tree workers that were missing (bucket monitor, GC scheduler) and correct the reconciler's start-order note
- All other checked claims (config defaults, byte layouts, effect names, function signatures) matched source and were left as-is

Flagged, not fixed: two figures in the cross-AZ investigation are internally inconsistent and are marked "Figure to re-verify" pending the raw run data, which is not in the repo.


docs/DEVELOPMENT.md
```diff
@@ -14,9 +14,10 @@ This plugin lives inside the RabbitMQ server tree as a dependency. You do not cl
 ```bash
 git clone https://github.com/amazon-mq/upstream-to-rabbitmq-server.git
 cd upstream-to-rabbitmq-server
-git checkout streams-tiered-storage
 ```
 
+The plugin is on the default branch, `main`, which is what the clone checks out.
+
```
the-mikedavis, line 20: I think this part is not correct, there is https://github.com/amazon-mq/upstream-to-rabbitmq-server/tree/streams-tiered-storage which sets up the plugin

lukebakken: Good catch, fixed in 02b158c803fb184673adccff793f113f1fdb8947

lukebakken:
A second high-effort review gave this the 👍 - merging!

PR #363, opened 2026-08-14
Re-sync a replica whose manifest row is wedged in `pending`

### LINKED ISSUE #362: Replica manifest row can wedge in `pending` after a member-restart burst, permanently failing offset-`first` subscribes on one node
> [!NOTE]
> This issue was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who directed the investigation, the filing and the assignees, but did not author the prose. The logs and state come from Claude's test run against a live 3-node cluster, and every code reference was verified byte-for-byte against the deployed checkout (server `streams-tiered-storage-combined`, osiris `tiered-storage-combined`). The reconciler backstop gap was confirmed directly on the cluster; the end-to-end race that produces the wedge was not reproduced (the wedged row was cleared before it could be frozen), and the issue says so where it matters. The analysis and the proposed fixes are Claude's, so treat them as reviewable rather than settled.

## Summary

A replica's manifest cache row can wedge in `pending` indefinitely on one node. Every offset-`first` subscribe to that stream on that node then fails closed with `?RESPONSE_CODE_STREAM_NOT_AVAILABLE`, while the stream is otherwise healthy (`state=running`, 3 members, `lookup_local_member/2` returns a live pid). Nothing recovers it without a member restart or a manual resync. Observed on a 3-node cluster: one stream stayed wedged on node0 for over an hour, until a hand-sent resync resolved it in 8 seconds.

The trigger is a burst of osiris member restarts (the same "fell behind retention, resyncing" churn as #357). Every automatic path that would resolve the row is defeated across the burst, and once publishing stops there is nothing left to trigger a resync.

## Symptom

Subscribing to one stream failed with `StreamNotAvailableException`, but only when the client landed on node0. Isolated by pinning the client to each node:

| Node | 1 consumer | 2 consumers |
|------|------------|-------------|
| node0 `10.0.1.31` | fail | fail |
| node1 `10.0.1.243` | ok | fail |
| node2 `10.0.1.61` (writer) | ok | ok |

node0 failed at a single consumer, so it is not a concurrency effect; node1 failing only at two is the client spreading subscriptions and one landing on node0. node0 served every other stream fine, so the broken unit is one (node, stream) pair.

`rabbitmq_stream_s3_log_reader:init_offset_reader/2` fails closed when the cache row is `pending`, returning `{error, unavailable}`, which `rabbit_stream_reader` turns into `?RESPONSE_CODE_STREAM_NOT_AVAILABLE`. That is correct and deliberate: the remote tier's extent is unknown, so a spec that may live there must not be served from local data alone. The source comment says the consumer's retry "lands after resolution (normally milliseconds after member start)". The row never resolved.

## Root cause

The cache row starts `pending` and is resolved only by a `sync` from the writer landing in `maybe_apply_sync/8`, which seeds `seqs` and writes the manifest. The writer sends a sync in exactly three places: initial `register_acceptor`, an explicit `resync` request, and the periodic `reconcile_replicas`. For this stream on node0, every one was defeated.

State on node0 while wedged (read from the live process):

- `contexts` had the stream (a live member context was registered)
- `seqs` did not (no sync ever applied)
- `pending_resync` was empty (nothing would request a resync)
- the cache row was `{StreamId, pending, undefined}`, still pending after an hour

### The verified sequence

The initial sync was dropped because it raced ahead of context registration (node0 log, member start):

```
14:03:32.535764  dropped sync (epoch 1 seq 0 from 'rabbit@ip-10-0-1-61')
                 with no live reader context
14:03:32.538054  registered a context with a sync pending; requesting a
                 re-sync from writer node 'rabbit@ip-10-0-1-61'
```

`maybe_apply_sync/8` drops a sync whose context is not yet registered, recording the writer in `pending_resync`; context registration then calls `maybe_request_resync/2`, which sends one resync and clears `pending_resync`. This is the designed recovery and it usually works: the same drop-then-resync pattern appears on another stream in the same run and recovered in ~12s. (The writer's `core` was not undefined here: it logged `start_reading start_offset=0` at 14:03:32.535, before the first resync at .538, so the `core = undefined` resync-drop clause at `replica_reader.erl:503` was not involved.)

What wedged this stream is a burst of osiris member restarts. Under the firehose the replica repeatedly fell behind retention and resync-truncated (the #357 condition), restarting the local member 7 times between 14:05:27 and 14:06:57. Each restart's DOWN runs `release_stream/2`, which deletes the ETS row and clears both `seqs` and `pending_resync`; each re-registration recreates a fresh `pending` row.

Through the churn node0 kept requesting resyncs, but on gap detection (every broadcast edit arriving with `seqs = undefined` hits the catch-all gap clause and calls `request_resync`), not because the row was pending. Those fired 6 times. None resolved the row, and `syncs_rejected` stayed 0 cluster-wide, so no resync-sync was rejected as stale: the syncs simply did not seed `seqs` before the next restart wiped it.

The trap closes on the last restart:

```
14:06:57.806  fell behind retention ... terminating   (row + seqs + pending_resync cleared)
14:06:57.910  osiris replica starting ...              (row recreated pending)
              register_replica_context: pending_resync empty -> maybe_request_resync sends nothing
              register_acceptor cast to writer -> no-op (node0 already in writer's replicas map)
14:06:33      last resync request (BEFORE the final restart)
14:19:48      consumer attach still fails closed on the pending row (13 min later)
```

After 14:06:57 the stream stabilized, so no more edits arrived to gap-detect and no more resyncs were requested. The final re-registration could not request one either: `pending_resync` had just been cleared by the preceding `release_stream`, so `maybe_request_resync/2` had nothing to act on, and the `register_acceptor` cast it also sends (`hooks.erl:88`) is a no-op at the writer because node0 was still in the writer's `replicas` map. The writer monitors node0's `manifest_replica` process, which never died, only the osiris member churned, so the entry persisted. The row was left `pending` with nothing that would ever request the sync to resolve it.

### The reconciler cannot repair it

`reconcile_replicas` runs every 60s (`reconciliation_interval`) but is a no-op for this state on both sides:

- Writer (node2): reader attached, so it casts `reconcile_replicas`, which calls `register_replicas/2`. For node0 (already in `replicas`), `register_replica/2` hits `maps:is_key(Node, Replicas) -> true -> State` and sends no sync (`replica_reader.erl:822`).
- Replica (node0): `reconcile_child` for an `osiris_replica` checks `is_context_registered(StreamId)`, which returned `true`, so it does nothing. The check inspects context presence only; it never looks at whether the row is still `pending` (`hooks.erl:247`).

I confirmed the gap directly on the cluster: forced the wedged shape on a replica node via `sys:replace_state` (row -> `pending`, `seqs`/`pending_resync` entries removed, context left registered), ran a full `rabbitmq_stream_s3_hooks:reconcile/0` tick, and the row stayed `pending`. A single manual `{resync, node()}` cast to the writer then resolved it immediately (and I restored the stream that way).

## Why the existing alerting missed it

`resolve_failed_closed` counted the fail-closed attaches correctly (22 on node0), but `manifest_resolution_failures` stayed 0 on all three nodes. Its source comment names it as the alertable signal for a stuck row, but it only increments in the writer's own `resolve_and_start` retry loop (`replica_reader.erl:1446`), a writer that cannot resolve its manifest from the store. Here the writer resolved fine; the failure is entirely replica-side, where a resolved manifest never arrived. No counter tracks "row has been `pending` too long", so the one state that matters is unmeasured.

## Suggested fixes

In rough order of preference:

1. Request a resync when re-registering a context onto a still-`pending` row. `register_replica_context` recreates the row `pending` on every member restart but only requests a resync if `pending_resync` holds the stream, which the preceding `release_stream` just cleared. A `pending` row with no `seqs` entry has no baseline and must resync; request one unconditionally in that case rather than relying on `pending_resync`. This closes the observed trap.
2. Make the reconciler verify the row, not just the context. Have the replica-side `reconcile_child` also check whether the row is `pending` (with no `seqs` entry) past a grace period and request a resync if so. A cheap standing backstop for the whole class; the writer's `resync` handler already does the right thing on receipt.
3. Have the writer's `register_acceptor` re-sync an already-registered node instead of early-returning. Re-sending the persisted manifest is cheap and idempotent (the replica's staleness check drops it if not needed).

Option 1 plus option 2 is the right pair. Whatever the fix, close the observability gap too: a row that stays `pending` with a registered context past a grace period should increment a counter, so this state is detectable rather than silent.

## Reproducing

Not yet reproduced end-to-end; the wedged row was cleared before it could be frozen, and the trigger is a timing race. Established from source and logs: the row wedges when a member-restart burst leaves the final re-registration with `pending_resync` cleared and `register_acceptor` a no-op, then no further edits arrive to gap-detect. The realistic trigger is a replica repeatedly falling behind retention under a firehose (the #357 restarts). To attempt it: many streams, heavy publish, let a replica resync-churn, then stop publishing on that stream and attempt an offset-`first` subscribe on the replica node.

A direct test of fix 2, without the race: put a stream's cache row into `{pending, undefined}` with a registered context and empty `pending_resync`, trigger a reconcile tick, and assert the row resolves. This is the check I ran manually above and would make a good regression guard.

## Environment

- Server: `streams-tiered-storage-combined` (`amazon-mq/upstream-to-rabbitmq-server`), commit 6d508eacdc
- Plugin: commit 13f7ca0
- osiris: `tiered-storage-combined` (`amazon-mq/upstream-to-osiris`), commit 2735cdd
- Erlang/OTP 27 (erts 15.2.7.9)
- Cluster: 3 x `m7g.8xlarge`, one `m7g.16xlarge` load generator, source mode
- Wedge observed 2026-08-13, stream created 14:03, still failing at 14:19
- Broker log level: debug

lukebakken:
> [!NOTE]
> This PR description was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who directed the work and approved each commit but did not author this prose. The bug was found by Claude building a lagging-consumer fan-out test; the mechanism was traced against the deployed checkout and verified live on a 3-node cluster, both by direct manipulation and under real load. The analysis and the test are Claude's, so treat them as reviewable rather than settled.

*Issue #, if available:* fixes #362

*Description of changes:* a replica's manifest cache row can wedge in `pending` indefinitely on one node, so every offset-`first` subscribe to that stream on that node fails closed with `STREAM_NOT_AVAILABLE` while the stream is otherwise healthy. Nothing recovers it without a member restart or a manual re-sync; one stream stayed wedged for over an hour.

The row is resolved only by a sync from the writer landing in `maybe_apply_sync/8`. The writer sends one from `register_acceptor`, from an explicit `resync` request, and from `reconcile_replicas`. A member-restart burst (the "fell behind retention" resync churn) defeats all three: each restart's `release_stream` clears the row, `seqs` and `pending_resync`, and the final re-registration cannot request a sync (`pending_resync` was just cleared) while `register_acceptor` is a no-op at the writer because the node never left the writer's `replicas` map. The writer monitors the node's `manifest_replica` process, which does not restart, not the osiris member, which does. Once publishing stops, no edits arrive to gap-detect and request a sync, so the row stays `pending`.

**The cause fix (`rabbitmq_stream_s3_replica_reader`).** The writer's `register_acceptor` handler now re-syncs the persisted manifest even when the node is already registered, instead of early-returning. Every member restart re-fires `on_init(acceptor)`, which re-sends `register_acceptor`, so the writer heals the row on the next restart. The re-sync is idempotent: the replica's `is_stale_sync/3` drops it if the replica is already current. Both `sync` and `apply_edits` are cast from the same single writer-side reader process and read the same monotonic persisted manifest, so the extra sync cannot race or invert a broadcast into divergence.

**The backstop (`rabbitmq_stream_s3_hooks` + `rabbitmq_stream_s3_manifest_replica`).** The periodic reconciler previously did nothing for a context-registered replica. It now re-requests a sync when the row is still `pending`, via a new `is_awaiting_sync/1` that reads the ETS row state rather than the `seqs` map: `put_manifest/2,3` (GC, the read path, writer reseed) resolves a row without recording a sequence, so a `seqs`-absence check would false-positive on healthy rows.

*Testing:* added `is_awaiting_sync_test_`, covering the wedged shape, resolution by sync, and the `put_manifest` resolve that must not read as awaiting. It has teeth: with the earlier `seqs`-based predicate the `put_manifest` case fails; with the row-state predicate all cases pass. Full module eunit: 8 cases, 0 failures.

Verified live on a 3-node cluster. Forced the wedged shape on a replica and confirmed both recovery paths resolve it: the reconciler tick (`reconcile/0` -> `is_awaiting_sync` -> re-request) and a bare `register_acceptor` to the writer (what a member restart sends), which on unpatched code early-returned and left the row wedged. Then ran the lagging-consumer fan-out test for 30 minutes: roughly 490 "fell behind retention" resync-truncations per node (the wedge trigger, at far higher volume than the run that first wedged a stream) produced zero pending rows on all three nodes and zero broker errors.

### DIFF
--- a/src/rabbitmq_stream_s3_hooks.erl
+++ b/src/rabbitmq_stream_s3_hooks.erl
@@ -249,7 +249,15 @@ reconcile_child({_Id, Pid, worker, [osiris_replica]}) when is_pid(Pid) ->
         StreamId = stream_id_of(Pid),
         case rabbitmq_stream_s3_manifest_replica:is_context_registered(StreamId) of
             true ->
-                ok;
+                %% Context is registered, but the row may still be wedged in
+                %% pending with no applied sync (rabbitmq_stream_s3#362): a
+                %% member-restart burst can leave the final re-registration
+                %% unable to request a sync, and once publishing stops no edits
+                %% arrive to gap-detect and re-request one. Re-request a sync
+                %% from the writer when the row is unresolved. The writer's
+                %% register_acceptor handler re-syncs idempotently, so this is a
+                %% no-op once the row resolves.
+                maybe_resync_pending_replica(Pid, StreamId);
             false ->
                 ?LOG_INFO(
                     "Reconciliation: replica ~p for stream ~ts has no manifest "
@@ -286,6 +294,37 @@ register_replica_context(Pid, StreamId) ->
     rabbitmq_stream_s3_manifest_replica:register_replica_context(
         StreamId, Pid, Dir, Shared, Counter
     ).
+
+%% Re-request a manifest sync for a replica whose context is registered but
+%% whose cache row is still pending with no applied sync (rabbitmq_stream_s3#362).
+%% Sends {register_acceptor, node()} to the writer, whose handler re-syncs the
+%% persisted manifest idempotently. A resolved row reports not-awaiting, so this
+%% is a no-op on the steady-state ticks. The writer node comes from the queue
+%% record, the same source discover-time re-registration uses.
+maybe_resync_pending_replica(Pid, StreamId) ->
+    case rabbitmq_stream_s3_manifest_replica:is_awaiting_sync(StreamId) of
+        false ->
+            ok;
+        true ->
+            #{reference := Reference} = osiris_util:get_reader_context(Pid),
+            case rabbit_amqqueue:lookup(Reference) of
+                {ok, Q} ->
+                    WriterNode = maps:get(leader_node, amqqueue:get_type_state(Q)),
+                    ?LOG_INFO(
+                        "Reconciliation: replica for stream ~ts is registered but "
+                        "still awaiting its first sync; re-requesting from writer "
+                        "node ~p",
+                        [StreamId, WriterNode],
+                        #{domain => ?RMQLOG_DOMAIN_STREAM_S3}
+                    ),
+                    gen_server:cast(
+                        {via, rabbitmq_stream_s3_registry, {StreamId, WriterNode}},
+                        {register_acceptor, node()}
+                    );
+                _ ->
+                    ok
+            end
+    end.
 
 append_retention(StreamId, Config) ->
     Fun = {'fun', local_retention_fun(StreamId)},
--- a/src/rabbitmq_stream_s3_hooks.erl
+++ b/src/rabbitmq_stream_s3_hooks.erl
@@ -270,7 +270,7 @@ reconcile_child({_Id, Pid, worker, [osiris_replica]}) when is_pid(Pid) ->
     catch
         Class:Reason ->
             ?LOG_WARNING(
-                "Reconciliation could not re-register replica ~p: ~ts:~p",
+                "Reconciliation could not reconcile replica ~p: ~ts:~p",
                 [Pid, Class, Reason],
                 #{domain => ?RMQLOG_DOMAIN_STREAM_S3}
             )
@@ -307,23 +307,37 @@ maybe_resync_pending_replica(Pid, StreamId) ->
             ok;
         true ->
             #{reference := Reference} = osiris_util:get_reader_context(Pid),
-            case rabbit_amqqueue:lookup(Reference) of
-                {ok, Q} ->
-                    WriterNode = maps:get(leader_node, amqqueue:get_type_state(Q)),
+            case register_with_writer(StreamId, Reference) of
+                {ok, WriterNode} ->
                     ?LOG_INFO(
                         "Reconciliation: replica for stream ~ts is registered but "
-                        "still awaiting its first sync; re-requesting from writer "
-                        "node ~p",
+                        "still awaiting its first sync; re-requested a sync from "
+                        "writer node ~p",
                         [StreamId, WriterNode],
                         #{domain => ?RMQLOG_DOMAIN_STREAM_S3}
-                    ),
-                    gen_server:cast(
-                        {via, rabbitmq_stream_s3_registry, {StreamId, WriterNode}},
-                        {register_acceptor, node()}
                     );
-                _ ->
+                error ->
                     ok
             end
+    end.
+
+%% Look up the stream's writer node from the queue record and register this
+%% node as an acceptor with the writer's replica reader for manifest broadcast.
+%% The writer's register_acceptor handler re-syncs the persisted manifest
+%% idempotently. Returns {ok, WriterNode} so callers can log, or error when the
+%% queue record is unavailable (the writer's replica reader isn't up yet; the
+%% cast is skipped and the replica reader will proactively sync us on startup).
+register_with_writer(StreamId, Reference) ->
+    case rabbit_amqqueue:lookup(Reference) of
+        {ok, Q} ->
+            WriterNode = maps:get(leader_node, amqqueue:get_type_state(Q)),
+            gen_server:cast(
+                {via, rabbitmq_stream_s3_registry, {StreamId, WriterNode}},
+                {register_acceptor, node()}
+            ),
+            {ok, WriterNode};
+        _ ->
+            error
     end.
 
 append_retention(StreamId, Config) ->
@@ -412,18 +426,9 @@ attach_replica(Pid) ->
         StreamId, Pid, Dir, Shared, Counter
     ),
     %% Register with the writer's replica reader for manifest broadcast.
-    %% If the writer's replica reader isn't up yet, the cast is dropped;
+    %% If the writer's replica reader isn't up yet, the cast is skipped;
     %% the replica reader will proactively sync us on its startup.
-    case rabbit_amqqueue:lookup(Reference) of
-        {ok, Q} ->
-            WriterNode = maps:get(leader_node, amqqueue:get_type_state(Q)),
-            gen_server:cast(
-                {via, rabbitmq_stream_s3_registry, {StreamId, WriterNode}},
-                {register_acceptor, node()}
-            );
-        _ ->
-            ok
-    end,
+    _ = register_with_writer(StreamId, Reference),
     %% Inject the local-tier retention fun into the running replica while
     %% preserving its configured policy, via the same current-spec transform
     %% as attach_writer/1.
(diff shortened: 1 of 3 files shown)


lukebakken:
> [!NOTE]
> This comment was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken. The verification runs were executed on a live 3-node test cluster and the results are Claude's measurements from those runs.

Live-verified this fix on a 3-node cluster (source build at the commits on this branch).

Deterministic backstop test. Injected the exact wedged shape from the mechanism writeup onto a replica of a healthy remote-tiered stream: context registered, `seqs` entry cleared, `pending_resync` empty, ETS row forced to `{StreamId, pending, undefined}`. Confirmed `is_awaiting_sync/1` flips to `true` on that shape, then verified two resolution paths:

- Manual `reconcile/0`: row resolved in under 1s.
- Periodic 60s reconcile timer (no manual trigger): row stayed `pending` through the 10-50s samples and resolved exactly at the 60s tick.

The negative control is the important part: resolution tracked the reconciliation interval precisely rather than clearing immediately, and afterward `seqs` was re-seeded with `{0,1,<writer-node>}`, proving a real writer sync landed and was applied through `is_awaiting_sync` -> `maybe_resync_pending_replica` -> `register_with_writer`, not just an ETS poke.

Corroborating soak. A separate 15-minute retention-stress run (wide window filling the remote tier, then a mid-run drop to 400mb with a slow offset-`first` consumer pinned far back) held zero pending rows across all nodes for the entire run, sampled every 60s, while retention actively trimmed and the reaper deleted S3 objects underneath the consumer.

No code change from this; recording the verification.

the-mikedavis (approved):
This change is small and a nice correction; let's get it merged. I'm worried about this replication protocol in general, and I'm exploring ways to improve it. I want to do that as follow-ups though.

At a high level, the replication is balancing correctness (a manifest replica is an eventually consistent view of what the writer has persisted) with liveness (manifest replica is available to guide local retention and remote-tier readers). I think that it isn't currently set up to win on both, but I need to think through it more, and probably the models and property tests need to be more cohesive.

PR #365, opened 2026-08-21
Add defense-in-depth config options for object store access

the-mikedavis:
This is two new config options for extra defense-in-depth:

* `stream_s3.account_id` - when set, adds the `x-amz-expected-bucket-owner` to that account ID. Avoids any sort of nonsense happening if the bucket name is typo'd or recycled.
* `stream_s3.allow_static_credentials` - allows setting `stream_s3.access_key_id` and `stream_s3.secret_access_key` and defaults to false; discourages using static credentials stronger than before.


lukebakken (approved):
> [!NOTE]
> This review was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken. The findings below were verified against the branch source and eunit runs; the approval decision is @lukebakken's.

The defense-in-depth config work (`s3.account_id` / `x-amz-expected-bucket-owner` and ignore-static-credentials-by-default) looks good. The expected-bucket-owner header is applied at the single `sign_headers/11` chokepoint after the header merge, so a caller cannot override it, and it is correctly omitted when `account_id` is undefined.

A code review of the branch surfaced five items, all addressed in follow-up commits on this branch:

- Partial static credentials (only one of access_key_id / secret_key set) crashed `resolve_credentials_source/0` with a case_clause at init; it now falls back to managed credentials, with a regression test.
- Config reload did not evict a stale static credentials row from ETS; the managed branch now deletes it.
- Documented why the per-request account_id lookup is cheap (a single ac_tab ets:lookup_element, not a gen_server call).
- Test hygiene: `end_per_group` now unsets the app env it set, and the opt-in test pins the managed source deterministically instead of sampling the ambient environment.

PR #367, opened 2026-08-27
S3: Add configuration for KMS encryption context

the-mikedavis:
This is optional additional config that sets the header

    x-amz-server-side-encryption-context

This adds extra metadata that can be easily queried in CloudTrail.


lukebakken (commented):
> [!NOTE]
> This review was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who ran the review, decided which findings to act on and how, and approved the commits. Code and AWS documentation references were verified against the branch.

Ran a review pass over this branch. Five candidate findings: one refuted, four addressed in 1e8f66d, 84c8ab5, and 359bc93.

**Addressed**

- Empty KMS encryption context value: `stream_s3.kms_encryption_context.foo =` was accepted as an empty value. KMS rejects an empty context value, and since the same context is sent on every `GenerateDataKey` call that would stop all uploads. Added a non-empty schema validator so it fails at boot (84c8ab5).
- Duplicated key parsing: `key_stream_id/1` and `rabbitmq_stream_s3_gc:parse_key/1` both decoded the `rabbitmq/stream/<id>/` layout. Extracted `rabbitmq_stream_s3:split_key/1` as the single decode site; pure refactor (1e8f66d).
- Per-PUT recompute of the context: left as-is (the encode is negligible next to a whole-object upload) with a comment saying why (359bc93).
- Unreachable `cuttlefish:invalid` clause in the translation: kept as a guard against a future mapping change that widens what matches, now commented so it is not removed as dead code (84c8ab5).

**Refuted**

- A non-UTF-8 stream ID was flagged as able to crash `rabbit_json:encode/1` on the KMS path. It cannot: the stream ID is the osiris stream name, produced through `osiris_util:to_base64uri/1`, so it holds only `[A-Za-z0-9_-]` even though the AMQP queue name it derives from is not required to be valid UTF-8 (`rabbit_channel:check_name/2` does not enforce it). Documented at the encode site (359bc93).

**Docs**

- Verified the `operations.md` claim that S3 appends its own `aws:s3:arn` context pair on top of the one supplied, against the S3 documentation; it is correct. Added a citation (359bc93).

PR #369, opened 2026-09-02
Optimize remote reader throughput

the-mikedavis:
This is a pass at optimizing the max throughput of the remote reader when a subscription accepts messages as fast as possible. The goal we're chasing here is 5 Gbps (~596 MiB/s) for a single subscription. That's the limit of a single flow (i.e. connection) between two EC2 instances which are not in a cluster placement group. `iperf3` confirms this 5 Gbps limit experimentally in my tests. This branch reaches this, approaching 590 MiB/s in terms of message bodies. The remaining few percent are eaten up by framing. This matches local tier consumption.

There are a bunch of changes here. Headlines:

### Multi-fragment prefetch

To reach the desired throughput ceiling, a consumer would need to read through multiple fragments within a second - maybe up to 10 even. The old remote reader prefetched within the _current_ and _next_ fragments, but not beyond. The first optimization change is to prefetch across multiple fragments freely, based on the demand the consumer is exerting, rather than limit to two.

### Switch control signal from byte window to concurrency

S3 connections seem to deliver data at a roughly fixed rate. The old AIMD growth mechanism used to grow the request size. This doesn't really move the needle though. When delivering MiB of data, the latency doesn't show up against the throughput. This change switches the growth from bytes to concurrent requests. Concurrent requests at a fixed throughput stack and deliver a higher total throughput.

The byte window ends up with nothing to do and is deleted outright later in the branch. Request size is now fixed, and what adapts is how many requests run at once.

### Move time into the reader core

The reader core is meant to be totally pure. The shell therefore cared about time instead of the core. We can thread time as a variable through the core so that it can measure against the delivered supply (bytes delivered per 200ms period). This allows us to test time through the core, and tune minutes of virtual time in sub-second tests.

### Auto-tune inflight request target

This one even has graphics!

```
throughput
    │        ╭────────────
    │      ╭─╯
    │    ╭─╯
    │  ╭─╯
    │╭─╯
    └──────────────────────────  concurrency
      8   16   24  32  48   64
```

Using extra requests to S3 raises the total throughput initially as we stack up the number of bytes delivered. Eventually it's no longer fruitful. Something else becomes the limit - the flow, the store, or the consumer - and the extra requests just queue up behind it. The idea is to measure a period (200ms) of activity and determine if throughput increased or decreased. Based on this we can hill-climb to the optimal throughput given what S3 is delivering in the last set of periods. We double until doubling stops paying, then step by eighths around the best rate we recorded, turning back when things get worse, and continuing when things look flat.

### Split fetch and buffering budgets

The reader holds bytes in two places: ranges it asked S3 for and hasn't received, and bytes it received that the consumer hasn't read. Both cost memory, or soon will. One budget covering both doesn't work. Delivered-bytes move from the first to the second and sit there until the consumer reads them. So the buffer wins the budget over time and fetching gets the leftovers. That's worst when the consumer is slow, which is when we least want to stop fetching. So the budget is split down the middle. Fetching gets half and doesn't consult the buffer at all. A slow consumer can still stop the reader, but only by filling the whole budget.

### Scale the budget with the request size

The budget is in bytes, but it gets spent on requests. At a fixed size, doubling `prefetch_request_size` halves how many fit, and concurrency halves with it. So raising the request size lost throughput:

| request size | before | after |
|---|---|---|
| 4 MiB | 552-568 MiB/s | 587.9 |
| 8 MiB | 502 | 588.7 |
| 16 MiB | 365 | 588.8 |

The default is now 64 requests' worth rather than a fixed number of bytes, so it grows with the request size. Bigger requests cost memory now, instead of throughput.

### Tune defaults

Max-depth (resource cap) increases 8->64, and the memory bound 32MiB->256MiB. The in-flight target isn't configured any more: it starts at 1 and the search finds it. `prefetch_inflight_initial` (32) only applies with `prefetch_auto_tune` off, which is there as a way back and for reproducible measurements. This makes for a higher throughput ceiling. It only works really effectively when stacked on the changes above.

One thing to weigh: a lagging reader can now hold 256 MiB where it used to hold about 64. That is per consumer reading from S3, and only a consumer reading at full speed gets there.

These settings are in `rabbitmq.conf` now. Before they were app env only.

---


src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -1184,6 +1422,196 @@ decay(#state{cfg = #cfg{request_size = RequestSize}, window = Window} = State) w
...
+%% from the peak - doubling overshoots by up to half of where it lands, and
+%% against the ceiling by more - and a step of one would take an epoch each to
+%% walk back, so the reader would spend most of its run getting there.
+step_target(Target, up) -> Target + max(1, Target div 8);
+step_target(Target, down) -> Target - max(1, Target div 8).
+
+flip(up) -> down;
+flip(down) -> up.
+
+improved(Rate, Prev) ->
```
lukebakken, line 1613: `Rate * 100 > Prev * 105` is true for any nonzero `Rate` when `Prev` is 0, so a single zero-byte epoch makes the next one double the target.

The `prev_rate = undefined` clause above stores the reading rather than discarding it, so a zero epoch leaves `prev_rate = 0` rather than `undefined`. Verified at shipped defaults (`inflight_initial = 32`, `max_depth = 64`): target at init 32, after a 0-byte tick 32, after the next tick with any bytes at all 64, where the `Target >= max_depth` clause pins it. So `prefetch_max_depth` becomes the operating point within about 400 ms, which is the opposite of what `config.erl:180` promises ("a ceiling, not the operating point. What a reader runs at is `prefetch_inflight_initial`").

A zero-byte epoch is not exotic: credit-based stream delivery means a consumer that stops asking produces one, and the tick's own comment notes "A reader that has nothing in flight ticks over zero bytes". Guarding `improved/2` on `Prev > 0`, or treating a zero epoch as no signal, would cover it.

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -891,11 +1035,15 @@ maybe_start_requests(State0) ->
...
+    %% `none` for the look-ahead's tail is what defers it: it reads as "the
+    %% manifest ends here", so the frontier fills the current fragment and stops
+    %% without walking the iterator. It is passed rather than stored, so the
+    %% state's own look-ahead is untouched.
+    {State1, _Peeks, _Tail, _Attempted, Effects} = extend_frontier(State, [], none, []),
     {State1, Effects}.
 
 %% (Re-)issue ranges that are queued and not in flight. Their bytes are already
-%% counted in `outstanding/1`, so they are not window-gated - a range the reader
+%% counted in `committed/1`, so they are not budget-gated - a range the reader
```
lukebakken: On `issue_ready/1` just below: it gates re-issue on `max_depth` rather than `inflight_target`, so the contention back-off is undone by its own retry.

Verified with 64 ranges in flight, `slow_down` on all of them, then a tick: `tune/2`'s contention clause cuts the target 64 -> 48, the `fault` timer fires, `release(fault, _)` marks all 64 `ready`, and `ready(MaxDepth, _)` re-issues all 64. `inflight_target` is enforced only in `has_room/1`, i.e. only on first issue, so it never constrains the requests that immediately follow a throttling response.

The comment's reasoning for not window-gating committed ranges is right, and I am not suggesting they be dropped. It is specifically the depth cap being the fallback bound: `min(MaxDepth, inflight_target)` would keep the committed-ranges guarantee while letting the back-off mean something. Same shape applies to `pool_busy`.

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -295,11 +383,30 @@ step_(State0, {data, Id, Data, DoneOrContinue}) ->
     {Signals, Pipeline} = rabbitmq_stream_s3_read_pipeline:data(
         Id, Data, DoneOrContinue, pipeline(State0)
     ),
-    {State1, Effects1} = absorb_signals(Signals, State0#state{pipeline = Pipeline}),
+    Epoch = State0#state.epoch_bytes + iolist_size(Data),
```
lukebakken: `epoch_bytes` is credited before `absorb_signals/2` and without consulting what `read_pipeline:data/4` did with the frame, so it counts bytes that never reached a buffer: `take_req/2` returning `false` drops the delivery outright (a frame for a range a deadline or a refresh dropped), and `stage/2` clips over-delivery to `RangeEnd + 1 - Pos`.

The effect is directional rather than noisy - the rate reads high exactly after the event that wasted the bandwidth, and the search reads that as the extra concurrency paying off. Counting what `data/4` accepted rather than `iolist_size(Data)` would fix it.

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -473,6 +581,10 @@ step_(State0, {iterator_refreshed, Iterator}) ->
                 stream = StreamId,
                 cfg = Cfg,
                 window = Cfg#cfg.request_size,
+                %% The iterator refresh rebuilds the state, so the search starts
+                %% over. What it had learned was about a fragment the reader is
+                %% no longer on, and it converges again within a few epochs.
+                inflight_target = min(Cfg#cfg.max_depth, Cfg#cfg.inflight_initial),
```
lukebakken: This clause rebuilds `#state{}` from scratch, so `epoch_bytes` resets to 0 mid-epoch. The shell writes `epoch_at` only in `init/1` and in `handle_info(tune_tick, ...)`, and the core has no effect to ask it to restamp, so the next tick divides a partial epoch's bytes by a full epoch's elapsed time and reports an artificially low rate.

That lands on `improved(Rate, Prev)` with a depressed `Rate`, and combined with the zero-`Prev` case flagged on `improved/2` it is a plausible route into the epoch-after doubling the target. With N consumers refreshing together - retention-driven 404s, or a rebalance - that is N x 64 checkouts against a 200-connection pool.

Resetting the target here is clearly deliberate and the comment justifies it. It is the byte/time mismatch I would fix, by restamping the epoch alongside the reset.

test/prop_SUITE.erl
```diff
@@ -2408,19 +2407,28 @@ prop_remote_reader_core_load_bounded() ->
...
+            %% what may be on the wire comes from the concurrency target, which
+            %% cannot exceed `max_depth`; what may be held unread comes from the
+            %% AIMD window, which cannot exceed `window_max`. Each may cross its
+            %% ceiling by the one range `has_room/1` admits before checking
+            %% again, and each is floored at what the read in hand needs.
             MaxReadEnd = lists:max([0 | [O + B || {read, O, B, _} <- Events]]),
-            MaxOutstanding = max(WindowMax, MaxReadEnd) + RequestSize,
-            check_rrc_load(Events, S0, MaxOutstanding, MaxDepth)
+            CommittedBound = max(MaxDepth * RequestSize, MaxReadEnd) + RequestSize,
+            BufferedBound = max(WindowMax, MaxReadEnd) + RequestSize,
```
lukebakken: `BufferedBound` asserts a bound the implementation does not provide, and the comment above it states the reason it does not: "what may be held unread comes from the AIMD window, which cannot exceed `window_max`". `has_room/1` says the opposite - committed bytes all become buffered bytes, and its own comment records that the two budgets add rather than share. `CommittedBound` on the line above correctly carries `MaxDepth * RequestSize`; `BufferedBound` is missing the symmetric term.

With the property's own parameters (`request_size=4000`, `max_depth=8`, `window_max=32000`) and a 3-event prefix, `buffered` reaches 63000 against a bound of 36000. It passes today at 500 iterations, so this is latent rather than red - which is the part I would want to fix, because whether it stays passing is luck.

Separately, and the bigger gap for this PR: `gen_rrc_event/1` has no `tune_tick` clause and `gen_rrc_error_event/0` no `not_found`, so `tune/2`, `retarget/2`, `note_contention/2` and the new `checked/1` guards are never fuzzed, and `inflight_target` never moves off its initial value in any property. The down half of `tuner_stays_within_its_bounds` (`remote_reader_core_SUITE.erl:720`) is also vacuous: it asserts `Target >= 1 andalso Target =< 12`, which `retarget/2` enforces unconditionally on every write, so inverting `step_target/2` or deleting the fallback arm still passes.

lukebakken (commented):
> [!NOTE]
> This review was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken. The behavioural claims were verified by driving `rabbitmq_stream_s3_remote_reader_core` directly; @lukebakken set the bar for which findings were worth sending, and several candidates were cut against it. The prose and the technical framing are Claude's.

Reviewed the tuner and the look-ahead. Six things I am confident about, five I am raising as questions, and two notes on things I checked and found fine.

The theme of the confident set is that the controller's input is not trustworthy yet: `improved/2` treats a zero-byte epoch as an improvement, the epoch's byte count and its elapsed time can come from different intervals, and the contention back-off is undone by its own retry. Separately, nothing in the test or bench story exercises the search: the property never ticks it, and no bench run sets `S3B_AUTO_TUNE`, so every measurement here ran with `auto_tune` off while the plugin ships it on.

Inline on each. Three that cannot be anchored to a diff line:

- `prefetch_auto_tune` has no cuttlefish mapping. `priv/schema/rabbitmq_stream_s3.schema` has 30 mappings and no `prefetch_*` entry, so the switch its own comment calls "the way back if the search is ever the wrong call for a workload" is reachable only from `advanced.config`, not `rabbitmq.conf`. Same for `prefetch_inflight_initial` and `prefetch_max_lookahead`
- `docs/read-path.md` is untouched, and `AGENTS.md` names it as the document for this path. Statements that are now false: :92 "The window is also the reader's memory bound: it holds or has outstanding at most `prefetch_window_max` plus one request" (the two budgets add now, which `has_room/1` explains well - the doc just does not say so), :83 "fewer than `prefetch_max_depth` (8) requests are in flight", :87 "capped at `prefetch_window_max` (32 MiB)", and :108 "The reader looks one fragment ahead and no further". `docs/operations.md:540` ("up to `prefetch_max_depth` (8)") is the paragraph an operator sizes a request budget from, and the bucket list at :506 is now an 8 MiB stride to 128 MiB
- `remote_reader_inflight_target` is not in the metrics table at `docs/operations.md:490-498`, though its docstring says it should be read against `requests_in_flight`

Two I went looking for and am satisfied about, in case they come up: the memory bound is not unbounded - `has_room/1`'s comment states that the two budgets add and that worst case is their sum, which is a deliberate call and documented where it belongs. And `grow_count/1`'s extraction is term-for-term faithful, with the new `max(0, ...)` fixing a real pre-existing hazard where a negative `N` recursed downward opening connections.

Method note, since a few of these are behavioural claims: I drove `rabbitmq_stream_s3_remote_reader_core` directly with throwaway probe modules to confirm the tuner ones, and reproduced the `peeks`/`nexts` case end to end before deciding it was not reachable in production. `erlfmt -c` is clean, all changed modules compile warning-free with and without `-DTEST`, and `remote_reader_core_SUITE` (102) and `prop_SUITE` (51) both pass.

test/remote_reader_s3_bench.erl
```diff
@@ -0,0 +1,779 @@
...
+                        %% fragment, so sweeping this measures multi-fragment
+                        %% look-ahead against single-fragment reach rather than
+                        %% against a differently-configured reader.
+                        max_lookahead => env_int("S3B_LOOKAHEAD", Depth),
+                        %% Off, `S3B_DEPTH` is the concurrency the reader runs
+                        %% at, which is how every fixed-concurrency sweep here
+                        %% was measured. On, it is only the ceiling: the reader
+                        %% searches below it, and the sweeps are the ground
+                        %% truth for what it ought to find.
+                        auto_tune => env_int("S3B_AUTO_TUNE", 0) =:= 1,
```
lukebakken, line 193: `auto_tune` defaults to off here and nothing sets `S3B_AUTO_TUNE` - not `scripts/`, not the `Makefile`, not the docs. So `gmake bench-remote_reader_s3_bench`, every `s3-bench-sweep.sh` point, `s3-bench-validate.sh` and `s3-bench-defaults.sh` all run at a fixed `inflight_initial = S3B_DEPTH` with the search disabled, while the plugin ships `prefetch_auto_tune = true`.

That matters most for `s3-bench-defaults.sh`, whose stated question is what the plugin should ship with: none of the ramp, the anchor, or the contention back-off is exercised by the runs that justify the new defaults. It also makes the `target_final` and `target_max` columns constants equal to `S3B_DEPTH` in every reported row, though their comment says they "separate a search that settled on the wrong target from one that settled on the right target too slowly".

While in here, a few smaller things in this file and its scripts:

- `?S3_HOST`/`?S3_PORT` default to `10.89.0.2:9000`, but `scripts/s3-bench-env.sh:29` publishes `19000:9000`, and the `bench-%` target sets neither - so the sequence in this module's own moduledoc and in `docs/DEVELOPMENT.md:118-121` depends on the container landing on that exact bridge address
- `S3_BENCH_ENGINE` is a plain Make variable and never exported, while `s3-bench-sweep.sh:25` reads it from the environment with a `podman` default, so `gmake S3_BENCH_ENGINE=docker s3-bench-up && gmake S3_BENCH_ENGINE=docker s3-bench-validate` brings MinIO up under docker and then inspects with podman
- `set -euo pipefail` at `s3-bench-sweep.sh:16` means the `S3B_HOST="$($ENGINE inspect ...)"` assignment at :32 exits before the "try make s3-bench-up" guard on the next line can print
- `S3B_RATE_KBPS` and `S3B_JITTER_MS` are documented at :133 and exported by the sweep, but are not among the 14 env vars this module reads, so `./scripts/s3-bench-sweep.sh rate 100 200 400` runs three byte-identical configurations
- `docs/DEVELOPMENT.md:141` says "The harness checks this and warns with the exact predicted ceiling" for `net.core.rmem_max`; nothing reads `rmem` and nothing emits `WARNING`, so the `grep -E '^S3BENCH|^WARNING'` at `s3-bench-sweep.sh:114` filters for output that is never produced
- `-spec seed(map()) -> [#fragment_ref{}]` contradicts both clauses, which return `{Stream, Refs}`; `netem/1`, `netem_cmd/1`, `?BUDGET_BYTES` and `?FRAGMENT_BYTES` look dead

test/remote_reader_s3_bench.erl
```diff
@@ -0,0 +1,779 @@
...
+    Target = Bytes * 1_000_000 div (DrainMiBs * 1_048_576),
+    case Target - Elapsed of
+        Sleep when Sleep > 1000 -> timer:sleep(Sleep div 1000);
+        _ -> ok
+    end.
+
+drain_from(_Reader, _Pos, Budget, Acc) when Acc >= Budget ->
+    Acc;
+drain_from(Reader, Pos, Budget, Acc) ->
+    case rabbitmq_stream_s3_remote_reader:read(Reader, Pos, ?READ_BYTES, within_chunk) of
```
lukebakken, line 611: `drain_from/4` reads through `read/4`, which this module's own comment at :96 rules out: "Read through `read_iodata/4` rather than `read/4`, because the latter flattens the buffer's blocks into one binary - a copy of every byte delivered, which the real log reader does not make".

So every `mib_s` figure is charged a per-MiB copy the production path does not make, including the ones `s3-bench-validate.sh` compares against real broker measurements.

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -1184,6 +1422,196 @@ decay(#state{cfg = #cfg{request_size = RequestSize}, window = Window} = State) w
...
+    });
+tune(Rate, #state{prev_rate = undefined} = State) ->
+    %% Nothing to compare against yet - the first epoch, or the one after a
+    %% contention backoff. Take the reading and move on the next tick.
+    State#state{prev_rate = Rate};
+tune(
+    Rate,
+    #state{tune_phase = ramp, inflight_target = Target, prev_rate = Prev, cfg = Cfg} = State
+) ->
+    case improved(Rate, Prev) of
```
lukebakken, line 1509: Question rather than a finding: is there a reason not to gate the search on whether the reader was actually supply-limited during the epoch?

The tick is unconditional and its own comment notes a reader with nothing in flight ticks over zero bytes, but nothing downstream treats zero as "no signal". A consumer that pauses one epoch mid-ramp takes `improved(0, Prev)` -> false -> `retarget(Target div 2)` plus `prev_rate = undefined`, costing two more epochs before the target moves again. In `climb` with an anchor it is worse: `0 * 100 < Best * 90` fires the fallback branch and clears `best_rate`/`best_target`, the next zero epoch records `best_rate = 0` through the `Best =:= undefined` branch, and from there `classify(0, 0)` is `flat`, so the search walks the target up by `Target div 8` for `?TUNE_PROBE_LIMIT` steps, snaps back to a zero-rate "best", flips direction and repeats.

For credit-based stream delivery a flow-controlled consumer is the normal state, not the exception, so as written its concurrency is decided by whichever target happened to be set when it last had demand. `inflight_owing` having been at target for the epoch looks like the natural gate.

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -1018,76 +1195,127 @@ fetch_ceiling(#state{window = Window}) ->
...
-        {error, {group_fetch_failed, _}} -> {failed, true}
+%% Whether to re-attempt is read off that clock rather than stored beside the
+%% `failed` marker, so the two cannot drift apart: dropping the clock is what
+%% licenses the next attempt, which means every site that disowns the clock
+%% licenses one without having to remember to.
+%%
+%% `Attempted` says whether a fetch was actually made, which is what decides the
+%% retry clock: a pass that never reached the look-ahead owes nothing, and arming
+%% for it would keep alive the very clock that suppresses the next attempt.
+extend_peeks(State, Peeks, Tail) ->
```
lukebakken, line 1308: The "at most one synchronous group GET per pass" bound is gone, and I wanted to check that was intentional. The old `peek_next_fragment/2` resolved a single memo at most once per `step/2`, with the comment "Peeking per queued range would spend `max_depth` of them on the same object ... Either blocks the reader while the caller's read deadline burns."

`spill/5` now calls `extend_peeks/3` once per `extend_frontier/5` iteration, and `extend_frontier/5` recurses once per range it pushes, so a single `step/2` can perform up to `max_lookahead` blocking group GETs with the consumer's read deadline running. `max_lookahead` defaults to `prefetch_max_depth`, so 64. It is reached whenever one range covers one fragment, which the harness's own `?FRAGMENT_BYTES` comment describes (8 MiB fragments against 4 MiB requests).

Related, at `:1288`: a `pool_busy` from that descent surfaces as `{error, {group_fetch_failed, _}}` -> `peek_tail = failed` -> `arm_retry(fault, _)`, so pool exhaustion lands on the 1 s-doubling-to-30 s fault clock rather than the 25-500 ms pool clock. The reasoning for using the fault clock to pace descents is sound, but `pool_busy` resolving in milliseconds is a different condition from an unfetchable group object.

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -1184,6 +1422,196 @@ decay(#state{cfg = #cfg{request_size = RequestSize}, window = Window} = State) w
...
+%% while the connections open: it is the cost of growing, not evidence of having
+%% grown too far. Counting it would make every attempt to grow undo itself. It
+%% resolves on its own, and if the extra concurrency genuinely does not pay, the
+%% epoch rate says so.
+note_contention(slow_down, #state{epoch_contention = Count} = State) ->
+    State#state{epoch_contention = Count + 1};
+note_contention(_Reason, #state{} = State) ->
+    State.
+
+tune(_Rate, #state{cfg = #cfg{auto_tune = false}} = State) ->
```
lukebakken, line 1475: Small one: `auto_tune = false` is honoured here, but the shell arms `tune_tick` unconditionally in `init/1` and re-arms it on every tick, so a reader with the search off still wakes at 5 Hz for its whole life to compute a rate nothing consumes.

Also `note_contention/2` counts only `slow_down` and drops `pool_busy`. The stated reason - the pool grows reactively, so a checkout that finds nothing free is what raising concurrency looks like while connections open - holds below `general_pool_max_size`, but at the cap `grow_count/1` returns 0 and every checkout waits the 100 ms timeout and returns `pool_busy` indefinitely. The pool can tell those apart (`open >= max_size`) and the reader cannot. With depth 64 per reader, three or four readers reach the cap, and each then measures an epoch rate set by pool waiting rather than by its own concurrency, which `classify/2` reads as `flat` and answers by stepping the target further in the current direction.

src/rabbitmq_stream_s3_remote_reader_core.erl
```diff
@@ -247,6 +334,7 @@ init(StreamId, FragRef, Position, Iterator, Opts) ->
         stream = StreamId,
         cfg = Cfg,
         window = Cfg#cfg.request_size,
+        inflight_target = min(Cfg#cfg.max_depth, Cfg#cfg.inflight_initial),
```
lukebakken: Question about attach latency: `inflight_target` starts at 32, and `start_current_request/1` then loops until the fragment is covered or the target is reached, so `init/5` issues up to 32 ranges rather than one. Because `remote_reader:init/1` runs `execute_effects/2` before returning `{ok, State}`, those `get_range_async/3` calls - each a pool checkout that can cost the 100 ms checkout timeout, plus a signing - all happen inside `gen_server:init/1` while the consumer is blocked in `gen_server:start/3`.

On main, `has_room/1` was gated by `window = request_size`, so init issued exactly one range and the fetch side grew only by actually starving. The `note_miss/1` comment that stated the guarantee ("Growing once per missed read also keeps a consumer that reads a couple of records and disconnects from provoking a full window: it only ever misses once or twice") went with the change, and I did not find a replacement. So a consumer that attaches at an old offset, reads one record and disconnects now bills up to 32 signed range GETs and 32 checkouts instead of one.

If the higher start is the point, deferring the ramp until after `init/1` returns would keep it off the attach path.

src/rabbitmq_stream_s3_read_pipeline.erl
```diff
@@ -352,10 +440,41 @@ push(
...
+    P.
+
+find_next(_Fragment, []) ->
+    error;
+find_next(Fragment, [{#fragment_ref{offset = Fragment}, Buffer} | _]) ->
+    {ok, Buffer};
+find_next(Fragment, [_ | Rest]) ->
+    find_next(Fragment, Rest).
+
+replace_next(Fragment, Buffer, [{#fragment_ref{offset = Fragment} = FragRef, _} | Rest]) ->
```
lukebakken, line 477: Not a live bug as far as I can tell, but an invariant worth writing down or asserting.

`replace_next/3` has no `[]` clause, unlike `find_next/3` directly above it, so an unseated fragment reaching it from `set_buffer_for/3` is a `function_clause` that takes the reader down and surfaces to the consumer as `{error, {remote_reader_down, _}}` rather than a diagnosable `error`.

The same assumption exists in the core: `not_found_refresh/1` reads the 404'd offset off `peek_head/2`, which is only the right fragment while the heads of `peeks` and `nexts` name the same one. `spill/5` walks past a fragment without seating it when `range_in_fragment/3` returns `none`, and that is the one way they can diverge. I built a manifest that does it - `[{0,500},{100,0},{200,500},{300,500}]`, where fragment 100 has a zero-length data region - and confirmed the consequence: after the transition the reader is on 200 with 100 still at the head of `peeks`, and a 404 for 300 emits `{refresh_iterator, 200}`, the offset of the fragment being read rather than the one that 404'd.

I then went looking for whether that manifest is producible and concluded it is not: `range_in_fragment/3` returns `none` only when `Frontier >= ?SEGMENT_HEADER_B + FragSize`, which for an unseated fragment needs `FragSize` to be exactly 0, and `fragment_iterator:next/1` only ever yields `?MANIFEST_KIND_FRAGMENT` entries as `#fragment_ref{}` while group entries (the documented size-0 case) are descended rather than returned. So I am not reporting a defect. But the pairing is now maintained across up to 64 fragments by `register_next/2`, `drop_fragment/2`, `drop_prefetch/1` and `advance/1`, it is only asserted in `checked/1` under `-ifdef(TEST)`, and the failure mode in production is a bare crash. A `[]` clause returning `error`, or a note on `spill/5` recording why the heads cannot diverge, would keep the next change to that code honest.

the-mikedavis:
I squashed a bunch of the review :point_up: but I didn't get a chance to swing through and resolve the threads that are done.

I tested this in 3-node clusters in EC2 w/ m7g.16xlarge and this is a measurable improvement, but not the 5 Gbps we're seeking. Before this change the remote reader would reach about 249 MiB/s in my tests, now it reaches 389 MiB/s (56% improvement). It's disappointing, though, because that's much below the benchmark script's measurements and much lower than the target, and similarly, local reads (590 MiB/s). It's a step forward but does fall short :sweat:

the-mikedavis:
With the most recent changes we're reaching 5 Gbps now. I will continue testing to see if the defaults need any further tuning to reach this reliably, and rewrite the PR description to get it ready for merging.

the-mikedavis:
Re-validating this as I am cleaning up the branch, I saw single-consumer throughput drop to 550 MiB/s. I thought that it might be a regression from an unintentional change in the cleanup but actually the two different test brokers ended up with streams with different chunk sizes. Mean chunk size of the broker where I saw 590 MiB/s was 761.4 KB, mean size for 550 MiB/s was 205.9 KB. I think the publishing was roughly the same so I'm not sure where the difference came from. Large-chunk broker saw `rate(rabbitmq_stream_s3_read[1m])` hover around 1.6 reads/s while small-chunk saw upwards of 5.5-5.8k reads/s.

Another angle I want to look at before opening this up is request size. It's currently 4 MiB range-GET size. But for high throughput around 5 Gbps that's a pretty fair number of requests per second, and I'd be a bit worried about hitting S3 requests-per-second limits. Ideally I want the reader to work as well with 4 MiB as it does with 8 MiB or 16 MiB.

the-mikedavis:
The most recent changes make consumption throughput identical across 4 MiB / 8 MiB / 16 MiB requests. I will add an issue for auto-tuning request size. It's good to tune since larger requests are slightly faster. But it's a significant slice of work, worth its own PR.

Edit: https://github.com/amazon-mq/rabbitmq-stream-s3/issues/372

test/prop_SUITE.erl
```diff
@@ -1875,7 +1895,7 @@ prop_remote_reader_core_reply_bytes_exact() ->
                 end,
             Opts = #{
                 request_size => RequestSize,
-                window_max => RequestSize * 8,
+                max_memory => RequestSize * 16,
```
lukebakken, line 1899: This fixture now fuzzes a strictly serial reader, so the interleaving it exists to cover is gone.

`Opts` carries no `auto_tune` key, so `build_cfg/1` defaults it to `true` (core:1382) and `initial_target(#cfg{})` is 1 (core:1472), and this property's step generator never emits a `{tune_tick, _}`, so nothing moves the target for the whole run. I instrumented the property itself rather than a replica, tracking the maximum `length(outstanding_rrc_ranges/1)` across all 500 iterations: it is **1**. On `main` this fixture set `window_max => RequestSize * 8` and the miss-driven window got to 8, which its own comment at `:2113` recorded ("`window_max` is 8 requests in this fixture").

With one range outstanding there is never a predecessor, so `#req.staged` is never non-empty, and `answer_rrc_ranges/3`'s rotation and its `continue`/`done` alternation are both no-ops. Two comments now claim otherwise: `:1871-1873` ("with several ranges of a fragment in flight, responses interleave... This is the property that covers reassembly") and `:2144` ("Fetching gets 8 requests in this fixture; the slack is for the miss that widens the window").

Adding `auto_tune => false, inflight_initial => 8` to `Opts` restores it, and reads as the deliberate choice for a fixture whose subject is reassembly rather than the search.

For what it is worth I checked the sibling property before writing this: `prop_remote_reader_core_survives_failure_interleavings` reaches **5** concurrent ranges by the same measurement, because failures re-request tails and short completions add gap requests. So this is the one fixture that went serial, not both.

scripts/s3-bench-validate.sh
```diff
@@ -0,0 +1,98 @@
...
+# `grouped` costs one descent, not one per fragment transition. A rebalance
+# factors 1024 leaves into a group and the iterator keeps the descended-into
+# entries on its own stack, so the run behind a group node is walked in memory.
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+# name | request | memory | depth | fragment | manifest | drain | measured
+SCENARIOS=(
+  "shipped-defaults  4  64   8   64 grouped 0   249.1"
```
lukebakken: This row is named `shipped-defaults` but the tuple is request 4 MiB, memory 64 MiB, depth 8, which is what the plugin shipped *before* this branch. `s3-bench-defaults.sh:55` gives that identical tuple its correct name, `pre-4.4-defaults`, and its `shipped-defaults` row on the line above is `4 256 64`.

Since this is the script `docs/DEVELOPMENT.md:137` presents as the gate on trusting the harness, the label matters more here than anywhere else: as it stands the validation step certifies the old configuration under the new one's name.

scripts/s3-bench-lib.sh
```diff
@@ -0,0 +1,12 @@
...
+# summarise what it prints. Sourced, not run.
+
+# Median of one field across a run's S3BENCH lines, read from stdin. `sed -n
+# ...p` prints only what it matched, so a line that does not carry the field
+# contributes nothing rather than passing itself through - a run that died after
+# printing a partial line would otherwise reach the callers' python3 call as
+# source text and, under `set -e`, take the whole sweep down with a SyntaxError.
+# An empty result is what the callers check for.
+median_field() {
+  sed -n "s/.*$1=\([0-9.]*\).*/\1/p" | sort -n | awk '{a[NR]=$1} END{print a[int((NR+1)/2)]}'
```
lukebakken: `median_field` reports the lower of two samples rather than a median. `awk '{a[NR]=$1} END{print a[int((NR+1)/2)]}'` with NR=2 evaluates `int(1.5)` to 1 and prints the first element of the sorted input, so the smaller value wins (checked: `249.1, 71.0` prints `71.0`).

Both summarising scripts default `REPS` to 2, and the failure mode they document (`s3-bench-sweep.sh:105-110`: "It happens on roughly half of runs ... costs ~3.5x throughput") is exactly the case where one rep of two degrades. So whenever that fires the printed figure and its `err%` are the degraded run, in the two scripts whose output is the evidence for the new defaults.

For an even count either `(NR+1)/2` rounded up or the mean of the two middles would do; simplest is probably `a[int(NR/2)+1]`.

scripts/s3-bench-validate.sh
```diff
@@ -0,0 +1,98 @@
...
+BUDGET="${S3B_BUDGET_MIB:-2048}"
+
+# shellcheck source=scripts/s3-bench-lib.sh
+. ./scripts/s3-bench-lib.sh
+
+# The search stays off here, pinned rather than left to the harness's default.
+# The measured values above come from brokers that ran at a fixed depth, so a
+# run that searched for its own would not be comparable with them - which is the
+# only thing this script is for.
+export S3B_AUTO_TUNE=0
```
lukebakken, line 66: With `S3B_AUTO_TUNE=0` here, neither entry point `docs/DEVELOPMENT.md:122-123` tells a reader to run measures the search. `Makefile:47-48` (`bench-%`) sets no `S3B_*` at all, and the module default is `env_int("S3B_AUTO_TUNE", 0) =:= 1` with depth 8 and 64 MiB, so both documented commands run the pinned pre-4.4 path while the plugin ships `prefetch_auto_tune = true`. `s3-bench-defaults.sh` is the only entry point that measures what ships, and it is the one not mentioned in the docs.

Pinning the search off for a reproducibility gate is a reasonable choice on its own; what seems off is that nothing a reader is pointed at exercises the shipped configuration. Worth a line in `DEVELOPMENT.md` naming `s3-bench-defaults.sh` as the one that does.

test/remote_reader_s3_bench.erl
```diff
@@ -0,0 +1,844 @@
...
```
lukebakken: Still reading through `read/4`, which this module's own comment at `:102-106` rules out: "Read through `read_iodata/4` rather than `read/4`, because the latter flattens the buffer's blocks into one binary - a copy of every byte delivered, which the real log reader does not make on the data path". `read/4` is `flatten(read_iodata(...))` with `flatten({ok, IoData}) -> {ok, iolist_to_binary(IoData)}` (remote_reader.erl:285-293).

So every `mib_s` the harness prints is charged a per-MiB copy production does not make, including the figures `s3-bench-validate.sh` compares against real broker measurements. `iolist_size/1` on the next line works on either shape, so it is a one-word change.

docs/read-path.md
```diff
@@ -80,22 +80,23 @@ A range in that state that has *not* failed stays queued until its closing frame
...
 
-The window is also the reader's memory bound: it holds or has outstanding at most `prefetch_window_max` plus one request. A consumer that stops reading needs no separate throttle - the buffer fills to the window and no further requests are issued.
+Fetching does not consult the buffer at all, so a full buffer can never stop a reader issuing: every delivered byte moves from committed to buffered and stays there until the consumer reads it, so a bound the two shared would let the buffer lock the fetch side out entirely.
```
lukebakken: "Fetching does not consult the buffer at all, so a full buffer can never stop a reader issuing" is not what the code does: `room_for_bytes/1` (core:1180) checks `buffered(State) + Committed >= memory_ceiling(State) -> {stall, buffer}`, and this branch adds a counter for exactly that outcome, `remote_reader_prefetch_stall_buffer`, "Placement passes stopped by the memory ceiling: the consumer is far behind".

The true statement is the narrower one `fetch_ceiling/1`'s own comment makes, that *`fetch_ceiling`* does not consult the buffer. The paragraph below at `:93` is already right, so it is really this sentence that needs the qualifier. Worth fixing because `AGENTS.md:18` names this file as the one to read before touching the read path.

docs/operations.md
```diff
@@ -496,20 +536,36 @@ Owned by `rabbitmq_stream_s3_remote_reader`. One counter set per node, summed ac
...
+| `rabbitmq_stream_s3_remote_reader_prefetch_stall_depth`         | `prefetch_max_depth` itself                              |
+| `rabbitmq_stream_s3_remote_reader_prefetch_stall_fetch_budget`  | Half `prefetch_max_memory`, the cap on what may be committed |
+| `rabbitmq_stream_s3_remote_reader_prefetch_stall_buffer`        | The memory ceiling: the consumer is far behind           |
+| `rabbitmq_stream_s3_remote_reader_prefetch_stall_reach`         | Nothing left to ask for - the manifest or look-ahead ran out |
+| `rabbitmq_stream_s3_remote_reader_prefetch_stall_peek_failed`   | A look-ahead group fetch is failing                      |
+
+An idle reader emits none of these: a reader with no read in hand and nothing
+outstanding is finished rather than stalled.
+
+Read `committed_bytes` against `fetch_ceiling_bytes` and `buffered_bytes`
```
lukebakken: "The two are bounded independently, so a reader can be at one and far from the other; summed they cannot say which" points an operator the wrong way, and `:604` of this same file states it correctly ("sum to what a node's readers hold now, against `rabbitmq_stream_s3_remote_reader_memory_ceiling_bytes`").

They are bounded jointly in `room_for_bytes/1`, and `memory_ceiling/1` is the whole `max_memory` rather than half, so following this paragraph an operator reads `buffered_bytes` at 200 MiB against a 256 MiB ceiling, concludes there is headroom, and misses that the sum is what is pinned and that `prefetch_stall_buffer` is climbing.

src/rabbitmq_stream_s3_config.erl
```diff
@@ -163,17 +167,72 @@ general_pool_max_size() ->
...
+-spec prefetch_max_memory() -> pos_integer().
+prefetch_max_memory() ->
+    application:get_env(?APP, prefetch_max_memory, ?MEMORY_REQUESTS * prefetch_request_size()).
+
+%% The most range GETs one reader may ever have in flight - a ceiling, not the
+%% operating point. What a reader runs at is searched for from measured
+%% throughput, or `prefetch_inflight_initial` when `prefetch_auto_tune` is off.
+%%
+%% Set generously for that reason. It bounds a reader's share of the general
+%% pool, and a pool that cannot serve a checkout surfaces as `pool_busy` or
```
lukebakken, line 190: Question rather than a finding, about this sentence: "a pool that cannot serve a checkout surfaces as `pool_busy` or `pool_exhausted`, so a ceiling above what the pool can serve costs a backoff, not a stall".

That holds for range GETs. It looks like it does not hold for the look-ahead's group GETs, which go through the same general pool: `extend_peeks_/3` calls `fragment_iterator:next/1` into `manifest:fetch_group/4`, which on any non-404 error including `pool_busy` and `pool_exhausted` sleeps 100 ms and retries twice (manifest.erl:307-321), so one descent can hold `step/2` for up to about 500 ms with the consumer's read deadline running, and then returns `{error, {group_fetch_failed, _}}`. `arm_peek_retry(failed, true, _)` puts that on the fault clock (1 s doubling to 30 s) rather than the 25-500 ms pool clock, and `extend_peeks_/3` will not re-attempt while that clock is armed. With `peeks` empty at a fragment boundary that reads like a stall of up to 30 s from a condition that clears in milliseconds.

I have not reproduced pool exhaustion end to end, so treat this as a question about the reasoning rather than a measurement. Using the pool clock when the descent failed for a pool reason would seem to close it. Two mitigations I did confirm, which is why this is a question and not a finding: `rebalance_threshold` is 1024, so walking even 64 fragments crosses at most one group boundary, and each descent is memoised in `peeks`. That also retracts the harsher version of this I raised last round, where I claimed up to 64 blocking group GETs in one pass; that does not hold.

lukebakken (commented):
> [!NOTE]
> This review was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken. Every claim was verified against the branch, and the concurrency and memory figures were measured by driving `rabbitmq_stream_s3_remote_reader_core` directly. @lukebakken set the standard of evidence; several candidate findings were narrowed, dropped or withdrawn against it. The prose is Claude's.

Second pass, on the reworked branch. Much smaller than last time: the tuner findings from the previous round are all fixed, and I confirmed each one rather than taking the diff's word for it. `tune(0, State)` now precedes every clause that stores a rate so `prev_rate` cannot hold 0; `ready/3` applies both gates; the sample survives an iterator refresh with a comment saying why; `init/5` issues exactly 1 range at the shipped defaults (measured); `replace_next/3` raises `{unseated_fragment, _}` and the peeks/nexts pairing is asserted structurally, which I re-checked for reachability under multi-fragment prefetch and still could not reach.

What is left is one real coverage regression, a handful of statements that no longer match the code, and two bench-harness things. Inline on each.

One that cannot be anchored, because the file is not in the diff: `docs/investigations/remote-reader-memory.md:3` is a status-update banner whose job is to say what superseded the write-up, and it still names `prefetch_window_max` at 32 MiB as the current per-consumer ceiling. That setting is gone and the ceiling is now `prefetch_max_memory` at 256 MiB, so the one place someone sizing reader memory would look understates it by 8x. `AGENTS.md:22` points readers at `docs/investigations/` before debugging a hard problem.

Also, and I suspect you already have this since you mentioned cleaning the branch up: the deleted window survives in about 10 comments under `src/` and in four test names, and four of those comments describe the mechanism rather than just using the word, so they now say something untrue. `remote_reader_core_SUITE.erl:1875` and `:2639` ("Two misses grow the window" / "A miss widens the window to two requests") and `remote_reader.erl:946` are the ones I would fix; the rest are cosmetic.

On your S3 request-rate question, arithmetic rather than a guess: 4 MiB range GETs at 625 MB/s is about 149 GET/s per reader, against roughly 5,500 GET/s per prefix. All readers of one stream share the prefix, so it takes on the order of 37 concurrent full-rate readers of the same stream to reach it. I looked for places the design spends requests it does not need and did not find one: `fail/3` restarts a range at `flushed` so a failure refetches only the un-flushed tail, a short completion queues exactly one gap request, and `frontier/2` takes the max of the buffer end and the queue end so the frontier never walks backwards. The two amplifiers are both on the deadline path and both already documented.

One thing I could not substantiate, recorded so nobody re-raises it from first principles later: I went looking for whether a slow or idle consumer now holds more than a fast one, on the theory that `memory_ceiling/1` covers buffered plus committed while `fetch_ceiling/1` only bounds committed, and that nothing reaps an idle reader (`maybe_stop/1` needs `stopping`, set only on `become_local`). Driving the core over 40 ticks of virtual time at the shipped defaults against a 1 GiB manifest, I measured peak held of 67 MiB for a consumer that stops, 67 MiB at 5 MB/s, 67 MiB at 50 MB/s and 132 MiB reading flat out, so it comes out the ordinary way round. My harness answers every outstanding range in full on each tick, which keeps `committed` at 0 and is not how S3 behaves, so I would not treat those numbers as the last word either. Flagging it only as a thing that looked plausible from the ceilings and did not show up in practice.

One test note: `ct-log_reader` gives 17 of 18 for me, with `local_to_remote_transition` failing deterministically at the `?assertEqual(remote, mode(Reader2))`. I believe it fails the same way on `main` and is not from this branch, but I did not verify that myself, so it is worth a glance. Everything else is green: eunit 268, `remote_reader_core` 112, `prop` 51, `config_schema` 1, `fmt-check` clean, and a release build with no warnings.

src/rabbitmq_stream_s3_config.erl
the-mikedavis: This is correct but the cost is worse than estimated, so I made a different fix. The look-ahead now takes one attempt with a 100ms timeout and no retries on its own. `peek_tail` corries the reason that a failure occurred, so we know whether to call with the pool timeout or the failure timeout. Offset resolution keeps the longer timeout window.

the-mikedavis:
I cleaned up the instances of "window" vocabulary. I think I have seen the same about the `log_reader` suite flake locally. Odd that it doesn't occur in CI

